PR #3, opened 2020-05-25
Necessary bits for packaging purposes

Spredzy:

PR #5, opened 2020-06-22
Packaging: Introduce setup.py

Spredzy:


tox.ini
```diff
@@ -1,14 +1,17 @@
...
 basepython = python3
 commands=
     poetry install
+    poetry env info
+
+[testenv:linters]
+commands=
+    {[testenv]commands}
     poetry run flake8 --version
     poetry run flake8 setup.py docs ansible_builder test
```
AlanCoding: It looks like the conflict is with https://github.com/ansible/ansible-builder/pull/3/files#diff-b91f3d5bd63fcd17221b267e851608e8L24, is the intent to revert that and then merge this?

Spredzy: Which conflict ? I am not following.

Spredzy: Dunno how this happened. Thanks for pointing it out. resolved.

AlanCoding:
to clarify the rule - nothing internally should ever use `setup.py`. This is adding and enforcing the file contents, but it's not using it.

Spredzy:
@AlanCoding This is an entirely true statement. From a contributor perspective, the only file that matters is `pyproject.toml` but give many tooling we use are not fully compatible with it we still need/require the presence of a `setup.py` file. As time goes by, I don't expect the `pyproject.toml` to change that much. So hopeffully this check on the long run will become close to transparent for mosst contributors.

AlanCoding (approved):

AlanCoding:
failure gave

> Because ansible-builder depends on dephell (^0.8.3) which doesn't match any versions, version solving failed.

and I'm not yet quite sure how to read this

Spredzy:
recheck

Spredzy:
I think I missed to commit the lock file changes. Let me addd it to the change and see what that gives.

shanemcd (approved):

PR #10, opened 2020-06-23
Add custom message when running 'create' without EE file

beeankha:


shanemcd (approved):

ansible_builder/main.py
```diff
@@ -29,8 +30,14 @@ class Definition:
     def __init__(self, *args, filename):
         self.filename = filename
 
-        with open(filename, 'r') as f:
-            self.raw = yaml.load(f)
+        try:
+            with open(filename, 'r') as f:
+                self.raw = yaml.load(f)
+        except FileNotFoundError:
```
Spredzy, line 36: @beeankha I know its too late given its merged but you can still catch a YAML exception and warn that the `execution-environment.yml` is not a valid yaml file.

Issue #18, opened 2020-06-25
Files referenced in -f option should be relative to the file, not cwd


AlanCoding:
consider:

```
$ cat alan/alan_ee.yml 
---
# subset of what may eventually be the container to replace the Ansible venv in AWX
version: 1
dependencies:
  galaxy: scm_metadata.yml
  python: requirements.txt
  system:
    - git  # for project updates
 
additional_build_steps: |
  RUN some-command
  ENV SOME_VAR=some_value
```

now if I run this command from the same directory (meaning `alan_ee.yml` is nested in the "alan" directory)

```
$ ansible-builder create -f alan/alan_ee.yml -c alan/bc
Traceback (most recent call last):
  File "/Users/alancoding/.virtualenvs/ansible-builder_test/bin/ansible-builder", line 11, in <module>
    load_entry_point('ansible-builder', 'console_scripts', 'ansible-builder')()
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/cli.py", line 34, in run
    ab = prepare()
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/cli.py", line 30, in prepare
    return AnsibleBuilder(**vars(args))
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 19, in __init__
    build_context=self.build_context)
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 65, in __init__
    self.build_steps()
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 71, in build_steps
    [self.steps.append(step) for step in GalaxySteps(containerfile=self)]
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 93, in __new__
    copy(src, dest)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/shutil.py", line 248, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: 'scm_metadata.yml'
```

it fails. But it runs in the other directory.

```
$ cd alan/
$ ansible-builder create -f alan_ee.yml -c bc
Complete! Build context is at: bc
[...]

PR #28, opened 2020-06-30
Read galaxy requirements.yml as relative path optionally

AlanCoding:
Connect https://github.com/ansible/ansible-builder/issues/18

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -77,7 +77,11 @@ class Definition:
 
     @property
     def galaxy_requirements_file(self):
-        return self.raw.get('dependencies', {}).get('galaxy')
+        galaxy_file = self.raw.get('dependencies', {}).get('galaxy')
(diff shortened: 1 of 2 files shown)


AlanCoding:
For background, this allows running with:

```
$ cat alan/alan_ee.yml 
---
# subset of what may eventually be the container to replace the Ansible venv in AWX
version: 1
dependencies:
  galaxy: scm_metadata.yml
  python: requirements.txt
  system:
    - git  # for project updates
 
additional_build_steps: |
  RUN some-command
  ENV SOME_VAR=some_value
```

Where `ansible-builder` is ran from the parent folder of the `alan` folder. Without this patch, this spec needs to put in `galaxy: alan/scm_metadata.yml`, which isn't portable at all. I could have put this same content in a folder with any other name. The name "alan" is very intentionally arbitrary.

beeankha (approved):
lgtm!

PR #29, opened 2020-06-30
Start a pattern for command help text for documentation

AlanCoding:


AlanCoding:
```
$ ansible-builder build --help
usage: ansible-builder build [-h] [-t TAG] [-f FILENAME] [-b BASE_IMAGE]
                             [-c BUILD_CONTEXT]
                             [--container-runtime CONTAINER_RUNTIME]

This does everything that the "create" command does, and then builds the
image. The build context will be populated from the execution environment
spec. After that, the specified container runtime podman/docker will be
invoked to build an image from that definition. After building the image, it
can be used locally or published using the supplied tag.

optional arguments:
  -h, --help            show this help message and exit
  -t TAG, --tag TAG     The name for the container being built.
  -f FILENAME, --file FILENAME
                        The definiton of the execution environment.
  -b BASE_IMAGE, --base-image BASE_IMAGE
                        The parent image for the execution environment.
  -c BUILD_CONTEXT, --context BUILD_CONTEXT
                        The directory to use for the build context. Defaults
                        to $PWD/context.
  --container-runtime CONTAINER_RUNTIME
                        Specifies which container runtime to use.
```

beeankha (approved):
I think this is fine to merge.  However, per some discussions I've had with @shanemcd ,  eventually I think we might want to separate out the help text into a separate file and import the help text into this `cli.py` file so we have a less "cluttered" structure for the CLI code.

Issue #12, opened 2020-06-23
New "Build" Command Needed in CLI


beeankha:
We need a new `build` command in the CLI, which will use the `-f`, `-b` and `-c` flags/arguments.

beeankha:
Some more notes:
- [x] ONLY for the `build` command, we need a `-t` `--tag` flag (this names the container)
- [x] After `create` gets called, there should be a path for the build context
- [x] The `build` command needs to call `docker build -f Containerfile .`
- [x] Need a  `--container-runtime` flag in order to specify either Docker or Podman builds (defaulting to Podman)
- [x] Implement a unit test that asserts that command contains certain elements, whether or not we override things like the `build context` path or the tag (from https://github.com/ansible/ansible-builder/pull/16#discussion_r445885193)

beeankha:
- [x] A visual "spinner" and/or stdout print in real-time would be nice while container is being built, etc.

Better Documentation:
- [x] Need to have a paragraph come up when running `ansible-builder -h` that explains what Builder is/does
- [x] Need to update the docstrings for the `create` and `build` commands to be more specific/helpful for the user

beeankha:
- [ ] Allow user to set a custom verbosity level
- [ ] Implement `mock` into more tests

AlanCoding:
Can we combine or close some of these, and just say "document and test the basic feature set"? @kdelee

kdelee:
I've tried out these commands and I think we can close this and yes, we need to write up a set of basic test cases we want to implement but that does not have to block closing this issue.

PR #35, opened 2020-07-06
Show More Output for Create/Build Commands

beeankha:
Connected to Issue https://github.com/ansible/ansible-builder/issues/12 

With this PR, output like the below will show up when running a command like `poetry run ansible-builder build --container-runtime docker -t hello_world`:

Screen Shot 2020-07-06 at 9 43 53 AM


AlanCoding (changes requested):
This needs changes to `description` of the `build_command_parser`

beeankha:
On second thought, I am re-combining `create` and `build` for when someone invokes the `ansible-builder build` command.   Might separate them out at another time, but I think for now this structure makes more sense.

AlanCoding (approved):

Issue #26, opened 2020-06-29
execution_environments.yml missing "dependencies" section makes ansible-builder build throw traceback


kdelee:
should fail and tell me nicely that my execution_environment.yml is missing the "dependencies" section

kdelee:
@shanemcd ran into this while demoing current state of things
@beeankha may be of interest

AlanCoding:
The tests make heavy use of empty specs.

https://github.com/ansible/ansible-builder/blob/31d565303b3e29de835340e75e486467eb467320/test/test_main.py#L53

It makes a Containerfile with just the base image right? I feel like it's too early to be mucking with this.

However, it occurred to me that we should warn if the user provides keys that _are not_ used, because that's more likely to be a valid complaint, and would probably solve the same issue you had. I don't think any human would make an empty spec file and expect an error message. I _do_ think humans will guess what keys to put in, get it wrong, and have that interpreted by the program as an empty spec.

_and when I say "spec" I should probably say "definition"_

kdelee:
this is what I'm showing: if you touch an empty file `execution-environment.yml` you can reproduce

Guess its doing a raw `get` on the `galaxy` subdict of `dependencies` 

```
$ touch execution-environment.yml
$ ansible-builder create
Traceback (most recent call last):
  File "/home/elijah/.cache/pypoetry/virtualenvs/ansible-builder-d47oRQeY-py3.8/bin/ansible-builder", line 11, in <module>
    load_entry_point('ansible-builder', 'console_scripts', 'ansible-builder')()
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/cli.py", line 77, in run
    ab = prepare()
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/cli.py", line 73, in prepare
    return AnsibleBuilder(**vars(args))
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/main.py", line 23, in __init__
    self.containerfile = Containerfile(
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/main.py", line 100, in __init__
    self.build_steps()
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/main.py", line 106, in build_steps
    [self.steps.append(step) for step in GalaxySteps(containerfile=self)]
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/main.py", line 124, in __new__
    if not definition.galaxy_requirements_file:
[...]

PR #38, opened 2020-07-07
Introspect collection content inside of container

shanemcd:


ansible_builder/main.py
```diff
@@ -145,62 +175,63 @@ def get_dependency(self, entry):
...
-        requirements_path = self.definition.get_dependency('galaxy')
-        if requirements_path:
+        galaxy_requirements_path = self.definition.get_dependency('galaxy')
+        if galaxy_requirements_path:
             # TODO: what if build context file exists? https://github.com/ansible/ansible-builder/issues/20
-            shutil.copy(requirements_path, self.build_context)
+            dest = os.path.join(self.build_context, galaxy_requirements_path)
+            exists = os.path.exists(dest)
+            if not exists or not filecmp.cmp(galaxy_requirements_path, dest, shallow=False):
+                shutil.copy(galaxy_requirements_path, dest)
```
AlanCoding, line 146: For the example I had, this is giving me:

```
Traceback (most recent call last):
  File "/Users/alancoding/.virtualenvs/ansible-builder/bin/ansible-builder", line 33, in <module>
    sys.exit(load_entry_point('ansible-builder', 'console_scripts', 'ansible-builder')())
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/cli.py", line 85, in run
    if action():
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 48, in build
    self.containerfile.prepare_galaxy_steps()
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 209, in prepare_galaxy_steps
    shutil.copy(galaxy_requirements_path, dest)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/shutil.py", line 248, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: 'examples/bc/examples/pytz/requirements.yml'
```

I think the usage is reasonable, and I'm sure we can get a unit test to cover this.

ansible_builder/utils.py
```diff
@@ -1,8 +1,51 @@
...
+            output.append(line.rstrip())
+        sys.stdout.write(line)
+
+    rc = process.poll()
+    return (rc, output)
+
+
+def introspect():
+    paths = []
+    path_root = os.path.join(constants.base_collections_path, 'ansible_collections')
```
AlanCoding: I really want to make this path configurable. That's lighting up all kinds of ideas in my head. This might get added as an option to _all_ commands. That's an interesting discussion to have, but doesn't need to be covered here in any way.

ansible_builder/cli.py
```diff
@@ -8,6 +8,24 @@
 
 
 def prepare(args=sys.argv[1:]):
+    args = parse_args(args)
+    return AnsibleBuilder(**vars(args))
+
+
+def run():
+    args = parse_args()
+    if args.action in ['build', 'create']:
```
AlanCoding: looks like something else needs to be hooked up, trying to run this I get

```
$ ansible-builder introspect
An error has occured.
```

shanemcd: Oops. This should be fixed now 😅

ansible_builder/utils.py
```diff
@@ -1,8 +1,51 @@
...
+    if not os.path.exists(path_root):
+        # add debug statements at points like this
+        return paths
+    for namespace in sorted(os.listdir(path_root)):
+        for name in sorted(os.listdir(os.path.join(path_root, namespace))):
```
AlanCoding: There should be a check before this that the location is a directory, otherwise `os.listdir` can error being given a file.

ansible_builder/cli.py
```diff
@@ -43,6 +66,8 @@ def prepare(args=sys.argv[1:]):
             'After building the image, it can be used locally or published using the supplied tag.'
         )
     )
+
+    subparsers.add_parser('introspect')
     # TODO: Need to update the docstrings for the create and build commands to be more specific/helpful
```
beeankha: This `TODO` is left over from a previous PR; @AlanCoding changed the docstrings for those commands so this line can now be deleted.

Same with the line above that states `# TODO: Need to have a paragraph come up when running ansible-builder -h that explains what Builder is/does`

ansible_builder/main.py
```diff
@@ -145,62 +147,63 @@ def get_dependency(self, entry):
...
-    def build_steps(self):
+    def prepare_galaxy_steps(self):
         self.steps = [
             "FROM {}".format(self.base_image),
             ""
         ]
-        requirements_path = self.definition.get_dependency('galaxy')
-        if requirements_path:
+        galaxy_requirements_path = self.definition.get_dependency('galaxy')
+        if galaxy_requirements_path:
```
kdelee, line 139: @beeankha I think the `else` condition of this `if` block would be equivalent to "no galaxy requirements found" and if we want to log a statement we would do it in an `else` block after this

ansible_builder/main.py
```diff
@@ -145,62 +147,63 @@ def get_dependency(self, entry):
...
+
+
+        command = [
+            self.container_runtime, "run", "--rm", self.tag,
+            "python", "-c", "from ansible_builder.utils import introspect; introspect()"
+        ]
+        rc, output = run_command(command, capture_output=True)
+        if rc is not 0:
+            print('No collections requirements file found, skipping ansible-galaxy install...')
+            return
```
beeankha: This should address the traceback issue filed in https://github.com/ansible/ansible-builder/issues/26, @kdelee

ansible_builder/utils.py
```diff
@@ -1,8 +1,56 @@
 import subprocess
+import sys
+import os
 
+from . import constants
```
AlanCoding: A design tweak I would like to make is avoiding importing other modules inside of utils or constants. Given how we're moving actual source around, I would prefer hard code the value twice.

To go even further, the default that would make the most sense is to search _all_ the paths in the Ansible core default, and return categories for each in the YAML output, exactly like `ansible-galaxy collection list`. That way, the method speaks a universal truth and cuts off ties to implementation of the rest of the project.

ansible_builder/main.py
```diff
@@ -145,62 +148,67 @@ def get_dependency(self, entry):
...
         if python_req_path:
             shutil.copy(python_req_path, self.build_context)
+
+
+        command = [
+            self.container_runtime, "run", "--rm", self.tag, "python", "-c",
+            textwrap.dedent("""
+                import yaml
+                from ansible_builder.utils import introspect
+                print(yaml.dump(introspect(), default_flow_style=False))
```
AlanCoding: Just make this a method in `ansible_builder.utils`, then condense this to

```
python -c "from ansible_builder.utils import printspection; printspection()"
```

method name subject to revision.

ansible_builder/main.py
```diff
@@ -44,8 +44,13 @@ def build_command(self):
         ]
 
     def build(self):
-        self.create()
-        return run_command(self.build_command())
+        self.containerfile.prepare_galaxy_steps()
+        self.containerfile.write()
```
AlanCoding, line 50: It's very possible that the code errors after this, but before the final `.write()` succeeds. In that case, the user needs to know. I would suggest that we add a comment to the end of the Containerfile that it's incomplete. It would also be good to add debug statements before each write. Both would be best.

ansible_builder/cli.py
```diff
@@ -71,20 +78,7 @@ def prepare(args=sys.argv[1:]):
                        help='Specifies which container runtime to use; use for both "build" and "create" commands. '
                        'Defaults to podman.')
 
+    #import pdb; pdb.set_trace()
```
AlanCoding: delete line

AlanCoding:
The big thing: I would strongly prefer that the `ansible-builder` package not be in the base image.

```
$ grep -r "CollectionDefinition" ansible_builder
ansible_builder/utils.py:    from .main import CollectionDefinition # avoid circular import
ansible_builder/utils.py:        CD = CollectionDefinition(path)
ansible_builder/main.py:class CollectionDefinition(BaseDefinition):
```

There's not a whole lot that `CollectionDefinition` does, and it's only ever used by the utils once, and only for one thing which could be replaced quickly. I only bring that up because importing _that_ shouldn't be a reason to require the entire package. Making the `introspect` method fully self-contained is good idea all-around. This logic may get handed off to the Galaxy team eventually as this is moved to the API.

This is a hard problem, I've thought about several bad alternatives, such as a subprocess to call Ansible to run a module with the introspect logic, or use `podman cp` to copy an introspect script over. Not good ideas.

What I think would be a good idea:

Make a new module like `ansible_builder.introspect` with a `if __name__ == '__main__':` bit in it. Have the build command copy that into the build context and a line to the Containerfile `ADD introspect.py /build/`. Then invoke that as a python script. Then remove `ansible-builder` from the base image.

----

With that all out of the way, I'm okay to merge it and refactor that specific thing later, if we're on the same page.

ansible_builder/main.py
AlanCoding: This won't work, it regresses the case where there is a user-specified python file

```yaml
---
version: 1
dependencies:
  python: project/requirements.txt
```

The requirements file gets copied into the build context, but then it never gets added in the Containerfile.

ansible_builder/main.py
```diff
@@ -145,62 +150,68 @@ def get_dependency(self, entry):
...
+                 tag=constants.default_tag):
 
         self.build_context = build_context
         os.makedirs(self.build_context, exist_ok=True)
         self.definition = definition
         self.path = os.path.join(self.build_context, filename)
         self.base_image = base_image
+        self.container_runtime = container_runtime
+        self.tag = tag
+        self.steps = []
```
AlanCoding: The code `self.steps =` shouldn't occur twice. This should be replaced by

```python
        self.steps = [
            "FROM {}".format(self.base_image),
            ""
        ]
```

and have that removed from `prepare_galaxy_steps`. As it is it both prepares galaxy steps and initializes the steps... although they have already been initialized.

AlanCoding (approved):

ansible_builder/main.py
```diff
@@ -173,6 +125,15 @@ def __init__(self, definition,
             ""
         ]
 
+    def prepare_introspection_steps(self):
+        source = ansible_builder.introspect.__file__
```
AlanCoding, line 129: 👍

AlanCoding (approved):

beeankha (approved):
:shipit:

Issue #21, opened 2020-06-25
Process direct requirements and additional build steps from the user's spec


AlanCoding:
Example `execution-environment.yml`:

(EDIT: direct listing of system dependencies not supported after all)

```yaml
---
version: 1
dependencies:
  python: requirements.txt
  system:
    - git

additional_build_steps: |
  RUN some-command
  ENV SOME_VAR=some_value
```

This tool should output a Container file with:

 - `pip` install for requirements
 - `dnf`? install for system requirement
 - additional lines for whatever is given in the additional specs

Schema of EE spec is subject to change, but for now we want to just get something in.

Note that `galaxy` in dependencies is processed in the current code base.

kdelee:
@AlanCoding example given in issue has a couple problems:

1) https://github.com/ansible/ansible-builder/issues/50 is not done, including the list of system dependencies causes a traceback, but will delay looking at that until the issue is closed
2) looks like code can't handle the multiline string you are using in your example, it wants a `prepend` and/or `append` section

```
cat execution-environment.yml 
---
version: 1
dependencies:
  python: requirements.txt

additional_build_steps: |
  RUN touch foo
  ENV ELIJAH_VAR=elijah
```

```
 ansible-builder build
requirements.txt
requirements.txt
Traceback (most recent call last):
  File "/home/elijah/.cache/pypoetry/virtualenvs/ansible-builder-d47oRQeY-py3.8/bin/ansible-builder", line 11, in <module>
    load_entry_point('ansible-builder', 'console_scripts', 'ansible-builder')()
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/cli.py", line 20, in run
    if action():
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/main.py", line 57, in build
    self.containerfile.prepare_prepended_steps()
  File "/home/elijah/sfw/ansible/ansible-builder/ansible_builder/main.py", line 205, in prepare_prepended_steps
    prepended_steps = additional_prepend_steps.get('prepend')
AttributeError: 'str' object has no attribute 'get'
```

looks like need to gracefully reject a `additional_build_steps` section that is not a dictionary 

put up a PR with possible error message enhancement

kdelee:
now testing w/ proper syntax, and its working for me:
```
---
version: 1
dependencies:
  python: requirements.txt

additional_build_steps:
  prepend: |
    ENV ELIJAH_VAR=elijah
```
[...]

PR #46, opened 2020-07-10
Enable Additional Build Commands to be Specified in Definition File

beeankha:
Addresses part of Issue https://github.com/ansible/ansible-builder/issues/21.

With this PR, a user can add info in an `additional_build_steps` section in the Execution Environment definition file, like so:

Screen Shot 2020-07-10 at 12 40 34 PM

...and the commands will be detected + recorded during the `ansible-builder build` process:

Screen Shot 2020-07-10 at 12 40 20 PM

The containerfile will get those additional commands written into it:

Screen Shot 2020-07-10 at 1 17 32 PM

Commands can be run before requirements are dealt with (`prepend`) vs after (`append`), which allows for more customization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,21 @@ version: 1
 dependencies:
   galaxy: requirements.yml
   python: requirements.txt
+
+additional_build_steps:
+  prepend: |
+    RUN whoami
+    RUN cat /etc/os-release
+  append:
+    - RUN echo This is a post-install command!
+    - RUN ls -la /etc
 ```
 
-The entries such as `requirements.yml` and `requirements.txt` may be a relative
-path from the directory of the execution environment definition's folder,
-or an absolute path.
+The entries such as `requirements.yml` and `requirements.txt` may be a relative path from the directory of the execution environment definition's folder, or an absolute path.
+
+Additional commands may be specified in the `additional_build_steps` section, either for before the main build steps (`prepend`) or after (`append`).  The syntax needs to be either a:
+  - multi-line string (example shown in the `prepend` section above)
+  - dictionary (as shown via `append`)
 
 ## Collection Execution Environment Dependencies
 
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -5,7 +5,7 @@ import shutil
 import filecmp
 
 from . import constants
-from .steps import GalaxySteps, PipSteps, IntrospectionSteps
+from .steps import AdditionalBuildSteps, GalaxySteps, PipSteps, IntrospectionSteps
 from .utils import run_command
 import ansible_builder.introspect
 
@@ -16,6 +16,7 @@ class AnsibleBuilder:
                  base_image=constants.default_base_image,
                  build_context=constants.default_build_context,
                  tag=constants.default_tag,
+                 get_additional_commands=None,
                  container_runtime=constants.default_container_runtime):
         self.action = action
         self.definition = UserDefinition(filename=filename)
@@ -28,7 +29,8 @@ class AnsibleBuilder:
             base_image=base_image,
             build_context=self.build_context,
             container_runtime=self.container_runtime,
-            tag=self.tag)
+            tag=self.tag,
+            get_additional_commands=get_additional_commands)
 
     @property
     def version(self):
@@ -44,6 +46,7 @@ class AnsibleBuilder:
         ]
 
     def build(self):
+        self.containerfile.prepare_prepended_steps()
         self.containerfile.prepare_introspection_steps()
         self.containerfile.prepare_galaxy_steps()
         print('Writing partial Containerfile without collection requirements')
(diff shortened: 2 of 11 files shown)


beeankha:
With the addition of `colors.py` module, output is now more colorful and easier to parse:

Screen Shot 2020-07-10 at 2 03 25 PM

ansible_builder/main.py
```diff
@@ -44,13 +47,15 @@ def build_command(self):
...
+        self.containerfile.prepare_prepended_steps()
         self.containerfile.prepare_introspection_steps()
         self.containerfile.prepare_galaxy_steps()
-        print('Writing partial Containerfile without collection requirements')
+        print(MessageColors.OK + 'Writing partial Containerfile without collection requirements' + MessageColors.ENDC)
         self.containerfile.write()
         run_command(self.build_command)
         self.containerfile.prepare_pip_steps()
-        print('Rewriting Containerfile to capture collection requirements')
+        print(MessageColors.OK + 'Rewriting Containerfile to capture collection requirements' + MessageColors.ENDC)
```
AlanCoding, line 56: This is really nicely written.

If I was going to point to the next thing to do, I would prefer to route this through an internal `debug` method, and pass the value OK/FAIL/etc. to that method, just so that the main method isn't touching the colors directly. But that's thinking about additional enhancement, and what you have is 👍

ansible_builder/main.py
```diff
@@ -120,11 +131,30 @@ def __init__(self, definition,
         self.base_image = base_image
         self.container_runtime = container_runtime
         self.tag = tag
+        self.get_additional_commands = get_additional_commands
```
AlanCoding: Did you mean to remove this?

The method `UserDefinition.get_additional_commands` is important, and is clear. However, `AnsibleBuilder. get_additional_commands` and `Containerfile. get_additional_commands` are dynamically assigned, and I don't see it doing anything, it looks like it would always be `None`, and would never be referenced.

ansible_builder/steps.py
```diff
@@ -1,6 +1,26 @@
...
+        in the Containerfile or Dockerfile.
+        """
+        self.steps = []
+        if isinstance(additional_steps, str):
+            lines = additional_steps.strip().splitlines()
+        elif isinstance(additional_steps, list):
+            lines = additional_steps
+        else:
+            lines = []
+            print(MessageColors.WARNING + "Unknown type found for additional_build_steps.  Ignoring..." + MessageColors.ENDC)
```
AlanCoding: my own preference would be failure.

As the volume of output grows, it becomes less and less likely that the user will see this warning. The only way to make sure they see it is to refuse to produce the output.

README.md
```diff
@@ -34,11 +34,21 @@ version: 1
...
 ```
 
-The entries such as `requirements.yml` and `requirements.txt` may be a relative
-path from the directory of the execution environment definition's folder,
-or an absolute path.
+The entries such as `requirements.yml` and `requirements.txt` may be a relative path from the directory of the execution environment definition's folder, or an absolute path.
+
+Additional commands may be specified in the `additional_build_steps` section, either for before the main build steps (`prepend`) or after (`append`).  The syntax needs to be either a:
+  - multi-line string (example shown in the `prepend` section above)
+  - dictionary (as shown via `append`)
```
AlanCoding, line 51: I don't quite see the need to support 2 different kinds of syntax at this point, but I defer to @shanemcd's decisions on those.

AlanCoding (approved):
Once we get the branches together, I'm ready to use this to complete the subversion example.

ansible_builder/steps.py
beeankha: Changed!

README.md
shanemcd: I dont feel strongly enough either way. I went back and forth between these 2 options when I was daydreaming, and it was enough to just do both. I will defer to @AlanCoding 😉

shanemcd (approved):
This is amazing. Thank you 😄

ansible_builder/main.py
```diff
@@ -81,11 +85,23 @@ def __init__(self, filename):
...
+            if isinstance(e, yaml.parser.ParserError):
+                sys.exit(MessageColors.FAIL + textwrap.dedent("""
+                An error occured while parsing the definition file:
+                {0}
+                """).format(str(e)) + MessageColors.ENDC)
+            elif isinstance(e, FileNotFoundError):
+                sys.exit(MessageColors.FAIL + """
+                Could not detect '{0}' file in this directory.
+                Use -f to specify a different location.
+                """.format(constants.default_file) + MessageColors.ENDC)
```
AlanCoding: I think what you really want is

```python
except FileNotFoundError:
  ...
except yaml.parser.ParserError as e:
  ...
```

same thing functionally, just a matter of python standard practice

PR #53, opened 2020-07-26
[alt 1] Remove the unused prepare method

AlanCoding:
I was trying to outline the program flow, and noticed that `prepare` didn't have any references to it, but we still had tests running it. If someone wants to refactor the actual code to make the testing prettier, then go ahead, but I'm trying to avoid doing that unnecessarily.

### DIFF
--- a/test/test_cli.py
+++ b/test/test_cli.py
@@ -1,18 +1,29 @@
-from ansible_builder.cli import prepare
+from ansible_builder.cli import run, sys
+from contextlib import contextmanager
+
+import os
+import pytest
+
+from unittest import mock
 
 
-def test_custom_image(exec_env_definition_file, tmpdir):
+@pytest.fixture
+def cli_mock():
+    @contextmanager
+    def rf(args):
+        with mock.patch.object(sys, 'argv', args):
+            with pytest.raises(SystemExit):
+                yield
+    return rf
+
+
+def test_custom_image(exec_env_definition_file, tmpdir, cli_mock):
     content = {'version': 1}
     path = str(exec_env_definition_file(content=content))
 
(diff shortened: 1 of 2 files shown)


test/test_cli.py
```diff
@@ -1,18 +1,39 @@
...
-def test_build_context(good_exec_env_definition_path, tmpdir):
-    path = str(good_exec_env_definition_path)
-    build_context = str(tmpdir)
-    aee = prepare(['build', '-f', path, '-c', build_context])
+def test_invalid_path(tmpdir, cli_mock):
+    args = ['ansible-builder', 'build', '-f', 'foo/invalid', '-c', str(tmpdir)]
+    with cli_mock(args) as sys_exit:
+        run()
 
-    assert aee.build_context == build_context
```
shanemcd, line 18: I'm a little concerned that this isn't actually testing anything now. I created the `prepare` function so that we could test that CLI args are parsed correctly and the appropriate attributes are set on the `AnsibleBuilder` object, without invoking the rest of the program. It seems like this is now just testing your mock. Am I missing something?

AlanCoding: But when you actually run the program, you don't run `prepare`, you run `run`. It doesn't matter if `prepare` parses the args correctly or not, because it has no bearing on what happens when the user it.

If you just need a helper methods for the test, then define `prepare` inside of this file.

This version is absolutely testing something. In fact, it's probably testing too much, becoming something more like an integration test with the only exception that it has the `autouse` mock of `run_command` in place.

shanemcd:
Thank you for being flexible. Closing this in favor of https://github.com/ansible/ansible-builder/pull/55.

PR #52, opened 2020-07-24
Accept user bindep file

AlanCoding:
Connect https://github.com/ansible/ansible-builder/issues/50

Running the subversion example

```
$ ./examples/run.sh subversion

Running Example: subversion
Untagged: exec-env-subversion:latest
Writing partial Containerfile without collection requirements
Running command:
  docker build -f bc/Dockerfile -t exec-env-subversion bc
Sending build context to Docker daemon  7.168kB
Step 1/3 : FROM shanemcd/ansible-runner
 ---> f0c79677d9b1
Step 2/3 : ADD introspect.py /usr/local/bin/introspect
 ---> Using cache
 ---> cba3da5faf2e
Step 3/3 : RUN chmod +x /usr/local/bin/introspect
 ---> Using cache
 ---> 13c1dc7b98f2
Successfully built 13c1dc7b98f2
Successfully tagged exec-env-subversion:latest
Running command:
  docker run --rm exec-env-subversion introspect
[]

Rewriting Containerfile to capture collection requirements
Running command:
  docker build -f bc/Dockerfile -t exec-env-subversion bc
Sending build context to Docker daemon  8.192kB
Step 1/6 : FROM shanemcd/ansible-runner
 ---> f0c79677d9b1
Step 2/6 : ADD introspect.py /usr/local/bin/introspect
 ---> Using cache
 ---> cba3da5faf2e
Step 3/6 : RUN chmod +x /usr/local/bin/introspect
 ---> Using cache
 ---> 13c1dc7b98f2
Step 4/6 : ADD bindep.txt /build/
 ---> Using cache
 ---> a8e7d75fea57
Step 5/6 : RUN pip3 install bindep
 ---> Using cache
 ---> 303631a88289
Step 6/6 : RUN yum -y install $(bindep -b -f /build/bindep.txt)
 ---> Running in aea05bc2c333
CentOS-8 - AppStream                            2.1 MB/s | 5.8 MB     00:02    
CentOS-8 - Base                                 1.3 MB/s | 2.2 MB     00:01    
CentOS-8 - Extras                                25 kB/s | 7.0 kB     00:00    
Ansible Runner for EL 8 - x86_64                 35 kB/s | 7.7 kB     00:00    
Extra Packages for Enterprise Linux Modular 8 - 184 kB/s |  82 kB     00:00    
Extra Packages for Enterprise Linux 8 - x86_64  8.6 MB/s | 7.4 MB     00:00    
Dependencies resolved.
================================================================================
[...]

PR #34, opened 2020-07-02
Install collection metadata dependencies

AlanCoding:
Connect https://github.com/ansible/ansible-builder/issues/22

this is a solution to that issue, for the limited case of python requirements. We have had productive conversations about the system requirements, but right now I do not have all the syntax pieces needed to write a Containerfile for bindep processing, much less the ability to incorporate that into the tool.

The output of the example:

```
$ source examples/pytz/run.sh 
Untagged: ansible-execution-env:latest
Deleted: sha256:a63317cd59d847ba8be5b703a597ddd175642e9222c4ff1debf712dc3f53e9e3
Deleted: sha256:bdd252c7c6ac905fa327aa31c9faacaeed95d20cbbfdd9a7ab08c7feb4480730
Using temporary directory to obtain collection information:
  /var/folders/fj/6z5pn3wj5w31zvn2pws34svm0000gn/T/ansible_builder_q0bbxune
Running command:
  ansible-galaxy collection install -r examples/pytz/requirements.yml -p /var/folders/fj/6z5pn3wj5w31zvn2pws34svm0000gn/T/ansible_builder_q0bbxune
[WARNING]: You are running the development version of Ansible. You should only run Ansible from "devel" if you are modifying the Ansible engine, or trying out features under
development. This is a rapidly changing source of code and can become unstable at any point.
Starting galaxy collection install process
[WARNING]: The specified collections path '/var/folders/fj/6z5pn3wj5w31zvn2pws34svm0000gn/T/ansible_builder_q0bbxune' is not part of the configured Ansible collections paths
'/Users/alancoding/.ansible/collections:/usr/share/ansible/collections'. The installed collection won't be picked up in an Ansible run.
Process install dependency map
Starting collection install process
Installing 'awx.awx:0.0.1-devel' to '/var/folders/fj/6z5pn3wj5w31zvn2pws34svm0000gn/T/ansible_builder_q0bbxune/ansible_collections/awx/awx'
Created collection for awx.awx at /var/folders/fj/6z5pn3wj5w31zvn2pws34svm0000gn/T/ansible_builder_q0bbxune/ansible_collections/awx/awx
awx.awx (0.0.1-devel) was installed successfully
Running command:
  docker build -f examples/bc/Dockerfile -t ansible-execution-env:latest examples/bc
Sending build context to Docker daemon  3.072kB
Step 1/5 : FROM shanemcd/ansible-runner
 ---> 076625a99a3b
Step 2/5 : ADD requirements.yml /build/
[...]

PR #54, opened 2020-07-27
Add validation step, combine makedirs and copy code

### LINKED ISSUE #17: Clean error messages for missing files
If the data inside of the EE spec doesn't exist, it falls back to traceback.

```
$ ansible-builder create -f alan/alan_ee.yml -c alan/bc
Traceback (most recent call last):
  File "/Users/alancoding/.virtualenvs/ansible-builder_test/bin/ansible-builder", line 11, in <module>
    load_entry_point('ansible-builder', 'console_scripts', 'ansible-builder')()
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/cli.py", line 34, in run
    ab = prepare()
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/cli.py", line 30, in prepare
    return AnsibleBuilder(**vars(args))
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 19, in __init__
    build_context=self.build_context)
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 65, in __init__
    self.build_steps()
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 71, in build_steps
    [self.steps.append(step) for step in GalaxySteps(containerfile=self)]
  File "/Users/alancoding/Documents/repos/ansible-builder/ansible_builder/main.py", line 93, in __new__
    copy(src, dest)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/shutil.py", line 248, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: 'adjusted_requirements.yml'
```

where

```
$ cat alan/alan_ee.yml 
---
# subset of what may eventually be the container to replace the Ansible venv in AWX
version: 1
dependencies:
  galaxy: adjusted_requirements.yml
  python: requirements.txt
  system:
    - git  # for project updates
 
additional_build_steps: |
  RUN some-command
  ENV SOME_VAR=some_value
```

and the file `adjusted_requirements.yml` does not exist.

### LINKED ISSUE #20: Figure out how to handle existing build context folder
Right now, if a folder already exists at the location for `-c`, the context, it will merge new files with whatever is in there.

This is probably undesirable, because any files that are not created anew are will almost certainly not going to be used, and that goes against [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

> Inadvertently including files that are not necessary for building an image results in a larger build context and larger image size. This can increase the time to build the image, time to pull and push it, and the container runtime size.

So if the build context already exists, then sensible default behaviors could be:

 - delete the folder
 - error
 - warn

_feedback requested_

Maybe this could become a toggle via a `--force` option?

AlanCoding:
I'm willing to say this Fixes #20 and Fixes #17 

The core reason for doing this is that `python` and `galaxy` are accepted in dependencies now. I didn't realize the complexity of what @shanemcd had previously done with `filecmp.cmp`, and this right away reveals that we are inconsistent with how we treat the python file vs. the galaxy file. We may _always_ copy one, but not the other. I'm looking at adding the `system` file in #52, so it's critical that whatever validation we do in advance of copying this stuff happens consistently for all the files.

### tests

This undoes a lot of what I did in https://github.com/ansible/ansible-builder/pull/34/files#diff-3e85841b8cff61e2a613aa9b0c63a081, which fixed tests leaving artifacts around. Now, a fake directory is no longer necessary for things that don't actually call the `.build` command.

### file paths

The name `get_dep_abs_path` is intentionally obscure, because it's something that we should need early in the process, but only briefly, and never after then. The python/galaxy/system files could be any crazy place on the computer, but `ansible-builder` will copy them into the build context. After they are copied, they only have their `basename` like `requirements.yml`. The code is confusing switching between the source path and the plain path. This formats it so that we will always use the plain path when in the parts figuring out the galaxy steps.

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -52,7 +52,7 @@ class AnsibleBuilder:
         ]
 
     def build(self):
-        self.containerfile.initialize_folder()
+        self.containerfile.create_folder_copy_files()
         self.containerfile.prepare_prepended_steps()
         self.containerfile.prepare_introspection_steps()
         self.containerfile.prepare_galaxy_steps()
@@ -171,7 +171,7 @@ class Containerfile:
             ""
         ]
 
-    def initialize_folder(self):
+    def create_folder_copy_files(self):
         """Creates the build context file for this Containerfile
         moves files from the definition into the folder
         """
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -1,6 +1,5 @@
 import os
 import yaml
-import sys
 import shutil
 import filecmp
 import textwrap
@@ -10,6 +9,14 @@ from .colors import MessageColors
 from .steps import AdditionalBuildSteps, GalaxySteps, PipSteps, IntrospectionSteps
 from .utils import run_command
 import ansible_builder.introspect
+
+
+# Files that need to be moved into the build context
+CONTEXT_FILES = ['galaxy', 'python', 'system']
+
+
+class DefinitionError(RuntimeError):
+    pass
 
 
 class AnsibleBuilder:
@@ -25,7 +32,6 @@ class AnsibleBuilder:
         self.build_context = build_context
         self.container_runtime = container_runtime
         self.containerfile = Containerfile(
-            filename=constants.runtime_files[self.container_runtime],
             definition=self.definition,
             base_image=base_image,
             build_context=self.build_context,
@@ -46,6 +52,7 @@ class AnsibleBuilder:
         ]
 
     def build(self):
+        self.containerfile.initialize_folder()
         self.containerfile.prepare_prepended_steps()
         self.containerfile.prepare_introspection_steps()
         self.containerfile.prepare_galaxy_steps()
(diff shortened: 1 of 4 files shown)


ansible_builder/main.py
```diff
@@ -46,6 +52,7 @@ def build_command(self):
         ]
 
     def build(self):
+        self.containerfile.initialize_folder()
```
AlanCoding: Key point here:

Adding a call at the start of `.build` that will create the build context folder, and also copy files into it.

That's the big organizational change that most of the rest of this diff supports.

shanemcd: Not crazy about this method name. `initialize_build_context`?

AlanCoding: it creates the folder and copies files, so what about `create_folder_copy_files`

I'm fine with any of the names mentioned so far.

shanemcd (approved):

PR #57, opened 2020-07-30
Add processing of python requirements to run AWX case to completion

### LINKED ISSUE #45: Scrub the python requirement files for duplicate and unnecessary requirements
Installing the "awx" case in the examples right now will give this error:

```
Double requirement given: ansible>=2.9.0 (from -r /usr/share/ansible/collections/ansible_collections/ovirt/ovirt/requirements.txt (line 1)) (already in ansible (from -r /usr/share/ansible/collections/ansible_collections/ansible/netcommon/requirements.txt (line 1)), name='ansible')
The command '/bin/sh -c pip3 install     -r /usr/share/ansible/collections/ansible_collections/ansible/netcommon/requirements.txt     -r /usr/share/ansible/collections/ansible_collections/ansible/posix/requirements.txt     -r /usr/share/ansible/collections/ansible_collections/awx/awx/requirements.txt     -r /usr/share/ansible/collections/ansible_collections/azure/azcollection/requirements-azure.txt     -r /usr/share/ansible/collections/ansible_collections/community/vmware/requirements.txt     -r /usr/share/ansible/collections/ansible_collections/google/cloud/requirements.txt     -r /usr/share/ansible/collections/ansible_collections/ovirt/ovirt/requirements.txt' returned a non-zero code: 1
```

All `ansible` entries can be safely removed from this command, and the same goes for a number of common test libraries.

Any duplicate entires can be safely removed if it is confirmed that the version range does not conflict.

However, reading the requirements file in the first place is a larger challenge. Some implementations of that:

 - https://github.com/davidfischer/requirements-parser/blob/master/requirements/parser.py
 - https://stackoverflow.com/questions/49689880/proper-way-to-parse-requirements-file-after-pip-upgrade-to-pip-10-x-x/59969843#59969843

Note that this issue does not rise to the level of any form of "dependency resolution". This is only slightly more complicated than parsing python `requirements.txt`. Our bar for success for this specific work item is that the [cataloged](https://github.com/AlanCoding/collection-dependencies-demo/blob/master/sniff_req/discovered.json) requirements of collections migrated from Ansible 2.9 can mutually install without error.

AlanCoding:
Fixes #45 

This gets it so that the "awx" case in my examples PR (also still open) will run to completion and finish the playbook.

This changes several major mechanics of how `ansible-builder` runs:

 - `introspect.py` now returns lines of requirements files, instead of the location of requirements files
 - a dependency is added to parse python requirements
 - after reading the output of the introspect script, the main program does light processing of the requirements, and combines then into a single `requirements.txt` inside of the build context. This also means that the user's python requirements are combined, and not copied into the build context anymore
 - obviously, the pip steps are now just 1 `-r`

### DIFF
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -8,6 +8,7 @@ from .colors import MessageColors
 from .main import AnsibleBuilder, DefinitionError
 from . import constants
 from .introspect import add_introspect_options, process
+from .requirements import sanitize_requirements
 
 
 def run():
@@ -25,9 +26,15 @@ def run():
     elif args.action == 'introspect':
         for folder in args.folders:
             data = process(folder)
-            print()
-            print('Dependency data for {0}'.format(folder))
-            print(yaml.dump(data, default_flow_style=False))
+            if args.sanitize:
+                data['python'] = sanitize_requirements(data['python'])
+                print()
+                print('Sanitized dependencies for {0}'.format(folder))
+                print(yaml.dump(data, default_flow_style=False))
+            else:
+                print()
+                print('Dependency data for {0}'.format(folder))
+                print(yaml.dump(data, default_flow_style=False))
         sys.exit(0)
 
     print(MessageColors.FAIL + "An error has occured." + MessageColors.ENDC)
@@ -97,6 +104,13 @@ def parse_args(args=sys.argv[1:]):
         )
     )
     add_introspect_options(introspect_parser)
+    introspect_parser.add_argument(
+        '--sanitize', help=(
+            'Sanite and de-duplicate requirements. '
+            'This is normally done separately from the introspect script, but this '
+            'option is given to more accurately test collection content.'
+        ), action='store_true'
+    )
 
     args = parser.parse_args(args)
 
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -106,7 +106,7 @@ def parse_args(args=sys.argv[1:]):
     add_introspect_options(introspect_parser)
     introspect_parser.add_argument(
         '--sanitize', help=(
-            'Sanite and de-duplicate requirements. '
+            'Sanitize and de-duplicate requirements. '
             'This is normally done separately from the introspect script, but this '
             'option is given to more accurately test collection content.'
         ), action='store_true'
--- a/ansible_builder/introspect.py
+++ b/ansible_builder/introspect.py
@@ -5,15 +5,31 @@ import sys
 import yaml
 import argparse
 
+
 base_collections_path = '/usr/share/ansible/collections'
 default_file = 'execution-environment.yml'
+
+
+def pip_file_data(path):
+    req_list = []
+    with open(path, 'r') as f:
+        for line in f.read().split('\n'):
+            if not line:
+                continue
+            if line.startswith('-r') or line.startswith('--requirement'):
+                _, new_filename = line.split(None, 1)
+                new_path = os.path.join(os.path.dirname(path or '.'), new_filename)
+                req_list.extend(pip_file_data(new_path))
+            else:
+                req_list.append(line)
+    return req_list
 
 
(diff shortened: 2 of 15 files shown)


AlanCoding:
This is still going to have some rough edges, but the real _dealbreakers_ like throwing in `--no-binary` options and stuff, have been resolved and the example is running well enough at this point.

Also, demo of this stuff:

```
$ ansible-builder introspect test/data/

Dependency data for test/data/
python:
- pyvcloud>=14
- pytz
- tacacs_plus
- 'pyvcloud>=18.0.10  # duplicate with test.metadata collection'
system:
- test/bindep/bindep.txt
```

and

```
$ ansible-builder introspect --sanitize test/data/

Sanitized dependencies for test/data/
python:
- pyvcloud>=14,>=18.0.10
- pytz
- tacacs_plus
system:
- test/bindep/bindep.txt
```

In terms of surface area between parts, and to the user / other developers, this is pretty big.

I want to file an issue to leave comments about where requirement lines came from (in terms of collections).

ansible_builder/main.py
```diff
@@ -185,6 +186,8 @@ def create_folder_copy_files(self):
             requirement_path = self.definition.get_dep_abs_path(item)
             if requirement_path is None:
                 continue
+            if item == 'python':
+                continue  # will be put into combined requirements.txt
```
AlanCoding, line 190: I _think_ my preference here would be to, instead, read the file and save it to an instance property. Basically, move the logic below at `with open(self.definition.get_dep_abs_path('python'), 'r') as f:` to here.

But... I don't want to be too specific about it right now. Since this is working, I somewhat prefer to finish the processing of collection `bindep.txt` files, see how that code change looks, and change some of the processing order once we have a better sense of that flow.

AlanCoding:
This is adding the dependency:

https://github.com/davidfischer/requirements-parser

It doesn't do _much_ other than simple string parsing. But, I can't find any other semi-reliable means of splitting out the version constraints. I still don't want to _process_ version constraints, but I do want to _combine_ them. You see that reflected in tests here.

AlanCoding:
Future enhancement request:

I would like to add comments to each line in the combined requirements file which says which collection(s) the requirement came from. This will involve more changes to the schema of the JSON from introspect.

ansible_builder/introspect.py
```diff
@@ -5,15 +5,31 @@
...
+        for line in f.read().split('\n'):
+            if not line:
+                continue
+            if line.startswith('-r') or line.startswith('--requirement'):
+                _, new_filename = line.split(None, 1)
+                new_path = os.path.join(os.path.dirname(path or '.'), new_filename)
+                req_list.extend(pip_file_data(new_path))
+            else:
+                req_list.append(line)
+    return req_list
```
AlanCoding, line 25: My first attempt looked nothing like this:

https://stackoverflow.com/questions/49689880/proper-way-to-parse-requirements-file-after-pip-upgrade-to-pip-10-x-x/59969843#comment111716393_59969843

I found importing from `pip` to be pretty unusable. Maybe there's something I didn't get, but...

https://github.com/davidfischer/requirements-parser/blob/ad404add0423d7e1f419c3a48dc884375d57f376/requirements/parser.py#L33-L39

This is how the `requirements-parser` library handles these cross-file cases. It seemed relatively uncomplicated, and there's not much wiggle room in the allowed syntax to put in more bizarre forms of references.

ansible_builder/cli.py
```diff
@@ -97,6 +104,13 @@ def parse_args(args=sys.argv[1:]):
         )
     )
     add_introspect_options(introspect_parser)
+    introspect_parser.add_argument(
+        '--sanitize', help=(
+            'Sanite and de-duplicate requirements. '
```
beeankha: `Sanite` --> `Sanitize`

beeankha (approved):

shanemcd (approved):

Issue #50, opened 2020-07-24
Process system requirements


AlanCoding:
for my own notes, from this comment:

https://github.com/ansible/ansible-builder/issues/22#issuecomment-650206193

https://opendev.org/opendev/system-config/src/branch/master/docker/python-builder/scripts/assemble

the critical syntax looks like:

```
bindep -l newline >> /output/bindep/run.txt || true
yum -y install $(cat /output/bindep/run.txt)
```

(we may or may not use the intermediary file)

In my testing, the `-q` option (for quiet) is also needed, and very important, a newline is needed at the end of the `bindep.txt` file.

This will be implemented as a `RUN` in the image.

AlanCoding:
This is implemented for a user's file, but not yet for collecting bindep files from collections themselves.

AlanCoding:
Also need to update the `README.md`

AlanCoding:
Updating and documenting this issue...

We merged #52 which addressed this for user-specified bindep file, meaning the `execution-environment.yml` file had an entry `system: bindep.txt` in it. The WIP #59 addresses it for collections. Here is where I'm at with that, the flow loosely reflects the order in which my own thought process shaped up.

The aggressive proposal is most similar to how the similar problem for python requirements was addressed in #57. However, I would, right away, push _against_ this option because bindep stuff has fairly different issues from python stuff. We for bindep, we only really need to detect if the _final result_ is blank or not blank. Including entries which are obviously not relevant isn't a problem, and it's kind of part of the design. Not mentioned here - we may still like to scrub a combined `bindep.txt` file for _literal duplicates_, because this case will occur commonly in the existing collections, like `python3-devel [test platform:rpm]`, which is copy+pasted for several collections like `frr.frr`, `arista.eos`, and many others.

My preference is clearly the in-between proposal.

#### Current PR:

 - in introspect script, exclude lines which are blank or have "test" in them, only "count" bindep.txt if it's not blank after that
 - introspect script returns `bindep.txt` file locations
 - Do a single-line RUN layer in Containerfile for each `bindep.txt` file from user or collections
 - (this can still error if bindep reqs are not relevant to platform)

#### Agressive proposal:

 - in introspect script, only remove obviously blank lines or comment lines
 - introspect script returns lines from `bindep.txt` files
 - Add bindep as a requirement of `ansible-builder` itself
 - import functions from bindep in order to scrub the requirements, remove those which are obviously N/A, test profile in particular
[...]

PR #59, opened 2020-07-31
Initial processing of bindep.txt from collections

AlanCoding:
Connect #50

This still needs more cleanup, but I want people to know where I'm going.

This corresponds to a change to my `ovirt.ovrit` branch which adds the `bindep.txt` of:

```
gcc [platform:rpm]
python3-devel [platform:rpm]
libcurl-devel [platform:rpm]
openssl-devel [platform:rpm]
libxml2-devel [platform:rpm]
```

This produces the `Dockerfile`

```
FROM shanemcd/ansible-runner

RUN pip3 install --upgrade pip setuptools
ENV PYCURL_SSL_LIBRARY=openssl
ADD introspect.py /usr/local/bin/introspect
RUN chmod +x /usr/local/bin/introspect
ADD requirements.yml /build/

RUN ansible-galaxy role install -r /build/requirements.yml --roles-path /usr/share/ansible/roles
RUN ansible-galaxy collection install -r /build/requirements.yml --collections-path /usr/share/ansible/collections
RUN pip3 install bindep
RUN dnf -y install $(bindep -b -f /usr/share/ansible/collections/ansible_collections/ovirt/ovirt/bindep.txt)
ADD requirements.txt /build/
RUN pip3 install --upgrade -r /build/requirements.txt
```

For the awx case with the definition:

```
---
version: 1
dependencies:
  galaxy: requirements.yml
additional_build_steps:
  prepend:
    - RUN pip3 install --upgrade pip setuptools
    - ENV PYCURL_SSL_LIBRARY=openssl
```

And with that, it does run to completion with all the python requirements (particularly pycurl) installing.

### DIFF
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -8,7 +8,7 @@ from .colors import MessageColors
 from .exceptions import DefinitionError
 from .main import AnsibleBuilder
 from . import constants
-from .introspect import add_introspect_options, process
+from .introspect import add_introspect_options, process, write_files, simple_combine
 from .requirements import sanitize_requirements
 
 
@@ -25,17 +25,17 @@ def run():
             print(e.args[0])
             sys.exit(1)
     elif args.action == 'introspect':
-        for folder in args.folders:
-            data = process(folder)
-            if args.sanitize:
-                data['python'] = sanitize_requirements(data['python'])
-                print()
-                print('Sanitized dependencies for {0}'.format(folder))
-                print(yaml.dump(data, default_flow_style=False))
-            else:
-                print()
-                print('Dependency data for {0}'.format(folder))
-                print(yaml.dump(data, default_flow_style=False))
+        data = process(args.folder, user_pip=args.user_pip, user_bindep=args.user_bindep)
+        if args.sanitize:
+            data['python'] = sanitize_requirements(data['python'])
+            data['system'] = simple_combine(data['system'])
+            print()
+            print('Sanitized dependencies for {0}'.format(args.folder))
+        else:
+            print()
+            print('Dependency data for {0}'.format(args.folder))
+        print(yaml.dump(data, default_flow_style=False))
+        write_files(data, write_pip=args.write_pip, write_bindep=args.write_bindep)
         sys.exit(0)
 
     print(MessageColors.FAIL + "An error has occured." + MessageColors.ENDC)
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -8,7 +8,7 @@ from .colors import MessageColors
 from .exceptions import DefinitionError
 from .main import AnsibleBuilder
 from . import constants
-from .introspect import add_introspect_options, process, write_files, simple_combine
+from .introspect import add_introspect_options, process, simple_combine
 from .requirements import sanitize_requirements
 
 
@@ -25,17 +25,17 @@ def run():
             print(e.args[0])
             sys.exit(1)
     elif args.action == 'introspect':
-        data = process(args.folder, user_pip=args.user_pip, user_bindep=args.user_bindep)
+        data = process(args.folder)
         if args.sanitize:
             data['python'] = sanitize_requirements(data['python'])
             data['system'] = simple_combine(data['system'])
             print()
-            print('Sanitized dependencies for {0}'.format(args.folder))
+            print('# Sanitized dependencies for {0}'.format(args.folder))
         else:
             print()
-            print('Dependency data for {0}'.format(args.folder))
+            print('# Dependency data for {0}'.format(args.folder))
+        print('---')
         print(yaml.dump(data, default_flow_style=False))
-        write_files(data, write_pip=args.write_pip, write_bindep=args.write_bindep)
         sys.exit(0)
 
     print(MessageColors.FAIL + "An error has occured." + MessageColors.ENDC)
--- a/ansible_builder/introspect.py
+++ b/ansible_builder/introspect.py
@@ -10,9 +10,6 @@ import argparse
 base_collections_path = '/usr/share/ansible/collections'
 default_file = 'execution-environment.yml'
 
-begin_delimiter = '----begin_introspect_output----'
-end_delimiter = '----end_introspect_output----'
-
 
 def line_is_empty(line):
     return bool((not line.strip()) or line.startswith('#'))
@@ -196,19 +193,6 @@ def write_files(data, write_pip=None, write_bindep=None):
             f.write('\n'.join(simple_combine(data.get('system')) + ['']))
 
 
-def parse_introspect_output(stdout):
-    p = re.compile(
-        '{0}(?P<yaml_text>.+){1}'.format(begin_delimiter, end_delimiter),
-        flags=re.MULTILINE | re.DOTALL
-    )
-    m = p.search(stdout)
-    if m is None:
-        return None
-    yaml_text = m.group('yaml_text').strip()
-    data = yaml.safe_load(yaml_text)
-    return data
-
-
 def add_introspect_options(parser):
     parser.add_argument(
         'folder', default=base_collections_path, nargs='?',
@@ -246,8 +230,6 @@ if __name__ == '__main__':
     add_introspect_options(parser)
     args = parser.parse_args()
     data = process(args.folder, user_pip=args.user_pip, user_bindep=args.user_bindep)
-    print(begin_delimiter)
     print(yaml.dump(data, default_flow_style=False))
-    print(end_delimiter)
     write_files(data, write_pip=args.write_pip, write_bindep=args.write_bindep)
     sys.exit(0)
--- a/ansible_builder/introspect.py
+++ b/ansible_builder/introspect.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import os
+import re
 import sys
 import yaml
 import argparse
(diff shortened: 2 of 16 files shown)


ansible_builder/cli.py
```diff
@@ -25,17 +25,17 @@ def run():
...
-            if args.sanitize:
-                data['python'] = sanitize_requirements(data['python'])
-                print()
-                print('Sanitized dependencies for {0}'.format(folder))
-                print(yaml.dump(data, default_flow_style=False))
-            else:
-                print()
-                print('Dependency data for {0}'.format(folder))
-                print(yaml.dump(data, default_flow_style=False))
+        data = process(args.folder, user_pip=args.user_pip, user_bindep=args.user_bindep)
```
AlanCoding: Changed `args.folders` to `args.folder` - because the original idea was that we would introspect all the default Ansible collection search paths. This is unlikely to ever be supported. It's also very hard to reconcile with combining with the user-provided files.

That's a major change - before, we combined the user's content with the collection content _after_ the introspect script. Now, I am moving the user's files into the image and having the introspect script include that content in its output.

ansible_builder/introspect.py
```diff
@@ -125,14 +163,76 @@ def get_dependency(self, entry):
...
+        with open(write_pip, 'w') as f:
+            f.write('\n'.join(simple_combine(data.get('python')) + ['']))
+    if write_bindep and data.get('system'):
+        with open(write_bindep, 'w') as f:
+            f.write('\n'.join(simple_combine(data.get('system')) + ['']))
+
+
+def parse_introspect_output(stdout):
+    p = re.compile(
+        '{0}(?P<yaml_text>.+){1}'.format(begin_delimiter, end_delimiter),
```
AlanCoding: Added regex processing here.

Why:

Before, we called `docker run -rm <tag> introspect`. When we did that, the stdout constituted a full YAML document (usually).

Here, we changed `introspect` to be inside a `RUN` layer, as opposed to a command ran outside of the Containerfile. That means that its output is interlaced in with the rest of the build command output. Instead of running more unnecessary commands, I decided I would just use markers to extract it.

ansible_builder/introspect.py
```diff
@@ -125,14 +163,76 @@ def get_dependency(self, entry):
...
         )
     )
+    parser.add_argument(
+        '--user-pip', dest='user_pip',
+        help='An additional file to combine with collection pip requirements.'
+    )
+    parser.add_argument(
+        '--user-bindep', dest='user_bindep',
+        help='An additional file to combine with collection bindep requirements.'
+    )
```
AlanCoding: Again, this is changing it so that the introspect script combines the user's `requirements.txt` and `bindep.txt` files with those files inside of collections. So that gets added to the contract for the introspect command too.

```
$ ansible-builder introspect test/data/ --user-pip test/data/pip/project/requirements.txt 

Dependency data for test/data/
python:
  test.metadata:
  - pyvcloud>=14
  test.reqfile:
  - pytz
  - tacacs_plus
  - 'pyvcloud>=18.0.10  # duplicate with test.metadata collection'
  user:
  - bottle>=3.0
system:
  test.bindep:
  - subversion [platform:rpm]
  - subversion [platform:dpkg]
```

shows that it is picking up

```
$ cat test/data/pip/project/requirements.txt
bottle>=3.0
```

ansible_builder/introspect.py
```diff
@@ -125,14 +163,76 @@ def get_dependency(self, entry):
...
+    parser.add_argument(
+        '--write-pip', dest='write_pip',
+        help='Write the combined bindep file to this location.'
+    )
+    parser.add_argument(
+        '--write-bindep', dest='write_bindep',
+        help='Write the combined bindep file to this location.'
+    )
```
AlanCoding: Not using `--write-pip`, but added it for consistency.

We have the introspect script write the bindep file because of technicalities of how bindep operates. The first phase needs to have written a file ready to pass to `bindep`

ansible_builder/main.py
```diff
@@ -1,20 +1,25 @@
...
-CONTEXT_FILES = ['galaxy', 'python', 'system']
+# Files that need to be moved into the build context, and their naming inside the context
+CONTEXT_FILES = {
+    'galaxy': 'requirements.yml',
+    'python': 'requirements_user.txt',
+    'system': 'bindep_user.txt'
+}
+BINDEP_COMBINED = 'bindep_combined.txt'
+BINDEP_OUTPUT = 'bindep_output.txt'
+PIP_COMBINED = 'requirements_combined.txt'
```
AlanCoding, line 20: I changed the naming of most things that go into the build context folder.

Before, we accepted the naming from the user.... I'd rather not, because these go through our own phases of combination, sanitization, and transformation. I want these to be as clear as possible so that we can document them.

ansible_builder/main.py
```diff
@@ -49,18 +54,39 @@ def build_command(self):
             self.build_context
         ]
 
+    def run_in_container(self, command, **kwargs):
+        wrapped_command = [
+            self.container_runtime, 'run',
+            '--rm', self.tag, '/bin/bash', '-c'
+        ] + [' '.join(command)]
+        return run_command(wrapped_command, **kwargs)
```
AlanCoding, line 62: I had to change this to the `-c` pattern so I could run more than 1 command with a single invocation.

ansible_builder/main.py
```diff
@@ -49,18 +54,39 @@ def build_command(self):
...
+        if collection_data is None:
+            # In case the build has already been ran once, a cached layer may be used for introspect
+            rc, output = self.run_in_container(['introspect'], capture_output=True)
+            collection_data = ansible_builder.introspect.parse_introspect_output('\n'.join(output))
+        if collection_data.get('system'):
+            rc, output = self.run_in_container(
+                [
+                    'pip3', 'install', 'bindep;',
+                    'echo', '----begin_bindep_output----;',
+                    'bindep', '-b', '-f', '/build/{0}'.format(BINDEP_COMBINED)],
```
AlanCoding: Here, I both install and run bindep in a new container. This keeps bindep out of the python environment in the image, see https://github.com/ansible/ansible-builder/pull/76#issuecomment-674635447

ansible_builder/main.py
```diff
@@ -191,21 +217,12 @@ def create_folder_copy_files(self):
...
-            exists = os.path.exists(dest)
-            do_copy = True
-            if exists:
-                do_copy = not filecmp.cmp(requirement_path, dest, shallow=False)
-                if do_copy:
-                    print(MessageColors.WARNING + 'File {} had modifications and will be rewritten'.format(dest) + MessageColors.ENDC)
-            if do_copy:
-                shutil.copy(requirement_path, self.build_context)
+            dest = os.path.join(self.build_context, new_name)
+            copy_file(requirement_path, dest)
```
AlanCoding, line 240: Coding practices for copy/write had diverged in a few places, so I tried to fully finish shared methods for `copy_file` and `write_file`

ansible_builder/main.py
```diff
@@ -228,43 +245,41 @@ def prepare_appended_steps(self):
...
-        data = yaml.safe_load("\n".join(output))
+    def prepare_pip_steps(self, collection_pip):
+        pip_list = sanitize_requirements(collection_pip)
 
-        if python_req_file:
-            with open(self.definition.get_dep_abs_path('python'), 'r') as f:
-                user_py_reqs = f.read().split('\n')
-            data['python'].extend(user_py_reqs)
-        data['python'] = sanitize_requirements(data['python'])
-        pip_file = os.path.join(self.build_context, 'requirements.txt')
```
AlanCoding, line 255: Combining user's `requirements.txt` file was moved earlier in the process (phase 1 as opposed to phase 2), so that's why it's axed here. This was a phase 2 thing.

ansible_builder/requirements.py
```diff
@@ -14,24 +14,28 @@
...
+                    for prior_req in consolidated:
+                        if req.name == prior_req.name:
+                            prior_req.specs.extend(req.specs)
+                            prior_req.collections.append(collection)
+                            break
+                    continue
+                consolidated.append(req)
+                seen_pkgs.add(req.name)
+        except Exception as e:
+            print('Warning: failed to parse requirments from {}, error: {}'.format(collection, e))
```
AlanCoding, line 38: Added try-except here so that we would always get a result as opposed to erring.

I don't think a single bad-actor collection should halt the entire process. If a collection has stuff that cannot be parsed, we don't want to include any of its requirements, because that's a red flag.

ansible_builder/requirements.py
```diff
@@ -40,11 +44,13 @@ def sanitize_requirements(py_reqs):
...
-            sanitized.append(req.line)
+            new_line = req.line
         elif req.name:
             specs = ['{0}{1}'.format(cmp, ver) for cmp, ver in req.specs]
-            sanitized.append(req.name + ','.join(specs))
+            new_line = req.name + ','.join(specs)
         else:
             raise RuntimeError('Could not process {0}'.format(req.line))
 
+        sanitized.append(new_line + '  # from collection {}'.format(','.join(req.collections)))
```
AlanCoding, line 54: Added comments to denote the source of requirements.

ansible_builder/introspect.py
shanemcd: I don't love this. What was the reasoning here?

ansible_builder/main.py
shanemcd: For downstream we'll need to have bindep in the base image, so we should probably do that upstream too.

ansible_builder/introspect.py
AlanCoding: Actually, https://github.com/ansible/ansible-builder/issues/78 factored into my thinking.

That issue wasn't important enough to act on, but then several other changes started getting thrown into the mix where `introspect` and `bindep` get thrown into a series of multiple commands. It became easier to assume that the output will get "padded" with other text, because it's not a hard problem to solve.

ansible_builder/main.py
```diff
@@ -49,18 +52,39 @@ def build_command(self):
             self.build_context
         ]
 
+    def run_in_container(self, command, **kwargs):
+        wrapped_command = [self.container_runtime, 'run','--rm']
+
+        wrapped_command.extend(['-v', f"{os.path.abspath(self.build_context)}:/context:Z"])
```
shanemcd, line 58: If we're hardcoding this here, we can remove `volumes` from the kwargs passed to the invocations below.

ansible_builder/main.py
```diff
@@ -49,18 +52,39 @@ def build_command(self):
...
         self.containerfile.write()
-        run_command(self.build_command)
-        print(MessageColors.OK + 'Rewriting Containerfile to capture collection requirements' + MessageColors.ENDC)
-        self.containerfile.prepare_system_steps()
-        self.containerfile.prepare_pip_steps()
+        rc, output = run_command(self.build_command, capture_output=True)
+
+        rc, output = self.run_in_container(
+            ['python3', '/context/introspect.py', '--write-bindep', '/context/bindep_combined.txt'],
+            volumes=[f"{os.path.abspath(self.build_context)}:/context:Z"],
```
shanemcd: This can go

ansible_builder/main.py
```diff
@@ -49,18 +52,39 @@ def build_command(self):
...
+            capture_output=True)
+
+        collection_data = yaml.safe_load('\n'.join(output))
+        if collection_data.get('system'):
+            rc, output = self.run_in_container(
+                ['bindep', '-b', '-f', '/context/{0}'.format(BINDEP_COMBINED)],
+                volumes=[f"{os.path.abspath(self.build_context)}:/context:Z"],
```
shanemcd: and this

AlanCoding:
I was eager to rebase, and wound up pushing this in a really half-implemented form

AlanCoding:
This should be ready for review after it is rebased after the merge of #88

AlanCoding:
Now this has been rebased, tests are passing, and hopefully all the dust has settled. Ready for review

ansible_builder/main.py
```diff
@@ -49,18 +52,55 @@ def build_command(self):
...
+        python_lines = sanitize_requirements(collection_data['python'])
+
+        bindep_output = []
+        if system_lines:
+            write_file(os.path.join(self.build_context, BINDEP_COMBINED), system_lines + [''])
+
+            rc, bindep_output = self.run_in_container(
+                ['bindep', '-b', '-f', '/context/{0}'.format(BINDEP_COMBINED)],
+                allow_error=True, capture_output=True
+            )
```
AlanCoding, line 85: hmm.... should we _delete_ the `BINDEP_COMBINED` file here? I learn towards "no" but feel like I have to ask the question.

ansible_builder/main.py
AlanCoding: Yes, did that. I wanted this to be an assumption about running in the container.

One more thing to note - `abspath` make the syntax more verbose

```
docker run --rm -v /Users/alancoding/Documents/repos/ansible-builder/test/data/bc:/context:Z exec-env-pycurl /bin/bash -c python3 /context/introspect.py
```

I don't know that it's necessary. It's clearly not broken, so I don't want to touch it now, just submitting it for consideration. It could be the case that it has `..` in it, so if we got rid of this, maybe we would have to resolve the relative path, which might take a little more legwork.

shanemcd (approved):

AlanCoding:
regate

shanemcd:
regate

AlanCoding:
regate

PR #61, opened 2020-08-03
give sensible error when provide wrong type

kdelee:
we exepect a dictionary with subkeys here, if it is a multiline string
as was given in an example it blows up in not so obvious way.

found this when looking at https://github.com/ansible/ansible-builder/issues/21

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -204,8 +204,8 @@ class Containerfile:
         if additional_prepend_steps:
             if not isinstance(additional_prepend_steps, dict):
                 raise TypeError((
-                    "Expected 'additional_build_steps' to be a dictionary with keys 'prepend' and/or 'append',\n"
-                    f"found a {type(additional_prepend_steps)} instead"
+                    MessageColors.FAIL + "Expected 'additional_build_steps' to be a dictionary with keys 'prepend' and/or 'append',\n"
+                    f"found a {type(additional_prepend_steps)} instead" + MessageColors.ENDC
                 ))
             prepended_steps = additional_prepend_steps.get('prepend')
             if prepended_steps:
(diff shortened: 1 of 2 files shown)


ansible_builder/main.py
```diff
@@ -210,6 +216,11 @@ def prepare_prepended_steps(self):
     def prepare_appended_steps(self):
         additional_append_steps = self.definition.get_additional_commands()
         if additional_append_steps:
+            if not isinstance(additional_append_steps, dict):
+                raise TypeError((
+                    MessageColors.FAIL + "Expected 'additional_build_steps' to be a dictionary with keys 'prepend' and/or 'append',\n"
+                    f"found a {type(additional_append_steps)} instead" + MessageColors.ENDC
+                ))
```
beeankha: I'm not sure this needs to be in here... prepend is always the first thing that runs, and if the syntax in the definition file is:
```
additional_build_steps: |
    - RUN echo Oh hey this is an additional command!
    - RUN ls -la /etc
``` 
then `prepare_prepended_steps()` will always catch this, I think...

ansible_builder/main.py
```diff
@@ -126,6 +126,7 @@ def get_dep_abs_path(self, entry):
         if not req_file:
             return None
 
+        print(req_file)
```
kdelee: woops need to delete this

beeankha:
When incorrect syntax in `additional_build_steps` is provided (in this example I made it a list), now the output looks like this:

Screen Shot 2020-08-05 at 11 29 04 AM

kdelee:
@AlanCoding as much as I was the one who started down this path, and am good with changes...I wonder if instead we want to have some pre-validation of data types of each section instead of handling it in these lower level funtions. what do you think

AlanCoding:
> I wonder if instead we want to have some pre-validation of data types of each section instead of handling it in these lower level funtions. what do you think

I started down the path of a `.validate` method

https://github.com/ansible/ansible-builder/blob/fa11a62da6e572a869c07a392973b4381e0d1053/ansible_builder/main.py#L142

The core ideas are

1. it gets called early, meaning, before any _file writes_ take place, so that errors here will be clean errors
2. our unit tests just machine-gun rapid fire cases of wrongly formatted definitions
3. it raises `DefinitionError`, with a message of sufficient detail, and the cli module writes that & exits

AlanCoding:
Put up those revisions at https://github.com/kdelee/ansible-builder/pull/1

AlanCoding (approved):

Issue #56, opened 2020-07-29
Establish pattern for running integration tests by invoking script that outputs xml test result


kdelee:
We want to be able to execute integration tests from the testing pipeline/provider of our choice.

To make our integration tests easy to execute from whatever location, and consume the results, have both the test setup and test execution wrapped in a bash script that after the test run, leaves a junit style xml file as output.

Suggested implementation:

1) create a folder for integration tests that you can not run during normal unit tests
2) In that folder, create a `conftest.py` file with a fixture like this (may take some small changes to work here) [complete proxy cli fixture from runner](https://github.com/ansible/ansible-runner/blob/devel/test/integration/containerized/conftest.py#L21-L70)
3) create a test file and that file create a test that does something simple like:

psuedocode:
```
def test_help(cli):
   result = cli(['ansible-builder', 'help'])
   assert 'something that is in help text' in result.stdout
```

4) now that you have a test, create a shell script that 
   - does any python installs/environment setup necessary to run tests
   - calls pytest, passing to it path to the integration test folder and passes flag `--junitxml=path/to/results.xml`. Docs: https://docs.pytest.org/en/latest/usage.html#creating-junitxml-format-files

Acceptance criteria:
  - [ ] user/CI only needs to call `tools/run_integration.sh` or whatever name and script takes care of any dependency installs, runs tests, and leaves behind a `artifacts/results.xml` file that is always named the same thing

In the future more tests can be added, and this will be transparent to whatever CI runner is running the tests. If there are system requirements this is assumed to be taken care of by the CI runner, so please just document these dependencies.

Optional additional features:
- [ ] accept RUNNER_FORK and RUNNER_BRANCH environment variables to chose what runner to install from git
- [ ] accept TESTEXPR environment variable to pass to pytest `-k ` test selector option

kdelee:
@beeankha this is what we talked about

AlanCoding:
@kdelee is this already running and producing the junit output? We need to build out test cases, but that's not the same issue as this.

kdelee:
yes we can close

PR #64, opened 2020-08-04
add basic pattern for integration tests

kdelee:
@beeankha this is my first stab at getting tests running in a way that will work with our jenkins pipeline

I originally had suggested a script to call, but given that the only install method is pip I think tox is fine for now. I can add tox as a system dep to install prior to launching actual tests and tox will take care of setting up poetry and python deps etc.

I have included a few tests and I would like to add more stubs describing test cases we'd like to implement.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -34,6 +34,14 @@ version = "0.7.12"
 
 [[package]]
 category = "dev"
+description = "apipkg: namespace control and lazy-import mechanism"
+name = "apipkg"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+version = "1.5"
+
+[[package]]
+category = "dev"
 description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
 name = "appdirs"
 optional = false
@@ -401,6 +409,20 @@ name = "docutils"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 version = "0.16"
+
+[[package]]
+category = "dev"
+description = "execnet: rapid multi-Python deployment"
+name = "execnet"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+version = "1.7.1"
+
+[package.dependencies]
+apipkg = ">=1.4"
+
+[package.extras]
+testing = ["pre-commit"]
 
 [[package]]
 category = "dev"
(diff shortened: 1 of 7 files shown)


kdelee:
this is for https://github.com/ansible/ansible-builder/issues/56

poetry.lock
```diff
@@ -822,6 +844,35 @@ version = ">=0.12"
...
+py = "*"
+pytest = ">=3.10"
+
+[[package]]
+category = "dev"
+description = "pytest xdist plugin for distributed testing and loop-on-failing modes"
+name = "pytest-xdist"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+version = "1.34.0"
```
kdelee, line 865: all the changes here to python deps is because I added pytest-xdist to enable running pytest in parallel

test/integration/test_build.py
```diff
@@ -0,0 +1,27 @@
...
+def test_missing_galaxy_requirements_file():
+    """If a user specifies a galaxy requirements file, but we can't find it, fail sanely."""
+    pytest.skip("Not implemented")
+
+
+def test_build_streams_output(cli, build_dir_and_ee_yml, ee_tag):
+    """Test that 'ansible-builder build' streams build output."""
+    tmpdir, eeyml = build_dir_and_ee_yml("")
+    result = cli(f"ansible-builder build -c {tmpdir} -f {eeyml} -t {ee_tag}")
+    assert f"podman build -f {tmpdir}/Containerfile" in result.stdout
```
AlanCoding: It would be a big help for several people to make it configurable podman/docker, because we have to spin up a VM or something to use podman. This will change as we phase out Mac laptops, but for the moment, it's been helpful to get going fast by swapping the two.

tox.ini
```diff
@@ -14,7 +14,7 @@ deps = poetry >= 1.0.5
 commands =
     poetry install
     poetry env info
-    poetry run py.test -v test
+    poetry run py.test -v test -k "not integration"
```
AlanCoding, line 17: this is clever. I had assumed these would get split up into `test/unit` and `test/integration`, but clearly this works.

test/integration/test_build.py
AlanCoding: This is going to leave around images isn't it? I think we need some additional layer in `conftest.py` on top of what you have to run a build command, so that it assigns a randomized tag, and makes sure to remove the tag in cleanup.

What you have is great, and I see that making sense as a separate work item.

test/integration/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+import subprocess
+
+import tempfile
+import uuid
+
+import pytest
+
+
+@pytest.fixture
+def build_dir_and_ee_yml():
```
AlanCoding, line 11: it's not out of the question that this could be shared with the existing unit tests. As we get into the think of things, that's what I would prefer.

test/integration/test_build.py
kdelee: yes we can do something around that. Was going to probably make a fixture that detects available runtimes, and parametrizes this on that

AlanCoding (approved):
If @beeankha doesn't have any other objections, I would be fine with merging this and building on from there.

test/integration/conftest.py
kdelee: cool can move it up a level if you want to share this fixture

test/integration/test_build.py
AlanCoding: that sounds like more work than I had in mind.

Even if you just read from an (optional) environment variable, I think that would solve the problem.

AlanCoding: Using docker isn't really supported, but it's going to get developed against. That's kind of the split here in a nutshell.

kdelee: ack

beeankha (approved):
This looks great, thank you @kdelee !

kdelee:
should be able to merge after lint fixes

PR #65, opened 2020-08-04
Remove Unnecessary Return Code Check

beeankha:
In the case where Galaxy requirements are not present, the introspect command provides empty lists vs a non-zero return code; removing the `rc` check in the `prepare_pip_steps` method in `main.py` to match up with that expectation.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -5,7 +5,7 @@ import pytest
 
 @pytest.fixture(autouse=True)
 def do_not_run_commands():
-    cmd_mock = mock.MagicMock(return_value=[1, []])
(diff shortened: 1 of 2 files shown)


test/conftest.py
```diff
@@ -5,7 +5,7 @@
 
 @pytest.fixture(autouse=True)
 def do_not_run_commands():
-    cmd_mock = mock.MagicMock(return_value=[1, []])
+    cmd_mock = mock.MagicMock(return_value=[1, ['python:', '  - foo', 'system: []']])
```
AlanCoding, line 8: I would like to expand this, but I think it's okay for right now

AlanCoding (approved):

PR #71, opened 2020-08-10
Reformat Definition File-Related Error Messages, Update README

beeankha:
Updated the `DefinitionError` custom class to no longer show a traceback (plus added color to the output error messages), edited the README to show more specific definition file-related information.

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -92,10 +92,10 @@ class UserDefinition(BaseDefinition):
                 y = yaml.safe_load(f)
                 self.raw = y if y else {}
         except FileNotFoundError:
-            raise DefinitionError("""
+            raise DefinitionError(textwrap.dedent("""
             Could not detect '{0}' file in this directory.
             Use -f to specify a different location.
-            """.format(filename))
+            """).format(filename))
         except yaml.parser.ParserError as e:
             raise DefinitionError(textwrap.dedent("""
             An error occured while parsing the definition file:
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -1,6 +1,7 @@
 import filecmp
 import os
 import shutil
+import sys
 import textwrap
 import yaml
 
@@ -17,6 +18,7 @@ CONTEXT_FILES = ['galaxy', 'python', 'system']
 
 
 class DefinitionError(RuntimeError):
+    sys.tracebacklimit = 0
     pass
 
 
@@ -95,10 +97,10 @@ class UserDefinition(BaseDefinition):
                 y = yaml.safe_load(f)
                 self.raw = y if y else {}
         except FileNotFoundError:
-            raise DefinitionError("""
+            raise DefinitionError(MessageColors.FAIL + """
             Could not detect '{0}' file in this directory.
             Use -f to specify a different location.
-            """.format(constants.default_file))
+            """.format(constants.default_file) + MessageColors.ENDC)
         except yaml.parser.ParserError as e:
             raise DefinitionError(textwrap.dedent(MessageColors.FAIL + """
             An error occured while parsing the definition file:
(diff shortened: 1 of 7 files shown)


AlanCoding:
Isn't this already wrapped in the fail colors here?

https://github.com/ansible/ansible-builder/blob/b11316b442b45c8fbe4f60cc6004ad11f91dccc1/ansible_builder/cli.py#L24

I would rather do it in 1 place rather than every place this exception is raised. Sounds like this issue could be some bug in how colors are applied in this line?

beeankha:
Discovered that when the `-f` flag pointed at a file that didn't exist, the error message that comes up would list the default definition file name vs the specified file name, like so:

Screen Shot 2020-08-10 at 3 59 22 PM

Fixed this, now the correct error message comes up:

Screen Shot 2020-08-10 at 3 59 33 PM


When no filename is specified and the default definition file isn't found, the error message still correctly points to the default name:

Screen Shot 2020-08-10 at 4 01 25 PM

ansible_builder/main.py
```diff
@@ -17,8 +18,13 @@
 
 
 class DefinitionError(RuntimeError):
+    sys.tracebacklimit = 0
     pass
```
shanemcd: Should this `pass` get removed now?

ansible_builder/main.py
```diff
@@ -17,8 +18,13 @@
 
 
 class DefinitionError(RuntimeError):
+    sys.tracebacklimit = 0
```
shanemcd: What is this doing?

beeankha: It's eliminating the output of traceback before our custom error message prints out.

ansible_builder/main.py
beeankha: Removing!

ansible_builder/main.py
```diff
@@ -98,17 +103,15 @@ def __init__(self, filename):
             raise DefinitionError("""
             Could not detect '{0}' file in this directory.
             Use -f to specify a different location.
-            """.format(constants.default_file))
+            """.format(filename))
```
AlanCoding: Ah, is this the reason that it wasn't printing the right filename?

AlanCoding (approved):

ansible_builder/main.py
beeankha: Yep!  It was always filling that line with the default name vs what was specified (_if_ something was specified).

ansible_builder/steps.py
```diff
@@ -21,8 +21,11 @@ def __init__(self, additional_steps):
         elif isinstance(additional_steps, list):
             lines = additional_steps
         else:
-            print(MessageColors.FAIL + "Error: Unknown type found for additional_build_steps; "
-                  "must be list or multi-line string." + MessageColors.ENDC)
+            raise DefinitionError(
+                """
+                Error: Unknown type found for additional_build_steps; must be list or multi-line string.
+                """
+            )
```
AlanCoding, line 28: I don't think you necessarily need to keep this conditional, because it should be covered by the validate method.

But if you do, feel free to introduce new module `ansible_builder.exceptions` and put the definition of `DefinitionError` there so that you eliminate any possibility of circular imports. You really want to keep python imports without loops, and in-line imports are something you should avoid whenever you have the freedom to avoid them.

beeankha: Makes sense, I'll do that!  

Just noticed I forgot to remove the unused `from .colors import MessageColors` from `steps.py` as well, so I'll be pushing another commit shortly.

shanemcd:
One last thing: It looks like some unintentional indentation is going on. We probably want to wrap the multi-line strings in a `textwrap.dedent`.

shanemcd (approved):

PR #44, opened 2020-07-09
Significantly build out examples

AlanCoding:
I don't quite know how we want to handle this, but we _can_ merge it in its current form with no ill effects because we don't run these in any form of CI. I'm going to enumerate the ones that do not work:

 - awx
 - subversion
 - ansible.posix.at

The reason they don't work is because they are tied into active feature development. However, they fail in some exciting ways, here is a taste:

#### subversion

```
PLAY [localhost] ***************************************************************

TASK [Checkout subversion repository to specified folder] **********************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to find required executable svn in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}

PLAY RECAP *********************************************************************
localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   

```

#### at

```
PLAYBOOK: at.yml ***************************************************************
1 plays in at.yml

PLAY [localhost] ***************************************************************
META: ran handlers

TASK [Schedule a command to execute in 20 minutes as root] *********************
task path: /runner/project/at.yml:7
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: root
<127.0.0.1> EXEC /bin/sh -c 'echo ~root && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /root/.ansible/tmp `"&& mkdir /root/.ansible/tmp/ansible-tmp-1594325018.8054106-19-129616479202610 && echo ansible-tmp-1594325018.8054106-19-129616479202610="` echo /root/.ansible/tmp/ansible-tmp-1594325018.8054106-19-129616479202610 `" ) && sleep 0'
Using module file /usr/share/ansible/collections/ansible_collections/ansible/posix/plugins/modules/at.py
<127.0.0.1> PUT /runner/.ansible/tmp/ansible-local-13sjhkdn6f/tmpyxel_tk7 TO /root/.ansible/tmp/ansible-tmp-1594325018.8054106-19-129616479202610/AnsiballZ_at.py
<127.0.0.1> EXEC /bin/sh -c 'chmod u+x /root/.ansible/tmp/ansible-tmp-1594325018.8054106-19-129616479202610/ /root/.ansible/tmp/ansible-tmp-1594325018.8054106-19-129616479202610/AnsiballZ_at.py && sleep 0'
[...]

PR #72, opened 2020-08-10
Add integration test for subversion test case

AlanCoding:
This is migrating 1 case from https://github.com/ansible/ansible-builder/pull/44 to the new integration testing system.

I also added cleanup for the images that get created. This may or may not work entirely right.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -21,8 +21,14 @@ def build_dir_and_ee_yml():
 
 
 @pytest.fixture()
-def ee_tag():
-    return str(uuid.uuid4())[:10]
+def ee_tag(request, cli, container_runtime):
+    image_name = 'builder-test-' + str(uuid.uuid4())[:10]
+
+    def delete_image():
+        cli(f'{container_runtime} rmi {image_name}')
+
+    request.addfinalizer(delete_image)
+    return image_name
 
 
 class CompletedProcessProxy(object):
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -25,7 +25,7 @@ def ee_tag(request, cli, container_runtime):
     image_name = 'builder-test-' + str(uuid.uuid4())[:10]
 
     def delete_image():
-        cli(f'{container_runtime} rmi {image_name}')
+        cli(f'{container_runtime} rmi -f {image_name}')
 
     request.addfinalizer(delete_image)
     return image_name
--- a/test/integration/test_build.py
+++ b/test/integration/test_build.py
@@ -1,4 +1,5 @@
 import pytest
+import os
 
 
 def test_build_fail_exitcode():
@@ -25,3 +26,16 @@ def test_build_streams_output(cli, container_runtime, build_dir_and_ee_yml, ee_t
     result = cli(f"ansible-builder build -c {tmpdir} -f {eeyml} -t {ee_tag} --container-runtime {container_runtime}")
     assert f"{container_runtime} build -f {tmpdir}" in result.stdout
     assert f"The build context can be found at: {tmpdir}" in result.stdout
+
+
+def test_user_system_requirement(cli, container_runtime, ee_tag, tmpdir, data_dir):
+    bc = str(tmpdir)
+    ee_def = os.path.join(data_dir, 'subversion', 'execution-environment.yml')
+    cli(
+        f'ansible-builder build -c {bc} -f {ee_def} -t {ee_tag} --container-runtime {container_runtime}'
+    )
+    result = cli(
+        f'{container_runtime} run --rm {ee_tag} svn --help'
+    )
+    assert 'Subversion is a tool for version control' in result.stdout
+
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,6 +1,7 @@
 import yaml
 from unittest import mock
 import pytest
+import os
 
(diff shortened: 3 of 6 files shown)


test/integration/conftest.py
```diff
@@ -21,8 +21,14 @@ def tmp_dir_and_file(ee_contents):
...
-def ee_tag():
-    return str(uuid.uuid4())[:10]
+def ee_tag(request, cli, container_runtime):
+    image_name = 'builder-test-' + str(uuid.uuid4())[:10]
+
+    def delete_image():
+        cli(f'{container_runtime} rmi {image_name}')
+
+    request.addfinalizer(delete_image)
+    return image_name
```
kdelee, line 31: ooo looks good to me

kdelee: actually thinking about, this may act weird because `container_runtime` is parametrized....

test/integration/test_build.py
```diff
@@ -25,3 +26,16 @@ def test_build_streams_output(cli, container_runtime, build_dir_and_ee_yml, ee_t
     result = cli(f"ansible-builder build -c {tmpdir} -f {eeyml} -t {ee_tag} --container-runtime {container_runtime}")
     assert f"{container_runtime} build -f {tmpdir}" in result.stdout
     assert f"The build context can be found at: {tmpdir}" in result.stdout
+
+
+def test_user_system_requirement(cli, container_runtime, ee_tag, tmpdir, data_dir):
```
kdelee, line 31: does this happily just run twice and the test and `ee_tag` get same value from `container_runtime` fixture as the test?

AlanCoding: It did seem so, it was comforting that it ran the same number of tests, it didn't crazily multiple them 2x

kdelee (commented):
looks good, left a question but if it works, it works!

test/integration/conftest.py
AlanCoding: I really don't think this will cause any problems. To test this, let me show you:

```diff
diff --git a/test/integration/conftest.py b/test/integration/conftest.py
index f54ac52..b35a42d 100644
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -22,6 +22,7 @@ def build_dir_and_ee_yml():
 
 @pytest.fixture()
 def ee_tag(request, cli, container_runtime):
+    raise Exception(f'runtime: {container_runtime}')
     image_name = 'builder-test-' + str(uuid.uuid4())[:10]
 
     def delete_image():
@@ -71,4 +72,5 @@ def container_runtime(request, cli):
     if cli(f'{request.param} --version', check=False).returncode == 0:
         return request.param
     else:
+        return request.param
         pytest.skip(f'{request.param} runtime not available')
```

that gives:

```
=================================================================================== ERRORS ===================================================================================
____________________________________________________________ ERROR at setup of test_build_streams_output[docker] _____________________________________________________________

request = <SubRequest 'ee_tag' for <Function test_build_streams_output[docker]>>, cli = <function cli.<locals>.run at 0x10e668950>, container_runtime = 'docker'

    @pytest.fixture()
    def ee_tag(request, cli, container_runtime):
>       raise Exception(f'runtime: {container_runtime}')
E       Exception: runtime: docker

test/integration/conftest.py:25: Exception
____________________________________________________________ ERROR at setup of test_build_streams_output[podman] _____________________________________________________________

request = <SubRequest 'ee_tag' for <Function test_build_streams_output[podman]>>, cli = <function cli.<locals>.run at 0x10e943a70>, container_runtime = 'podman'

    @pytest.fixture()
    def ee_tag(request, cli, container_runtime):
>       raise Exception(f'runtime: {container_runtime}')
E       Exception: runtime: podman

test/integration/conftest.py:25: Exception
___________________________________________________________ ERROR at setup of test_user_system_requirement[docker] ___________________________________________________________

request = <SubRequest 'ee_tag' for <Function test_user_system_requirement[docker]>>, cli = <function cli.<locals>.run at 0x10e943cb0>, container_runtime = 'docker'

    @pytest.fixture()
    def ee_tag(request, cli, container_runtime):
>       raise Exception(f'runtime: {container_runtime}')
E       Exception: runtime: docker

test/integration/conftest.py:25: Exception
___________________________________________________________ ERROR at setup of test_user_system_requirement[podman] ___________________________________________________________

request = <SubRequest 'ee_tag' for <Function test_user_system_requirement[podman]>>, cli = <function cli.<locals>.run at 0x10e9439e0>, container_runtime = 'podman'

    @pytest.fixture()
    def ee_tag(request, cli, container_runtime):
>       raise Exception(f'runtime: {container_runtime}')
E       Exception: runtime: podman

test/integration/conftest.py:25: Exception
```

So I think it's calling the fixture with all the right stuff in all the right places.

The only concern that I personally had is that we may be deleting an image which was never really created, due to an error, or falling into some case where it never builds anything. Let me demo this out:

```
$ docker rmi foobar
Error: No such image: foobar
$ echo $?
1
$ docker rmi -f foobar
Error: No such image: foobar
$ echo $?
0
```

So yes, I think this command _should_ use the `-f` option. I will push a commit that does that. Otherwise, I don't think there's any holes in this, and I really advocate merging this.

beeankha (approved):
This is awesome, thank you @AlanCoding !   🚀

Issue #69, opened 2020-08-06
Build caching regression


shanemcd:
Subsequent runs of `ansible-builder` should be near-instant when inputs dont change. I believed this regressed in https://github.com/ansible/ansible-builder/pull/57. Since we're always re-writing `context/requirements.txt`, this causes the pip install steps to always be run, even when the contents of `requirements.txt` doesn't change.

AlanCoding:
It looks like I'm not able to reproduce this, I put up test at #73. Do you think that using podman vs. docker will make a difference?

AlanCoding:
After I got some better test information, it turns out this _is_ happening specific to podman.

AlanCoding:
Link https://github.com/containers/buildah/pull/2539

The specific problem here might be resolved with #73, but I wouldn't trust that until we change to a different base image and have another round going at the testing.

AlanCoding:
@kdelee test coverage in

https://github.com/ansible/ansible-builder/blob/6d6ba5d5934d84d546300f31cfe6048a78555574/test/integration/test_build.py#L113

PR #73, opened 2020-08-11
Add test for cache reuse, move pytz example for use in test

AlanCoding:
Connect #69 

Currently passes with Docker.


ansible_builder/utils.py
```diff
@@ -20,8 +21,19 @@ def run_command(command, capture_output=False):
         sys.stdout.write(line)
 
     rc = process.poll()
-    if rc != 0:
-        print(MessageColors.FAIL + "An error occured, see output line(s) above for details." + MessageColors.ENDC)
+    if rc is not None and rc != 0:
+        print(MessageColors.FAIL + f"An error occured (rc={rc}), see output line(s) above for details." + MessageColors.ENDC)
```
AlanCoding, line 25: @beeankha for whatever reason, successful commands seem to sometimes give `None` for this, and because of that I was getting a lot of cases where perfectly good runs failed tests. I tried to look for some good documentation for why this was happening, and have come up empty handed.

test/integration/conftest.py
```diff
@@ -20,12 +20,35 @@ def tmp_dir_and_file(ee_contents):
     return tmp_dir_and_file
 
 
-@pytest.fixture()
-def ee_tag(request, cli, container_runtime):
+def run(args, *a, **kw):
```
AlanCoding, line 23: @kdelee so... I re-organized a lot of your stuff.

The alternative here would be to introduce a module `test.integration.lib` and import from that. This is the only case where the function / class shenanigans were used in full force. Elsewhere I did it by moving the scope of a few things up to session.

I learned from the masters.

beeankha (approved):
LGTM!

PR #76, opened 2020-08-14
Move AWX system deps into bindep

shanemcd:


AlanCoding (approved):
thanks 👍 

I'm still on the fence about whether I should try to push these requirements further upstream to an ovirt.ovirt patch, https://github.com/oVirt/ovirt-ansible-collection/compare/master...AlanCoding:ee_req?expand=1

ansible_builder/main.py
```diff
@@ -57,9 +57,9 @@ def build(self):
         print(MessageColors.OK + 'Writing partial Containerfile without collection requirements' + MessageColors.ENDC)
         self.containerfile.write()
         run_command(self.build_command)
-        self.containerfile.prepare_pip_steps()
         print(MessageColors.OK + 'Rewriting Containerfile to capture collection requirements' + MessageColors.ENDC)
         self.containerfile.prepare_system_steps()
+        self.containerfile.prepare_pip_steps()
```
AlanCoding, line 62: Now I realize my cognitive disconnect here

https://github.com/ansible/ansible-builder/pull/59/files#diff-2f525acfcc0134b3361215862821d769

I had made the same change of re-ordering these steps in that PR, but those changes got tied up in unrelated redesigns and has been hanging out for a while without getting merged into devel.

It's always a good idea in my book to prioritize merging uncontroversial small changes associated with larger feature branches 👍

AlanCoding:
And _now_ I remember why this felt circular.

We have another layer for `RUN pip3 install bindep`. It makes the most sense to put that in the combined file for pip requirements (otherwise, it may not have its own dependencies correctly resolved). However, running the system steps require having bindep installed in advance. The pip steps require the system steps to be ran in advance so stuff like pycurl can be re-compiled. Obviously you can't have it both ways.

One _possible_ future enhancement we could consider is to put bindep in its own virtual environment.

AlanCoding:
I had a new idea about how to do that -

How do we isolate a python install when running a command from an image? Gee, maybe we could use containers. The whole idea was to run a command like `docker run --rm alan bindep -b -f /build/bindep_combined.txt` (only once), which runs in an ephemeral container.

We can just... install bindep in that container, combined with that command.

PR #80, opened 2020-08-17
Put deps on a diet

shanemcd:
Removing dephell from our main dep list removes 38 dependencies. Also removed a few other things that aren't needed for tests.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,39 +1,5 @@
 [[package]]
 category = "dev"
-description = "Async http client/server framework (asyncio)"
-name = "aiohttp"
-optional = false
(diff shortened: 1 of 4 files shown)


AlanCoding:
This looks good to me 👍 , since it's a development tool, I think it makes sense that people can install `dephell` on their own to re-generate the setup.py file.

However, I don't know why the poetry.lock file is removed, and I take no position on that.

shanemcd:
`poetry.lock` hasn't been removed, the diff is just too big.

AlanCoding (approved):
no red flags from me

beeankha (approved):

PR #89, opened 2020-08-20
Add test for bad bindep file

AlanCoding:


AlanCoding:
We don't have a set of exit codes that work. See:

https://storyboard.openstack.org/#!/story/2008049

I'll probably file an issue.

PR #97, opened 2020-09-14
Organization refactor for external python-based introspect use

AlanCoding:
Looking at https://github.com/ansible/galaxy-importer/pull/92, I realize that I would like to have a way to call our introspect code here that is more congruent with what they're doing.

To that effect, I am making a completely non-functional change to move something that happens inside of the loop to another method that can be called on its own.

The intended use is that you just pass the path for a collection and get the effective dependency data.


beeankha (approved):
LGTM!

PR #99, opened 2020-09-15
Start migrating docs content into docs site

AlanCoding:


AlanCoding:
I did a test build:

Screen Shot 2020-09-16 at 1 31 23 PM

Screen Shot 2020-09-16 at 1 31 55 PM

That looks perfectly sufficient as far as I'm concerned.

Sometime it might be nice to jazz up the index page.

beeankha (approved):
This looks great, thank you @AlanCoding !

PR #107, opened 2020-09-30
Restructure and expand documentation

shanemcd:
I moved most of the stuff in usage.rst into the index page and played around with the headings.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,32 +1,68 @@
-.. ansible-builder documentation master file, created by
-   sphinx-quickstart on Tue Aug 18 18:59:26 2020.
-   You can adapt this file completely to your liking, but it should at least
-   contain the root `toctree` directive.
+.. ansible-builder documentation master file, created by sphinx-quickstart on
+   Tue Aug 18 18:59:26 2020.  You can adapt this file completely to your liking,
+   but it should at least contain the root `toctree` directive.
 
-Ansible Builder
-===============
-
-Ansible Builder is a tool that aids in the creation of Execution Environments,
-which are container images that serve as Ansible :term:`control nodes<Control node>`.
+Introduction
+============
 
 Using Ansible content that depends on non-default dependencies can be tricky.
 Packages must be installed on each node, play nice with other software installed
 on the host system, and kept in sync.
 
+To help simplify this process, we have introduced the concept of Execution
+Environments, which you can create with Ansible Builder.
+
+
+Execution Environments
+----------------------
+
+Execution Environments are container images that serve as Ansible `control
+nodes`_. Starting in version 2.0, `ansible-runner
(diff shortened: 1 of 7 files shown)


docs/index.rst
```diff
@@ -3,16 +3,185 @@
...
+.. contents::
+   :local:
+   :depth: 3
+
+
+Execution Environments
+----------------------
+
+Execution Environments are container images that serve as Ansible `control
+nodes`_. Starting in version 2.0, `ansible-runner`_ can make use of these images.
```
AlanCoding: Right here seems like the right place to add a link to the corresponding docs page for runner:

https://ansible-runner.readthedocs.io/en/latest/execution_environments.html

As you're adding a link to Ansible's docs page, it would make sense to cover this in the same change.

docs/usage.rst
```diff
@@ -1,132 +1,4 @@
...
-by placing these variables inside the ``env/settings`` file, inside of
-the private data directory.
-
-.. code:: yaml
-
-    ---
-    container_image: image-name
-    process_isolation_executable: podman
-    process_isolation: true
+Once you have created a definition, it's time to build your Execution Environment.
```
AlanCoding: Are you intending to add more to this?

Is this a thing where you're trying to split instructions for writing the spec from the instructions for using the spec?

If so, I feel like it would be good to have an abbreviated start-to-finish section for "I know what collections and containers are, how do I run my playbook?".

shanemcd: Yes, that's one of the reasons this is still a draft PR. I posted this to give people an idea of what's coming. Will continue working on it today.

docs/index.rst
shanemcd: Done

docs/reference.rst
```diff
@@ -0,0 +1,58 @@
+Definition Reference
```
AlanCoding: Looking really good! My only comment looking over the structure would be the title of the section, "Definition Reference" seems jargon-y, I'd image few people would understand it off the bat. Even calling it "Execution Environment Definition" could help. Or "Building Inputs to Builder" or "How to Define an Execution Environment"

shanemcd: I was actually thinking the same thing!

docs/usage.rst
```diff
@@ -1,116 +1,99 @@
...
+your Execution Environment.
 
-Sections following that provide instructions for how a collection
-maintainer should specify dependencies.
 
-The final section gives an example.
-
-Execution Environment Definition
---------------------------------
+``ansible-runner build``
```
maxamillion: Should this be `ansible-builder`?

AlanCoding (approved):

beeankha (approved):
Looks great!

shanemcd:
recheck

shanemcd:
recheck

Issue #74, opened 2020-08-13
Reduce image size by removing dnf/pip cache


AlanCoding:
Argument is made here https://www.redhat.com/sysadmin/tiny-containers that `dnf` steps would more ideally be done with cache cleanup in the same command:

```
RUN dnf install -y nginx \
  	&& dnf clean all \
  	&& rm -rf /var/cache/yum
```

We do a dnf install:

https://github.com/ansible/ansible-builder/blob/d1995bca4afb0c8a8433dc936cb4ba0460018c4b/ansible_builder/steps.py#L81

We could just add this syntax to our command. Presumably we should check image sizes before / after.

I'm not saying it needs to be done, pending feedback, just saving an issue for this.

----

A similar thing may be possible for pip via either `--no-cache-dir` or `~/.cache/pip`

AlanCoding:
Link https://github.com/ansible/ansible-runner/pull/499

shanemcd:
Should we do this by default, or should we document how this (and other tasks) might be done with `steps.append`?

AlanCoding:
in order for it to be effective it has to be done in the same layer, like:

https://github.com/ansible/ansible-runner/blob/e6020b7d33627a5d6656dca64d7ae09dd95cc75c/Dockerfile#L16

I think that will be a super simple patch

AlanCoding:
I think we are generally doing this

PR #109, opened 2020-09-30
Use some cache cleanup methods

AlanCoding:
Connect https://github.com/ansible/ansible-builder/issues/74


pabelanger:
I think we can close this now. We handle it better in multi-stage builds.

Issue #41, opened 2020-07-08
 Implement Custom Verbosity Level


beeankha:
Allow user to set a custom verbosity level using `-v` or `--verbosity` when running builder CLI commands.  

This work should get done _after_ a lot of the other `ansible-builder` issues have been completed, since we want to know exactly _what_ to be verbose about.

shanemcd:
I think we may have a use for this now. I'm not sure if we want to show the `print` statements in https://github.com/ansible/ansible-builder/pull/108/files all the time.

beeankha:
To do:
- [x] Set logger levels via specified verbosity
- [ ] Error out if non-available verbosity is selected 
- [x] Try to set up a system that assigns message color according to verbosity level
- [ ] Add info to documentation

shanemcd:
Implemented in https://github.com/ansible/ansible-builder/pull/112

PR #112, opened 2020-10-06
[WIP] Add a "Verbose" Option to the Build Command

beeankha:
Addressing the issue filed here https://github.com/ansible/ansible-builder/issues/41

### DIFF
--- a/ansible_builder/utils.py
+++ b/ansible_builder/utils.py
@@ -53,6 +53,7 @@ LOGGING = {
 
 
 def configure_logger(verbosity):
+    logging.basicConfig(stream=sys.stdout)
     LOGGING['loggers']['ansible_builder']['level'] = logging_levels[str(verbosity)]
     logging.config.dictConfig(LOGGING)
 
@@ -75,6 +76,8 @@ def run_command(command, capture_output=False, allow_error=False):
 
     rc = process.poll()
     if rc is not None and rc != 0 and (not allow_error):
+        for line in output:
+            logger.error(line)
         logger.error(f"An error occured (rc={rc}), see output line(s) above for details.")
         sys.exit(1)
 
--- a/ansible_builder/utils.py
+++ b/ansible_builder/utils.py
@@ -1,9 +1,9 @@
+import filecmp
+import logging
+import os
+import shutil
 import subprocess
 import sys
-import os
-import filecmp
-import shutil
-import logging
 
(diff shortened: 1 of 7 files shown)


beeankha:
To do:
- [ ] Set logger levels via specified verbosity
- [ ] Error out if non-available verbosity is selected

ansible_builder/utils.py
```diff
@@ -3,13 +3,27 @@
...
+
+def configure_logger(verbosity):
+    logging.basicConfig(stream=sys.stdout, level=logging_levels[str(verbosity)], format='%(message)s')
+
+
 def run_command(command, capture_output=False, allow_error=False):
-    print(MessageColors.HEADER + 'Running command:' + MessageColors.ENDC)
-    print(MessageColors.HEADER + '  {0}'.format(' '.join(command)) + MessageColors.ENDC)
+    logger.debug(MessageColors.HEADER + 'Running command:' + MessageColors.ENDC)
+    logger.debug(MessageColors.HEADER + '  {0}'.format(' '.join(command)) + MessageColors.ENDC)
```
AlanCoding: Big picture, wouldn't it be better to have the logging method apply the color, as opposed to having to put it inside the string passed into the log call?

There are plenty of ways you could do this, and the safest I think would be to, instead of calling `logger.debug`, call utility methods that you introduce like `debug.v("msg")`, where it then maps the method called to both the color and the logging method.

But if you wanted to use the python logging methods debug/info/warn/etc. then you could still put the color in the format statement, because you're _already_ constructing `format='%(message)s'`. Because we're not supporting python 2, you actually have the ability to do the dictionary lookup _inside_ the f-string, like:

https://github.com/borntyping/python-colorlog/blob/ed2d85cb0e9e175add18230a4720f1e0739c9e59/colorlog/colorlog.py#L25

That would also allow you to get rid of the re-typing of `MessageColors.HEADER` for every message

test/integration/conftest.py
```diff
@@ -67,7 +63,7 @@ def delete_images():
                 from_test = True
             if from_test:
                 run(f'{container_runtime} rmi -f {image_name}')
-                logger.warning(f'Deleted image {image_name}')
+                print(f'Deleted image {image_name}')
```
shanemcd: We should change this back.

shanemcd (approved):

PR #116, opened 2020-10-12
Add python 3.8 to envlist

AlanCoding:


AlanCoding:
maybe do, maybe don't need this

Issue #119, opened 2020-11-02
Allow base image to be specified in definition


shanemcd:
Rather than this being a CLI-only argument, it would be nice to be able to specify our base image in the YAML definition.

beeankha:
Closing per merge of PR https://github.com/ansible/ansible-builder/pull/129

PR #129, opened 2020-11-19
Base image can now be specified in the definition file

beeankha:
Addressing Issue https://github.com/ansible/ansible-builder/issues/119

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -25,14 +25,23 @@ PIP_COMBINED = 'requirements_combined.txt'
 class AnsibleBuilder:
     def __init__(self, action=None,
                  filename=constants.default_file,
-                 base_image=constants.default_base_image,
+                 base_image=None,
                  build_context=constants.default_build_context,
                  tag=constants.default_tag,
                  container_runtime=constants.default_container_runtime,
                  verbosity=0):
         self.action = action
         self.definition = UserDefinition(filename=filename)
-        self.base_image = base_image or self.definition.raw.get('base_image')
+
+        # Handle precedence of the base image
+        if base_image is not None:
+            self.base_image = base_image
+        if base_image is None:
+            if self.definition.raw.get('base_image'):
+                self.base_image = self.definition.raw.get('base_image')
+            else:
+                self.base_image = constants.default_base_image
+
         self.tag = tag
         self.build_context = build_context
         self.container_runtime = container_runtime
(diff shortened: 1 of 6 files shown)


beeankha:
recheck

ansible_builder/main.py
```diff
@@ -190,6 +198,15 @@ def validate(self):
                 if not os.path.exists(requirement_path):
                     raise DefinitionError("Dependency file {0} does not exist.".format(requirement_path))
 
+        ee_base_image = self.get_base_image()
```
AlanCoding: I hate to see `get_base_image` as its own method when it's not the canonical way that the value gets set in practice. I'd prefer to lose the method and use the same `self.definition.raw.get('base_image')` here

beeankha: Makes sense!

ansible_builder/main.py
```diff
@@ -32,12 +32,13 @@ def __init__(self, action=None,
                  verbosity=0):
         self.action = action
         self.definition = UserDefinition(filename=filename)
+        self.base_image = self.definition.raw.get('base_image') or base_image
```
AlanCoding: I don't know about the precedence. If I provide the base image both in the definition file, and with the CLI option... I would think the CLI option should win. That tends to be how we want config processing to go. @shanemcd what do you think? Or else we could error if given both.

We would have to re-think how the kwarg default works, it might make more sense for `None` to be the value from the kwarg.

beeankha: > If I provide the base image both in the definition file, and with the CLI option... I would think the CLI option should win.

Agreed, I'll switch those around and clarify the precedence in the docs.

beeankha: Have made some updates, will work on making sure integration tests pass next.

test/test_main.py
```diff
@@ -109,25 +105,29 @@ def test_definition_syntax_error(self, data_dir):
         path = os.path.join(data_dir, 'definition_files/bad.yml')
 
         with pytest.raises(DefinitionError) as error:
-            AnsibleBuilder(filename=path)
+            AnsibleBuilder(filename=path, base_image='my-custom-image')
```
AlanCoding: these test changes shouldn't be needed, should they? I'd like to see if we can roll some of those back.

ansible_builder/main.py
```diff
@@ -190,6 +191,21 @@ def validate(self):
                 if not os.path.exists(requirement_path):
                     raise DefinitionError("Dependency file {0} does not exist.".format(requirement_path))
 
+        ee_base_image = self.raw.get('base_image')
+        if ee_base_image is None:
+            raise DefinitionError(textwrap.dedent(
+                """
+                Error: Please specify a base image for the execution environment.
+                """)
```
AlanCoding: I don't know about this, I would prefer to keep the default behavior the same.

AlanCoding:
put up revised suggestions in https://github.com/beeankha/ansible-builder/pull/2

AlanCoding (approved):
looks like behavior aligns with docs so :+1:

shanemcd (approved):

beeankha:
recheck

Issue #126, opened 2020-11-19
Default verbose level to 2


chrismeyersfsu:
First time running `ansible-builder`, I thought nothing was happening. Consider verbose level 2 to be the default and have a `silence` option.

beeankha:
PR https://github.com/ansible/ansible-builder/pull/131 has been merged, closing this issue!

PR #131, opened 2020-12-07
Change Verbosity Default + Implementation

beeankha:
Addressing Issue https://github.com/ansible/ansible-builder/issues/126

With this PR, instead of invoking verbosity via `-v` or `-vv` (i.e. using `action=count`), the user can now type `-v 3` or `--verbosity 1` etc. Any integer invoked that is out of the range of 0 - 3 will error out and show the valid choices available:

Screen Shot 2020-12-07 at 12 46 01 PM

The verbosity now defaults to `2` instead of `0` so that even if you don't specify verbosity, some minimal output will be displayed.

Also added information about the `--verbosity` option to the documentation, as well as a new integration test that checks the build fails and emits the error above when invalid verbosity level is used.

### DIFF
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -102,11 +102,12 @@ def parse_args(args=sys.argv[1:]):
 
         p.add_argument('-v', '--verbosity',
                        dest='verbosity',
-                       action='count',
-                       default=0,
+                       type=int,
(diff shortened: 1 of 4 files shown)


shanemcd (approved):

pabelanger:
recheck

beeankha:
recheck

Issue #79, opened 2020-08-16
Support some means of customizing Galaxy server settings


AlanCoding:
The galaxy CLI install commands are part of the Containerfile. Because it's a clean, replicable, environment it will use the Ansible defaults. There needs to be a means of passing a token and other settings for a private account to an Automation Hub server.

The challenge here is that we can't have any credentials left anywhere in the image, in intermediary layers, and maybe not in the build context at all. We will want building with the build context to be replicable, but that seems impossible to reconcile with providing a protected secret token.

----

My only readily available solution that doesn't involving tweaking the requirements:

 - perform `ansible-galaxy collection download` as a subprocess before starting any form of building, and store the output in build context
 - ADD the downloaded content to the image
 - Inside the container, install the downloaded content

Downside is that this would require a relatively recent Ansible version.

Some other options are probably possible, but will require discussion.

AlanCoding:
Note: discussed solution is messy and not ideal.

We would prefer to use simple podman / docker build args, but need to assure they do not get recorded in image.

AlanCoding:
I am hesitating at that suggestion to use build args.

https://pythonspeed.com/articles/docker-build-secrets/

It looks like they are present in the history of the final image. There is no important circumstance where this would be acceptable for the Ansible settings here.

shanemcd:
It seems like build args might be ok if we use multi-stage builds. They do not show up in the output of `history` when run against the final image.

beeankha:
PR https://github.com/ansible/ansible-builder/pull/137 has been merged, closing this issue!

PR #137, opened 2020-12-08
Implement Ability to Support Custom Galaxy Server Settings

beeankha:
Addressing Issue https://github.com/ansible/ansible-builder/issues/79

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -5,7 +5,10 @@ import yaml
 
 from . import constants
 from .exceptions import DefinitionError
-from .steps import AdditionalBuildSteps, GalaxyInstallSteps, GalaxyCopySteps, PipSteps, BindepSteps
+from .steps import (
+    AdditionalBuildSteps, GalaxyInstallSteps, GalaxyCopySteps,
+    PipSteps, BindepSteps, AnsibleConfigSteps
+)
 from .utils import run_command, write_file, copy_file
 from .requirements import sanitize_requirements
 import ansible_builder.introspect
@@ -58,6 +61,10 @@ class AnsibleBuilder:
         return self.definition.version
 
     @property
+    def ansible_config(self):
+        return self.definition.ansible_config
+
+    @property
     def build_command(self):
         return [
             self.container_runtime, "build",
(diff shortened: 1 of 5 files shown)


ansible_builder/main.py
```diff
@@ -307,6 +317,19 @@ def prepare_system_steps(self, bindep_output):
...
+    def prepare_final_stage_steps(self):
+        self.steps.append("FROM {0}".format(self.base_image))
+        return self.steps
+
+    def prepare_galaxy_copy_steps(self):
+        if self.definition.get_dep_abs_path('galaxy'):
+            self.steps.extend(GalaxyCopySteps())
+        return self.steps
+
+    def delete_intermediate_image(self):
```
shanemcd: Not sure it makes sense to have this live in the `Containerfile` class. Maybe this should live somewhere in `AnsibleBuilder`?

beeankha: Moving it to the bottom of AnsibleBuilder class for now.

ansible_builder/steps.py
```diff
@@ -46,7 +46,21 @@ def __init__(self, requirements_naming):
...
+
+class GalaxyCopySteps(Steps):
+    def __init__(self):
+        """Assumes given requirements file name has been placed in the build context
+        """
+        self.steps = []
+        self.steps.extend([
+            "",
+            "COPY --from=builder {0} {0}".format(constants.base_roles_path),
+            "COPY --from=builder {0} {0}".format(constants.base_collections_path),
```
pabelanger, line 63: What is we copied in to the base image (from builder) the tarball folder of collections.   Then, we could run ansible-galaxy install tarballs/*.tar.gz again.

The main idea here is, there will be uscases where uses will build custom collections tarballs, and need to install them over something from the galaxy server. And people could remove the need for requirement files in the builder phase, and copy in their own tarballs.

This is kinda of how we are doing it with python wheels today in python-builder / python-base images: https://github.com/ansible/python-builder-image/ https://github.com/ansible/python-base-image/

shanemcd: I like this idea, but we may want to do this in a followup PR depending on whether @beeankha thinks she can accomplish this before she leaves for winter break. I think she was hoping to get something merged before then.

docs/definition.rst
```diff
@@ -30,9 +34,18 @@ definition file in the ``base_image`` section as a string; this is an
...
 command line, the CLI option will take precedence.
 
+Ansible Config File Path
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+When using an ``ansible.cfg`` file to pass a token and other settings for a
+private account to an Automation Hub server, listing the config file path here
+(as a string) will enable it to be included as a build argument in the initial
+phase of the build. The information contained within the config file will not
+appear in the output of ``history`` when run against the final image.
```
AlanCoding: but will be in the history of the intermediate image... so don't share that.

shanemcd (approved):

PR #142, opened 2020-12-10
Use --no-cache-dir for pip3

pabelanger:
By using --no-cache-dir option for pip, we disable the ability for pip
to cache wheels inside the container. This will allow us to save some
space inside the container.

### DIFF
--- a/ansible_builder/steps.py
+++ b/ansible_builder/steps.py
@@ -91,7 +91,7 @@ class PipSteps(Steps):
         self.steps.append("ADD {0} /build/".format(context_file))
         container_path = os.path.join('/build/', context_file)
         self.steps.append(
-            "RUN pip3 install --upgrade -r {content}".format(content=container_path)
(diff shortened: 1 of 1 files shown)


ansible_builder/steps.py
```diff
@@ -91,7 +91,7 @@ def __init__(self, context_file):
         self.steps.append("ADD {0} /build/".format(context_file))
         container_path = os.path.join('/build/', context_file)
         self.steps.append(
-            "RUN pip3 install --upgrade -r {content}".format(content=container_path)
+            "RUN pip3 install --no-cache-dir --upgrade -r {content}".format(content=container_path)
```
pabelanger, line 94: So, we have a few options here. I think we could expect users to setup their own pip.conf file, which is fair. Or be more aggressive here, in ansible-builder.

I'm not sure the best option honestly, if we start config file settings here, then we are on the hook for more pip settings in the future.

PR #143, opened 2020-12-11
Remove extra double-spacing from subcommands

AlanCoding:
This had been driving me nuts, the subcommands had extra spaces inserted between every line.

With this fix, this is how output looks:

Screenshot from 2020-12-10 21-42-23

The behavior in the case of `capture_output=True` gets a little bit wackier. The goal is to not print anything twice, or omit printing something important once. The important things are:

 - what command had an error and
 - what the output of that command was

Here is my manual testing, putting in `podman foo` for the first part of the build phase, which has its output captured.

Screenshot from 2020-12-10 21-56-35


beeankha (approved):
The output looks soooooo much better now! ✨

Issue #105, opened 2020-09-29
Allow for specifying separate build-time and runtime system-level dependencies


shanemcd:
It would be cool to have a way of avoiding crud like gcc and python3-devel in the final images.

This may be a perfect use case for bindep "profiles". There could be a `build` profile that we install at the beginning of the build and uninstall at the end.

pabelanger:
++ I dont mind looking into this

shanemcd:
I ended up hacking on this in #148

AlanCoding:
Closing as implemented, we have another issue discussing documentation for this.

Issue #147, opened 2020-12-12
Support special-purpose bindep profiles


AlanCoding:
We have a proposal that we would document certain bindep profiles that, if included by a collection, would trigger certain behavior by `ansible-builder`. We have discussed:

 - `epel` - if a collection had entries with this profile, then the epel repo would be enabled before installing those requirements
 - `compile` - if a collection had entries with this profile, then it would indicate that it is needed to install _other_ requirements (python ones specifically), but not required to be in the final build. Several are being deal with at the https://github.com/ansible/awx-ee repo

This would be a notable addition to the core feature set of ansible-builder, and need to be documented. Ping @pabelanger

AlanCoding:
I guess this is kind of a duplicate with https://github.com/ansible/ansible-builder/issues/105

AlanCoding:
Closing as replaced by https://github.com/ansible/ansible-builder/issues/156

PR #148, opened 2020-12-13
Pre-build wheels, before copying them into final image for installation

shanemcd:
Implementation of the idea in https://github.com/ansible/ansible-builder/issues/105

Depends-On: https://github.com/ansible/python-builder-image/pull/15
Depends-On: https://github.com/ansible/ansible-runner/pull/581

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -6,8 +6,7 @@ import yaml
 from . import constants
 from .exceptions import DefinitionError
 from .steps import (
-    AdditionalBuildSteps, GalaxyInstallSteps, GalaxyCopySteps, AnsibleConfigSteps,
-    PipDownloadSteps, PipInstallSteps, BindepSteps,
+    AdditionalBuildSteps, GalaxyInstallSteps, GalaxyCopySteps, AnsibleConfigSteps
 )
 from .utils import run_command, write_file, copy_file
 from .requirements import sanitize_requirements
@@ -21,10 +20,7 @@ CONTEXT_FILES = {
     'galaxy': 'requirements.yml'
 }
 
-BINDEP_BUILD_COMBINED = 'bindep_build_combined.txt'
 BINDEP_COMBINED = 'bindep_combined.txt'
-BINDEP_BUILD_OUTPUT = 'bindep_build_output.txt'
-BINDEP_RUNTIME_OUTPUT = 'bindep_runtime_output.txt'
 PIP_COMBINED = 'requirements_combined.txt'
 
 
@@ -99,24 +95,15 @@ class AnsibleBuilder:
         system_lines = ansible_builder.introspect.simple_combine(collection_data['system'])
         python_lines = sanitize_requirements(collection_data['python'])
 
-        bindep_build_output = []
-        bindep_runtime_output = []
         if system_lines:
-            write_file(os.path.join(self.build_context, BINDEP_COMBINED), system_lines + [''])
+            bindep_file = os.path.join(self.build_context, BINDEP_COMBINED)
+            write_file(bindep_file, system_lines + [''])
 
-            rc, bindep_build_output = self.run_in_container(
-                ['bindep', 'build', '-b', '-f', '/context/{0}'.format(BINDEP_COMBINED)],
-                allow_error=True, capture_output=True
-            )
+        if python_lines:
+            pip_file = os.path.join(self.build_context, PIP_COMBINED)
+            write_file(pip_file, python_lines)
 
-            write_file(os.path.join(self.build_context, BINDEP_BUILD_COMBINED), bindep_build_output + [''])
-
-            rc, bindep_runtime_output = self.run_in_container(
-                ['bindep', '-b', '-f', '/context/{0}'.format(BINDEP_COMBINED)],
-                allow_error=True, capture_output=True
-            )
-
-        return (bindep_build_output, bindep_runtime_output, python_lines)
+        return (system_lines, python_lines)
 
     def build(self):
         # Phase 1 of Containerfile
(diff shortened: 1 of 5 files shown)


pabelanger:
What do you think about using the same idea we did in ansible/python-builder image, where we run the assemble script, to pre-build wheels?  Then using the same idea, we copy them over into base, and can pip install or use install-from-bindep script, too.

shanemcd:
@pabelanger The main problem I see with reusing `assemble` at this point is its assumption that the source code for the project will be at `/tmp/src`, which it will use to build a wheel:

https://github.com/ansible/python-builder-image/blob/0070fc340cf12b66f54f0a801a15f7bf7dc04ae4/scripts/assemble#L57

If we can make this conditional, I have a pretty good idea about how we'd further refine this PR (either now or in a follow-up effort) to make use of the work you've done. I already used them as a source of inspiration here:

https://github.com/ansible/ansible-builder/blob/1d94bf8723dd505117a0adc984d984d5ee42dafe/ansible_builder/steps.py#L94-L95

https://github.com/ansible/ansible-builder/blob/1d94bf8723dd505117a0adc984d984d5ee42dafe/ansible_builder/steps.py#L110-L111

And I am just now realizing you went with a profile named `compile` instead of `build`. I'm happy to adopt your choice here. I'll make this change in the morning.

shanemcd:
Another thing I just remembered - if we are going to rely on these scripts being present in the base image, we will need to make python-builder and/or python-base available downstream as well.

AlanCoding:
It's a little hard for me to discern purpose by reading the code here. I think what you're doing here is using a bindep profile of "build":

https://github.com/ansible/ansible-builder/blob/1d94bf8723dd505117a0adc984d984d5ee42dafe/ansible_builder/main.py#L108

(but again, that's hard to read, some code comments would help even if we delay actual docs)

I was trying to start some conversation on that in https://github.com/ansible/ansible-builder/issues/147, when we were spitballing the `compile` name of the profile, but same intent.

As far as I have read, the implementation looks fine to me. It necessarily requires another Dockerfile stage, so that matches expectations fine.

pabelanger:
> @pabelanger The main problem I see with reusing `assemble` at this point is its assumption that the source code for the project will be at `/tmp/src`, which it will use to build a wheel:
> 
> [ansible/python-builder-image@`0070fc3`/scripts/assemble#L57](https://github.com/ansible/python-builder-image/blob/0070fc340cf12b66f54f0a801a15f7bf7dc04ae4/scripts/assemble#L57)
> 
> If we can make this conditional, I have a pretty good idea about how we'd further refine this PR (either now or in a follow-up effort) to make use of the work you've done. I already used them as a source of inspiration here:
> 
> https://github.com/ansible/ansible-builder/blob/1d94bf8723dd505117a0adc984d984d5ee42dafe/ansible_builder/steps.py#L94-L95
> 
> https://github.com/ansible/ansible-builder/blob/1d94bf8723dd505117a0adc984d984d5ee42dafe/ansible_builder/steps.py#L110-L111
> 
> And I am just now realizing you went with a profile named `compile` instead of `build`. I'm happy to adopt your choice here. I'll make this change in the morning.

Yes, you are right. I've started work on that in https://github.com/ansible/python-builder-image/pull/13

pabelanger:
If you updated the first comment and added
```
Depends-On: https://github.com/ansible/python-builder-image/pull/14
```
That will cause zuul to pull in ansible/python-builder:latest from that PR, and should fix the failure

pabelanger:
recheck

pabelanger:
Doh, we actually need to update .zuul.d/jobs.yaml for ansible-builder-tox-integration.

You'll add
```- python-builder-container-image```
to requires key

**Comment by pabelanger** [comment]
recheck

File: ansible_builder/main.py [hunk 1]
```diff
@@ -329,26 +329,46 @@ def prepare_galaxy_install_steps(self):
...
-            self.steps.extend(BindepSteps(BINDEP_OUTPUT))
+    def prepare_system_runtime_deps_steps(self):
+        requirements_file_exists = os.path.exists(os.path.join(self.build_context, PIP_COMBINED))
+        bindep_exists = os.path.exists(os.path.join(self.build_context, BINDEP_COMBINED))
+
+        if requirements_file_exists or bindep_exists:
+            self.steps.extend([
+                "COPY --from=builder /output/ /output/",
+                "RUN /output/install-from-bindep",
+                "RUN rm -rf /output"
```
AlanCoding: since it's another layer, I don't know that you save any size by removing the output folder.

implication being: maybe it would be more transparent to leave it in

shanemcd: Let me run some tests and report back.

ansible_builder/main.py
```diff
@@ -329,26 +329,46 @@ def prepare_galaxy_install_steps(self):
...
+        requirements_file_exists = os.path.exists(os.path.join(self.build_context, PIP_COMBINED))
+        if requirements_file_exists:
+            self.steps.append("ADD {0} /tmp/src/requirements.txt".format(PIP_COMBINED))
+
+        bindep_exists = os.path.exists(os.path.join(self.build_context, BINDEP_COMBINED))
+        if bindep_exists:
+            self.steps.append("ADD {0} /tmp/src/bindep.txt".format(BINDEP_COMBINED))
+
+        if requirements_file_exists or bindep_exists:
+            self.steps.append("RUN assemble")
```
AlanCoding, line 342: If we were being absolutists about code style, every string that goes in the Dockerfile would originate in the `ansible_builder.steps` module. If not, I envision the steps module eventually going away.

ansible_builder/main.py
pabelanger: AFAIK, until podman supports buildkit, the COPY command will put the wheels on disk.

I've experimented about standing up an HTTP server in the python-builder layer, they using pip install and set index-url to localhost, but still WIP.

At worst case, it might add a few MBs right now

AlanCoding:
This indirectly makes

https://github.com/ansible/python-builder-image/tree/main/scripts

A part of the ansible-builder code base.

I have tried to link and reference `ansible-runner` a great deal in the documentation, because builder makes no sense without runner. With this, the python-builder-image will become critical for understanding the technical operation of builder.

AlanCoding:
It wasn't obvious to me from the get-go, but we're also inheriting the special bindep profile behaviors from the python-builder-image, like `compile` and `epel`

https://github.com/ansible/python-builder-image/blob/2cfbb924e2d44206fb78101b3e9b9b402163161e/scripts/assemble#L36

Since these are pre-existing, I get that we may not have to write new documentation, but instead link to the existing documentation for these and explain how it fits into what ansible-builder does.

AlanCoding (approved):
This is pretty huge, but mostly in terms of the new functionality it's adding. It's well considered to not adversely affect the existing use cases... but it will take some time to digest the new capabilities.

ansible_builder/main.py
AlanCoding: I expected this folder is requirements files, so yeah the disk size should be totally trivial. It's more that deleting it isn't truly hiding it, if something is in there that shouldn't be.

pabelanger (approved):
This looks good! Tested it locally myself and worked as excepted

ansible_builder/main.py
shanemcd: @AlanCoding this directory contains the pre-built wheels that we are installing in the final image. I like your idea of leaving this metadata behind, so I'm going to switch to only removing `/output/wheels`

ansible_builder/main.py
shanemcd: I might further refactor this later. Dunno yet.

PR #157, opened 2020-12-16
Write out files to _build dir in build context

shanemcd:
This tidies the build context.

Before:

```
$ ls -1
bindep_combined.txt
bindep_output.txt
Containerfile
Dockerfile
introspect.py
LICENSE.md
README.md
requirements_combined.txt
requirements.txt
requirements.yml
tools
tox.ini
```

After:

```
$ ls -1
Containerfile
Dockerfile
execution-environment.yml
LICENSE.md
README.md
sources
tools
tox.ini
```

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -20,6 +20,9 @@ CONTEXT_FILES = {
     'galaxy': 'requirements.yml'
 }
 
+# Subdirectory in the build context where files to ADD get copied
+CONTEXT_BUILD_OUTPUTS_DIR = '_build'
+
 BINDEP_COMBINED = 'bindep_combined.txt'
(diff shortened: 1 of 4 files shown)


AlanCoding (approved):
My only concern is around the messaging. The "sources" folder is an output that ansible-builder produces. This could be confusing. It's more like an intermediate artifact in the process... I don't have a better word for that.

pabelanger:
> My only concern is around the messaging. The "sources" folder is an output that ansible-builder produces. This could be confusing. It's more like an intermediate artifact in the process... I don't have a better word for that.

Given it is an output, maybe 'build' is more appropriate? That is inline with say something like sphinx.

AlanCoding:
any variation on "build" sounds good to me

PR #158, opened 2020-12-16
Adopt pbr / dynamic versioning

shanemcd:


pabelanger:
So, I wonder if the right way to do this, is create a poetry plugin that support pbr.  I've found https://pypi.org/project/poetry-dynamic-versioning/ which _kinda_ does what we want.  However, I don't think we should use that, as we'd be fighting over which tools is setting up the version.

pabelanger:
recheck

pabelanger (approved):

PR #160, opened 2020-12-18
Add support for build args

shanemcd:
Also provides a `ANSIBLE_RUNNER_IMAGE` build arg that can be overridden.

I tried to develop the same functionality as `docker build`. 

Build args can be specified with `--build-arg KEY=value` or `--build-arg KEY` where `value` is set in the environment (this is only supported in Docker).

Example `Dockerfile`:

```
ARG ANSIBLE_RUNNER_IMAGE=quay.io/ansible/ansible-runner:devel
FROM $ANSIBLE_RUNNER_IMAGE as galaxy

ADD _build/requirements.yml /build/_build/requirements.yml

RUN ansible-galaxy role install -r /build/_build/requirements.yml --roles-path /usr/share/ansible/roles
RUN ansible-galaxy collection install -r /build/_build/requirements.yml --collections-path /usr/share/ansible/collections

RUN mkdir -p /usr/share/ansible/roles /usr/share/ansible/collections

FROM quay.io/ansible/python-builder:latest as builder

ADD _build/requirements_combined.txt /tmp/src/requirements.txt
ADD _build/bindep_combined.txt /tmp/src/bindep.txt
RUN assemble

FROM $ANSIBLE_RUNNER_IMAGE

COPY --from=galaxy /usr/share/ansible/roles /usr/share/ansible/roles
COPY --from=galaxy /usr/share/ansible/collections /usr/share/ansible/collections

RUN alternatives --set python /usr/bin/python3
```

### DIFF
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -2,7 +2,6 @@ import argparse
 import logging
 import sys
 import yaml
-import os
 
 from . import __version__
 
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import sys
 import yaml
+import os
 
 from . import __version__
 
@@ -100,6 +101,11 @@ def parse_args(args=sys.argv[1:]):
                        default=constants.default_container_runtime,
                        help='Specifies which container runtime to use. Defaults to {0}.'.format(constants.default_container_runtime))
 
+        p.add_argument('--build-arg',
+                       action=BuildArgAction,
+                       default={},
+                       dest='build_args')
+
         p.add_argument('-v', '--verbosity',
                        dest='verbosity',
                        type=int,
@@ -134,3 +140,17 @@ def parse_args(args=sys.argv[1:]):
     args = parser.parse_args(args)
 
     return args
+
+
+class BuildArgAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        key, *value = values.split('=')
+        attr = getattr(namespace, self.dest)
+
+        # None signifies that the build-arg will come from the environment.
+        # This is currently only supported by Docker. Podman will treat any
+        # usage of the $VALUE as a literal string.
+        if value:
+            attr[key] = value[0]
+        else:
+            attr[key] = None
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -39,6 +39,7 @@ class AnsibleBuilder:
     def __init__(self, action=None,
                  filename=constants.default_file,
                  base_image=None,
+                 build_args=None,
                  build_context=constants.default_build_context,
                  tag=constants.default_tag,
                  container_runtime=constants.default_container_runtime,
@@ -59,6 +60,7 @@ class AnsibleBuilder:
         self.build_context = build_context
         self.build_outputs_dir = os.path.join(build_context, CONTEXT_BUILD_OUTPUTS_DIR)
         self.container_runtime = container_runtime
+        self.build_args = build_args or {}
         self.containerfile = Containerfile(
             definition=self.definition,
             base_image=self.base_image,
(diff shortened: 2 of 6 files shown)


ansible_builder/cli.py
```diff
@@ -100,6 +101,11 @@ def parse_args(args=sys.argv[1:]):
                        default=constants.default_container_runtime,
                        help='Specifies which container runtime to use. Defaults to {0}.'.format(constants.default_container_runtime))
 
+        p.add_argument('--build-arg',
+                       action=BuildArgAction,
+                       default={},
+                       dest='build_args')
```
AlanCoding, line 106: I don't get how this jives with the existing `-b` option. The `-b` is the default, so why would someone ever intend to provide this, but not that?

AlanCoding:
can this be leveraged with append / prepend steps?

ansible_builder/cli.py
pabelanger: @AlanCoding The main idea here, is we do not want to re-generate a new Dockerfile if we change the base image, otherwise for downstream stuff We'll need 2 git branches for say stable-2.9 / stable-2.11 images of ansible-runner.

https://github.com/ansible/network-ee/pull/11 is an example of how we could use a single branch, but at job level pass in different ansible_runner images via build-args.

ansible_builder/main.py
```diff
@@ -298,7 +311,8 @@ def __init__(self, definition,
         self.container_runtime = container_runtime
         self.tag = tag
         self.steps = [
-            "FROM {0} as galaxy".format(self.base_image),
+            "ARG BASE_IMAGE={}".format(self.base_image),
```
pabelanger: maybe ANSIBLE_RUNNER_IMAGE, if we want to be specific that it should be some sort of ansible-runner based image.

ansible_builder/cli.py
shanemcd: @AlanCoding `-b` is shorthand for `--base-image`. This new option does not provide a shorthand option (neither do Docker or Podman for this feature).

shanemcd: I've also pushed an update to the docs to better explain this.

ansible_builder/main.py
shanemcd: Sure. I've pushed that up.

shanemcd:
@AlanCoding Yes, you could. By default we provide a build arg `ANSIBLE_RUNNER_IMAGE` that can be provided, which Paul mentioned above. But you you leverage this in custom build steps too.

pabelanger:
K, I've updated https://github.com/ansible/network-ee/pull/11 to depends-on this PR. To give working example.

docs/usage.rst
```diff
@@ -82,6 +82,28 @@ directory. To specify another location:
...
+``--build-arg``
+*************
+
+To use Podman or Docker's build-time variables, specify them the same way you would with ``podman build`` or ``docker build``.
+
+By default, the Containerfile / Dockerfile outputted by Ansible Builder contains a build argument ``ANSIBLE_RUNNER_IMAGE``, which can be useful for rebuilding Execution Environments without modifying any files.
+
+.. code::
+
+   $ ansible-builder build --build-arg FOO=bar
```
AlanCoding, line 104: :+1:

ansible_builder/cli.py
AlanCoding: Sorry, I should have been more specific.

I'm not questioning the utility of this feature, I understand its use case.

Assuming this is merged, I'm questioning the utility of the `-b` option. Before you would have done `-b foo_image:latest`, but after the change you are making, that is redundant with using `--build-arg ANSIBLE_RUNNER_IMAGE=foo_image:latest`.

So to make a prescriptive suggestion, lose the `-b` option. Keep its mention you added in the docs, but offer the newer syntax.

Everything about the base image option will cause friction with users. It suggests that we will support other build chains (we won't) or that we would support changing important things in ours like the OS (we won't) and with multi-stage builds there's even more fog about _which_ image it refers to (which the new syntax is more clear about).

I think this project is young enough that we can get away with removing an option. We are changing some pretty fundamental things in other ways, so removing it should be small potatoes.

AlanCoding:
This should be pretty close to getting moving, and I want to get it in soon. To express my personal preferences, before merging, I would like to:

 - remove the `-b` CLI option, and maybe add a new section in docs for release notes to mention it
 - add a build arg for the base `builder` image

pabelanger:
I think once we clean up the lint failure, this should be good? I've tested it again and works as expected.

AlanCoding:
Closing in favor of https://github.com/ansible/ansible-builder/pull/167

PR #164, opened 2020-12-21
Drop poetry

shanemcd:


AlanCoding (approved):

tox.ini
```diff
@@ -2,47 +2,34 @@
...
-    poetry env info
-    poetry run py.test -v test -k "not integration" {posargs}
+    py.test -v test -k "not integration" {posargs}
 
 [testenv:linters]
-basepython = python3
+deps =
+    flake8
+    pylint
+    yamllint
```
pabelanger, line 18: what if we put this into test-requirements.txt ?

Then we'd have
```
[testenv]
basepython = python3
install_command = pip install {opts} {packages}
deps = -r{toxinidir}/requirements.txt
       -r{toxinidir}/test-requirements.txt
```

tox.ini
```diff
@@ -2,47 +2,34 @@
...
-# environment via the basepython key
-basepython = python3
-
 [testenv]
-skip_install = true
-deps = poetry >= 1.0.5
+usedevelop = True
+setenv =
+  VIRTUAL_ENV={envdir}
+deps =
```
pabelanger, line 9: same comment below

tox.ini
```diff
@@ -2,47 +2,34 @@
...
+    flake8 ansible_builder test
+    yamllint --version
+    yamllint -s .
 
-# Tox target to ensure both pyproject.toml and setup.py are in sync
 [testenv:integration]
 # rootless podman reads $HOME
 passenv = HOME
+deps =
+    pytest-xdist
```
pabelanger, line 29: this also

pabelanger:
I like it, we'll have to revert our release jobs too.

tox.ini
shanemcd: I wanted to be super explicit here because I wanted to avoid pushing a broken package to pypi. 

yamllint declares a dependency on pyyaml. 

Right now, if you remove pyyaml from requirements.txt, the unit tests will fail. I like it this way.

PR #165, opened 2020-12-23
Fix ansible-builder --version command

pabelanger:
No longer use the hardcoded version string, we can look it up via PBR.


shanemcd:
I don't love the idea of making PBR a runtime dependency. I just tested this and it seems to work (taken from [here](https://stackoverflow.com/questions/48383737/get-version-from-git-tags-through-pbr)):

```
>>> import pkg_resources
>>> version = pkg_resources.get_distribution('ansible_builder').version
>>> print(version)
0.5.1
```

pabelanger:
Pull-request updated, HEAD is now 6b326cdbffea10552835cbab7ff7f5ab6a25168c

pabelanger:
> I don't love the idea of making PBR a runtime dependency. I just tested this and it seems to work (taken from [here](https://stackoverflow.com/questions/48383737/get-version-from-git-tags-through-pbr)):
> 
> ```
> >>> import pkg_resources
> >>> version = pkg_resources.get_distribution('ansible_builder').version
> >>> print(version)
> 0.5.1
> ```

Yah, there is some more functionality it includes, like getting pre-release version info

AlanCoding:
closing with replacement

Issue #159, opened 2020-12-18
Allow Customization of `builder` stage


DWSR:
I'm using `ansible-builder` `0.4.0`, which is hardcoded to use `quay.io/ansible/python-builder` as the base for the `builder` stage.

This is very limiting for me for a couple of reasons:
* I need to pull dependencies through a private Artifactory instance which is using a private CA cert and I have no way of getting that CA certificate to the container image right now.
* I need to pull dependencies through a private Artifactory instance and I can't tell pip in the `builder` stage to do so. If I put `--extra-index-url` or even `--index-url` at the top of my `requirements.txt`, I get the following warning: `/Users/brandonmcnama/.local/pipx/venvs/ansible/lib/python3.9/site-packages/requirements/parser.py:44: UserWarning: Private repos not supported. Skipping.`
* The python-builder image has a particular version of Python installed, which means I'm forced to use that same version of Python in my base image.

I'd like to propose the following:

* Allow the user to specify a builder image to use in the execution environment manifest.
* Allow the user to specify "before" and "after" steps that are prepended and appended to the instructions for the `builder` stage specifically.

AlanCoding:
> Allow the user to specify "before" and "after" steps that are prepended and appended to the instructions for the builder stage specifically.

Point taken, but, I would put this at much lower priority than supporting the `builder` image itself, because you can solve the same problem with customizing your base `builder` image.

In #160, we are changing the base image customization to a build arg. This is so that in a sequence of commands of `ansible-builder build` and then later `podman build`, the user can apply a custom setting to `podman build` as needed. This approach makes it obvious how to support a custom `builder` image, which is to make it a build argument in the syntax of the Containerfile/Dockerfile.

DWSR:
> Point taken, but, I would put this at much lower priority than supporting the `builder` image itself, because you can solve the same problem with customizing your base `builder` image.

This only solves doing things before, not after, and it makes those things less discoverable than having them declared in the execution env manifest. IMO it should be both, but I agree that customizing the steps around the builder is lower priority than allowing an arbitrary builder image to be used.

> In #160, we are changing the base image customization to a build arg. This is so that in a sequence of commands of `ansible-builder build` and then later `podman build`, the user can apply a custom setting to `podman build` as needed. This approach makes it obvious how to support a custom `builder` image, which is to make it a build argument in the syntax of the Containerfile/Dockerfile.

Sounds good!

AlanCoding:
[...]

PR #167, opened 2021-01-05
Add support for build args (version 2)

AlanCoding:
This is a continuation of https://github.com/ansible/ansible-builder/pull/160 with me adding my wishlist of things onto it.

Addresses some of https://github.com/ansible/ansible-builder/issues/159

### DIFF
--- a/ansible_builder/constants.py
+++ b/ansible_builder/constants.py
@@ -11,9 +11,10 @@ default_container_runtime = 'podman'
 base_roles_path = '/usr/share/ansible/roles'
 base_collections_path = '/usr/share/ansible/collections'
 
-# These can be overwritten by build args
-default_base_image = 'quay.io/ansible/ansible-runner:devel'
-builder_stage_image = 'quay.io/ansible/python-builder:latest'
+build_arg_defaults = dict(
+    ANSIBLE_RUNNER_BASE_IMAGE='quay.io/ansible/ansible-runner:devel',
+    PYTHON_BUILDER_IMAGE='quay.io/ansible/python-builder:latest'
+)
 
 if shutil.which('podman'):
     default_container_runtime = 'podman'
--- a/ansible_builder/constants.py
+++ b/ansible_builder/constants.py
@@ -1,7 +1,6 @@
 import shutil
 
 default_file = 'execution-environment.yml'
-default_base_image = 'quay.io/ansible/ansible-runner:devel'
 default_tag = 'ansible-execution-env:latest'
 default_build_context = 'context'
 runtime_files={
@@ -11,6 +10,10 @@ runtime_files={
 default_container_runtime = 'podman'
 base_roles_path = '/usr/share/ansible/roles'
 base_collections_path = '/usr/share/ansible/collections'
+
+# These can be overwritten by build args
+default_base_image = 'quay.io/ansible/ansible-runner:devel'
+builder_stage_image = 'quay.io/ansible/python-builder:latest'
 
 if shutil.which('podman'):
     default_container_runtime = 'podman'
--- a/ansible_builder/constants.py
+++ b/ansible_builder/constants.py
@@ -12,7 +12,7 @@ base_roles_path = '/usr/share/ansible/roles'
 base_collections_path = '/usr/share/ansible/collections'
 
 build_arg_defaults = dict(
-    ANSIBLE_RUNNER_BASE_IMAGE='quay.io/ansible/ansible-runner:devel',
+    ANSIBLE_RUNNER_IMAGE='quay.io/ansible/ansible-runner:devel',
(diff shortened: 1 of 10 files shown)


ansible_builder/main.py
```diff
@@ -38,30 +37,21 @@
...
                  verbosity=2):
         self.action = action
         self.definition = UserDefinition(filename=filename)
 
-        # Handle precedence of the base image
-        if base_image is not None:
-            self.base_image = base_image
-        if base_image is None:
-            if self.definition.raw.get('base_image'):
-                self.base_image = self.definition.raw.get('base_image')
```
AlanCoding, line 54: @beeankha I hat to just remove a feature, but if we're not taking the base image as a CLI option, then we shouldn't take it in the definition either.

If we still want some kind of support for this, I would suggest that we should support build args in the definition.

If a user gives a dict of build args in the definition file, then maybe those should become defaults in the Containerfile? I'm still noodling on that idea.

AlanCoding:
This effectively undoes https://github.com/ansible/ansible-builder/pull/129, which was a specific request in https://github.com/ansible/ansible-builder/issues/119

Again, I think the solution would be to allow maybe... build arg defaults(?) in the definition file. I'm not sure if I should try to tackle that right now or not. Part of the problem is that it makes no sense to specify build args in that way except for the 2 we use.

AlanCoding:
This should be relatively complete now, and I'm just debugging from here on.

AlanCoding:
This is still a little ugy:

```
$ ansible-builder build -c alan -f test/data/build_args/execution-environment.yml -t alan --container-runtime=podman -v3
```

```
$ cat alan/Containerfile 
ARG ANSIBLE_RUNNER_BASE_IMAGE=quay.io/ansible/ansible-runner:devel
ARG PYTHON_BUILDER_IMAGE=quay.io/ansible/python-builder:latest

FROM $ANSIBLE_RUNNER_BASE_IMAGE as galaxy


FROM $PYTHON_BUILDER_IMAGE as builder

FROM $ANSIBLE_RUNNER_BASE_IMAGE
ARG FOO
RUN echo $FOO
```

We don't have any means of excluding these stages if there's no use for it.

----

Otherwise, I've tested a few of the integration tests failing from the last round, and they seem to have been fixed. Marking this as ready for review, but will still be checking back for results when they finish.

beeankha (approved):

AlanCoding:
Looking at the failures:

```

2021-01-08 19:40:45.945074 \| ubuntu-bionic \| =========================== short test summary info ============================
--
1546 | 2021-01-08 19:40:45.945087 \| ubuntu-bionic \| FAILED test/integration/test_build.py::TestPytz::test_build_layer_reuse[podman]
1547 | 2021-01-08 19:40:45.945103 \| ubuntu-bionic \| ERROR test/integration/test_build.py::TestPytz::test_build_layer_reuse[docker]
1548 | 2021-01-08 19:40:45.945130 \| ubuntu-bionic \| ERROR test/integration/test_help.py::test_no_args[podman] - Failed: Running p...
1549 | 2021-01-08 19:40:45.945173 \| ubuntu-bionic \| ======== 1 failed, 37 passed, 5 skipped, 2 errors in 2903.67s (0:48:23) ========
```

These should get resolved by https://github.com/ansible/ansible-builder/pull/169

ansible_builder/constants.py
```diff
@@ -12,6 +11,11 @@
 base_roles_path = '/usr/share/ansible/roles'
 base_collections_path = '/usr/share/ansible/collections'
 
+build_arg_defaults = dict(
+    ANSIBLE_RUNNER_BASE_IMAGE='quay.io/ansible/ansible-runner:devel',
```
pabelanger: what are your thoughts on dropping 'base' here and using

   ANSIBLE_RUNNER_IMAGE
   
This is in-line with PYTHON_BUILDER_IMAGE below. because there is a PYTHON_BASE_IMAGE today and we can avoid overloading the idea of a 'base' image.

pabelanger: This also comes into play with ansible-base / ansible-core too.  So much base!

ansible_builder/main.py
```diff
@@ -293,12 +313,18 @@ def __init__(self, definition,
...
         filename = constants.runtime_files[container_runtime]
         self.path = os.path.join(self.build_context, filename)
-        self.base_image = base_image
-        self.builder_stage_image = 'quay.io/ansible/python-builder:latest'
         self.container_runtime = container_runtime
         self.tag = tag
+        # Build args all need to go at top of file to avoid errors
         self.steps = [
-            "FROM {0} as galaxy".format(self.base_image),
+            "ARG ANSIBLE_RUNNER_BASE_IMAGE={}".format(
```
pabelanger: same comment as above

pabelanger (approved):
This looks great, thanks!

PR #170, opened 2021-01-11
Try to surface some race condition errors better

AlanCoding:


AlanCoding:
If the function scope change doesn't break any more, then I'm actually relatively happy with this set of changes and would prefer to merge and then continue work off of that.

PR #172, opened 2021-01-12
Copy the entire Ansible share dir for bug workaround

AlanCoding:
I'm still testing this locally with the awx-ee, but the hope is that it is a workaround for https://github.com/containers/podman/issues/8948

replaces prior attempt https://github.com/ansible/ansible-builder/pull/171

Credit to @pabelanger for seeing the obvious solution I didn't.

### DIFF
--- a/ansible_builder/steps.py
+++ b/ansible_builder/steps.py
@@ -1,4 +1,5 @@
 import sys
+import os
 
 from . import constants
 from .exceptions import DefinitionError
@@ -58,8 +59,9 @@ class GalaxyCopySteps(Steps):
         self.steps = []
         self.steps.extend([
             "",
-            "COPY --from=galaxy {0} {0}".format(constants.base_roles_path),
-            "COPY --from=galaxy {0} {0}".format(constants.base_collections_path),
+            "COPY --from=galaxy {0} {0}".format(
+                os.path.dirname(constants.base_collections_path.rstrip('/'))  # /usr/share/ansible
+            ),
             "",
         ])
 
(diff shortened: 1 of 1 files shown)


ansible_builder/steps.py
```diff
@@ -58,8 +59,9 @@ def __init__(self):
         self.steps = []
         self.steps.extend([
             "",
-            "COPY --from=galaxy {0} {0}".format(constants.base_roles_path),
-            "COPY --from=galaxy {0} {0}".format(constants.base_collections_path),
+            "COPY --from=galaxy {0} {0}".format(
+                os.path.dirname(constants.base_collections_path.rstrip('/'))  # /usr/share/ansible
+            ),
```
pabelanger, line 62: I actually think we can also remove line 51 above, there is no longer the need to create the collections / roles folders, as we can default to ansible-galaxy doing it.

AlanCoding:
Verified that this is working for the awx-ee.

```
ansible-builder build -c . -t quay.io/ansible/awx-ee:latest -v3
```

Got to the final

```
Complete! The build context can be found at: .
```

ansible_builder/steps.py
pabelanger: @AlanCoding making sure you seen this comment.

AlanCoding: Yeah, I was still testing rebuilding the awx-ee and the blank example. It looks like both are fine with this change, so I'll push a commit.

pabelanger (approved):
Looks great

PR #173, opened 2021-01-12
Add ANSIBLE_GALAXY_CLI_COLLECTION_OPTS build-arg

pabelanger:
This allows a user to use the --pre flag for ansible-galaxy collection
install. This will allow the installation of pre-release collections
from galaxy / AH.

### DIFF
--- a/ansible_builder/steps.py
+++ b/ansible_builder/steps.py
@@ -44,7 +44,7 @@ class GalaxyInstallSteps(Steps):
             "",
             "RUN ansible-galaxy role install -r /build/{0} --roles-path {1}".format(
                 requirements_naming, constants.base_roles_path),
-            "RUN ansible-galaxy collection install -r /build/{0} --collections-path {1}".format(
+            "RUN ansible-galaxy collection install $ANSIBLE_GALAXY_CLI_COLLECTION_OPTS -r /build/{0} --collections-path {1}".format(
                 requirements_naming, constants.base_collections_path),
             "",
             "RUN mkdir -p {0} {1}".format(constants.base_roles_path, constants.base_collections_path),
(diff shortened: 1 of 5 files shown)


ansible_builder/steps.py
```diff
@@ -37,14 +37,13 @@ def __init__(self, requirements_naming):
         """Assumes given requirements file name has been placed in the build context
         """
         self.steps = []
-        self.steps.append(
-            "ADD {0} /build/{0}".format(requirements_naming)
-        )
         self.steps.extend([
+            "ARG ANSIBLE_GALAXY_CLI_COLLECTION_OPTS=\"\"",
```
AlanCoding: Right now, this default of `""` can't be changed with `build_arg_defaults` in the execution environment definition, and maybe it arguably should be customizable there.

AlanCoding (approved):
To do these defaults right, I think we should make a new class in the `steps` module, because the variable scope isn't right for it right now.

Love how the documentation looks, I'm good with this. Maybe a test would be good?

pabelanger:
Pull-request updated, HEAD is now c639cb315ce3ea6630a2db0273d0d55aae0b6fe3

pabelanger:
> To do these defaults right, I think we should make a new class in the `steps` module, because the variable scope isn't right for it right now.
> 
> Love how the documentation looks, I'm good with this. Maybe a test would be good?

Yah, I noticed it wasn't scoped right. Let me know what you think of this version. I'll look at a test.

pabelanger:
Pull-request updated, HEAD is now bd49424587847b5308b8ac07f3fe12f1f410b8e4

AlanCoding (approved):
:shipit:

PR #175, opened 2021-01-12
Delete test images test-by-test for multiprocessing concerns

AlanCoding:
still no idea if this will work

PR #176, opened 2021-01-13
Only teardown images if final process

AlanCoding:
Alright, so this is a little complicated.

Our integration tests keep failing with "layer not found" or "image not found" kind of errors. This is because the integration tests were never written to be ran in parallel, which is what we're doing in Zuul.

Tests run with 2 processes, and pytest multiprocessing isn't very smart about distributing the tests. So 1 process finishes while the other has a significant number of tests remaining.

The problem we have is that the first process to finish runs the teardown method of its autouse session-scoped fixture. This deletes all images that we work with, while the other process is running. Clearly podman does not like this, and everything fails after that point.

This is a quirky solution to track which processes are running and which are finished. Only the final process to finish will run the command to delete images.

However, I have realized that https://github.com/ansible/ansible-builder/pull/175 might have also fixed the issue, so I'll get results and then maybe we can choose either.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -1,5 +1,8 @@
 import os
 import subprocess
+from pathlib import Path
+from glob import glob
+import shutil
 
 import tempfile
 import uuid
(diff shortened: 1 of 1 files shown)


AlanCoding:
Looks like something about this logic didn't quite work in Zuul.

AlanCoding:
going with the other one

Issue #122, opened 2020-11-10
Add support to install ansible collection as tarball


pabelanger:
In some cases, we want to be able to install a collection tarball from the local file system. This could be before we don't have access to galaxy or we are testing an unreleased version of code.

We do this today by not using a requirement file, and use the CLI directly
```
ansible-galaxy collection install  vyos-vyos-1.1.1-dev2.tar.gz ansible-netcommon-1.4.2-dev1.tar.gz
```

AlanCoding:
I was kind of excited about the download command, demo with awx-ee:

```
ansible-galaxy collection download -r _build/requirements.yml -p alan
```

produces

```
$ tree alan/
alan/
├── amazon-aws-1.3.0.tar.gz
├── ansible-netcommon-1.4.1.tar.gz
├── ansible-posix-1.1.1.tar.gz
├── awx-awx-16.0.0.tar.gz
├── azure-azcollection-1.3.1.tar.gz
├── community-kubernetes-1.1.1.tar.gz
├── community-vmware-1.6.0.tar.gz
├── google-cloud-1.0.1.tar.gz
├── openstack-cloud-1.0.1.tar.gz
├── ovirt-ovirt-1.2.5.tar.gz
├── requirements.yml
└── theforeman-foreman-1.5.1.tar.gz

0 directories, 12 files
```

```
$ cat alan/requirements.yml 
collections:
- name: awx-awx-16.0.0.tar.gz
  version: 16.0.0
- name: azure-azcollection-1.3.1.tar.gz
  version: 1.3.1
- name: amazon-aws-1.3.0.tar.gz
  version: 1.3.0
- name: theforeman-foreman-1.5.1.tar.gz
  version: 1.5.1
- name: google-cloud-1.0.1.tar.gz
  version: 1.0.1
- name: openstack-cloud-1.0.1.tar.gz
  version: 1.0.1
- name: community-vmware-1.6.0.tar.gz
  version: 1.6.0
- name: ovirt-ovirt-1.2.5.tar.gz
  version: 1.2.5
- name: community-kubernetes-1.1.1.tar.gz
  version: 1.1.1
- name: ansible-posix-1.1.1.tar.gz
  version: 1.1.1
- name: ansible-netcommon-1.4.1.tar.gz
  version: 1.4.1
```

There are likely to be multiple ways to use this feature, and I would love if we had some way to support installing from the requirements file from `download` like this.

AlanCoding:
[...]

PR #180, opened 2021-01-14
Copy build context folder into galaxy layer

pabelanger:
After looking at how to support offline tarballs, it might be as simple
as copying the full build context into the galaxy layer. This is okay,
becaue the layer is only used at build time.

It also allows a user to add their our data, like collection tarballs,
in the sub-folder.  Allowing for offline install of tarballs.

Depends-On: https://github.com/ansible/ansible-builder/pull/182

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -20,9 +20,6 @@ CONTEXT_FILES = {
     'galaxy': 'requirements.yml'
 }
 
-# Subdirectory in the build context where files to ADD get copied
-CONTEXT_BUILD_OUTPUTS_DIR = '_build'
-
 BINDEP_COMBINED = 'bindep_combined.txt'
 PIP_COMBINED = 'requirements_combined.txt'
 
@@ -48,7 +45,8 @@ class AnsibleBuilder:
 
         self.tag = tag
         self.build_context = build_context
-        self.build_outputs_dir = os.path.join(build_context, CONTEXT_BUILD_OUTPUTS_DIR)
+        self.build_outputs_dir = os.path.join(
+            build_context, constants.user_content_subfolder)
         self.container_runtime = container_runtime
         self.build_args = build_args or {}
         self.containerfile = Containerfile(
@@ -310,7 +308,8 @@ class Containerfile:
                  tag=None):
 
         self.build_context = build_context
-        self.build_outputs_dir = os.path.join(build_context, CONTEXT_BUILD_OUTPUTS_DIR)
+        self.build_outputs_dir = os.path.join(
+            build_context, constants.user_content_subfolder)
         self.definition = definition
         filename = constants.runtime_files[container_runtime]
         self.path = os.path.join(self.build_context, filename)
@@ -342,7 +341,8 @@ class Containerfile:
             requirement_path = self.definition.get_dep_abs_path(item)
             if requirement_path is None:
                 continue
-            dest = os.path.join(self.build_context, CONTEXT_BUILD_OUTPUTS_DIR, new_name)
+            dest = os.path.join(
+                self.build_context, constants.user_content_subfolder, new_name)
             copy_file(requirement_path, dest)
 
         if self.definition.ansible_config:
@@ -354,7 +354,8 @@ class Containerfile:
     def prepare_ansible_config_file(self):
         ansible_config_file_path = self.definition.ansible_config
         if ansible_config_file_path:
-            context_file_path = os.path.join(CONTEXT_BUILD_OUTPUTS_DIR, 'ansible.cfg')
+            context_file_path = os.path.join(
+                constants.user_content_subfolder, 'ansible.cfg')
             return self.steps.extend(AnsibleConfigSteps(context_file_path))
 
     def prepare_prepended_steps(self):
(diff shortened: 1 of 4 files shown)


AlanCoding:
A concern I have is relative paths. The `execution-environment.yml` definition is passed by the `-f` option and the context is passed by the `-c` option. We don't really know where they are, and the entry for `dependencies.galaxy` is a `requirements.yml` that may or may not get copied over into the build context.

Since paths in `requirements.yml` are relative to _that_ file's path, this should work specifically in the case that the user locates their file at `_build/requirements.yml` from the context directory they give.

This still seems more aggressive than it needs to be. The same behavior could essentially be obtained by only copying over the `_build` subfolder in the build context. I wouldn't want to document saving your tar files anywhere else.

ansible_builder/steps.py
```diff
@@ -34,19 +34,21 @@ def __iter__(self):
...
 class GalaxyInstallSteps(Steps):
-    def __init__(self, requirements_naming):
+    def __init__(self, build_context, requirements_naming):
         """Assumes given requirements file name has been placed in the build context
         """
+        relative_galaxy_path = os.path.join(build_context, requirements_naming)
         self.steps = []
         self.steps.append(
-            "ADD {0} /build/{0}".format(requirements_naming)
+            "ADD {0} /build/{0}".format(build_context)
```
AlanCoding: Translating my suggestion...

```suggestion
            "ADD _build/{0} /build/_build/{0}".format(os.path.join(CONTEXT_BUILD_OUTPUTS_DIR, build_context))
```

pabelanger: Sorry, I am a little confused.

1) you are hardcoding _build here, over using CONTEXT_BUILD_OUTPUTS_DIR
2) build_context is already CONTEXT_BUILD_OUTPUTS_DIR to this could be doubled.

AlanCoding: Oh I got confused by the variable name.

```python
CONTEXT_BUILD_OUTPUTS_DIR = '_build'
```

That's just a string which is the name of a folder inside of the build context. So you're right, my suggestion doesn't make sense.

AlanCoding: Yeah, especially with this change we need a naming face-lift. Spitballing some things:

Move `CONTEXT_BUILD_OUTPUTS_DIR` to the constants module, and name it `USER_CONTENT_SUBFOLDER`.

Don't pass that as an argument, if the `steps` module needs it (and you don't want to hard code it), then import it.

AlanCoding:
Connect https://github.com/ansible/ansible-builder/issues/122

ansible_builder/steps.py
pabelanger: ++

pabelanger:
Pull-request updated, HEAD is now 492e9f98c32ce368e8bb4f445ecea7dc6bf9cbb4

ansible_builder/steps.py
```diff
@@ -37,16 +37,20 @@ class GalaxyInstallSteps(Steps):
     def __init__(self, requirements_naming):
         """Assumes given requirements file name has been placed in the build context
         """
+        relative_galaxy_path = os.path.join(
+            constants.user_content_subfolder, requirements_naming)
         self.steps = []
         self.steps.append(
-            "ADD {0} /build/{0}".format(requirements_naming)
+            "ADD {0} /build/{0}".format(
+                constants.user_content_subfolder)
```
pabelanger: to me, this is a little funky. It ends up being:
```
ADD _build /build/_build
```

I think I would rather see it become:
```
ADD _build /build
```
or
```
ADD _build /tmp/_build
```
I'm leaning towards /tmp, given these really are temp files.

ansible_builder/main.py
```diff
@@ -382,22 +383,23 @@ def prepare_appended_steps(self):
 
     def prepare_galaxy_install_steps(self):
         if self.definition.get_dep_abs_path('galaxy'):
-            relative_galaxy_path = os.path.join(CONTEXT_BUILD_OUTPUTS_DIR, CONTEXT_FILES['galaxy'])
```
pabelanger, line 380: this can be reverted

ansible_builder/main.py
```diff
@@ -382,22 +383,23 @@ def prepare_appended_steps(self):
...
             self.steps.append(
                 "ARG ANSIBLE_GALAXY_CLI_COLLECTION_OPTS={}".format(
                     self.definition.build_arg_defaults['ANSIBLE_GALAXY_CLI_COLLECTION_OPTS']))
-            self.steps.extend(GalaxyInstallSteps(relative_galaxy_path))
```
pabelanger, line 384: same

ansible_builder/steps.py
```diff
@@ -37,16 +37,20 @@ class GalaxyInstallSteps(Steps):
     def __init__(self, requirements_naming):
         """Assumes given requirements file name has been placed in the build context
         """
+        relative_galaxy_path = os.path.join(
+            constants.user_content_subfolder, requirements_naming)
```
pabelanger: we can remove this

pabelanger:
Pull-request updated, HEAD is now 09e990500943bf0c9243dcde1a13eb3a80e975aa

ansible_builder/steps.py
AlanCoding: > ADD _build /build/_build

Yes, what I expected.

> I'm leaning towards /tmp, given these really are temp files.

If they're temporary files, are they deleted in the final image? Or is our disposition that we want to keep them so that user can debug problems?

pabelanger: They will never get into the final image, but it is more what the requirements.yaml file looks like for offline collections.

Today it would be
```
- name: /build/_build/ansible-netcommon-1.4.1.tar.gz
- name: /build/_build/foo-1-1-1.tar.gz
```

AlanCoding (approved):
This was always the logical consequence of introducing the `_build` folder.

pabelanger:
Pull-request updated, HEAD is now a3c01c3961aee36dcbcfac6d62c485eaf4b4dbc8

pabelanger:
Pull-request updated, HEAD is now b12bb87e015e5c2374b311355d9794bb0fb32c93

pabelanger:
Pull-request updated, HEAD is now 3ffa3fcfecc71f01adcefc5f385e5c6e9b65b059

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
Pull-request updated, HEAD is now 11127c74eeab255abe5ace61d70179721cd4a488

PR #182, opened 2021-01-15
Allow user to toggle image clean up for integration tests

pabelanger:
It is possible, a user want to debug a container created via integration
tsts.  Also, we may want to disable clean ups, in the case of CI.

Either way, add a flag to toggle the deleting of images.

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -16,3 +16,5 @@
     vars:
       tox_envlist: integration
       tox_install_siblings: false
+      tox_environment:
+        KEEP_IMAGES: true
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -7,11 +7,10 @@ import uuid
 import pytest
 import logging
 
-
 logger = logging.getLogger(__name__)
 
-
 TAG_PREFIX = 'quay.io/example/builder-test'
+KEEP_IMAGES = bool(os.environ.get('KEEP_IMAGES', False))
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


test/integration/conftest.py
```diff
@@ -7,11 +7,10 @@
 import pytest
 import logging
 
-
 logger = logging.getLogger(__name__)
 
-
 TAG_PREFIX = 'quay.io/example/builder-test'
+KEEP_IMAGES = bool(os.environ.get('KEEP_IMAGES', False))
```
AlanCoding, line 13: The annoying problem with these is

```
$ KEEP_IMAGES=false python -c "import os; print(bool(os.environ.get('KEEP_IMAGES')))"
True
$ KEEP_IMAGES=0 python -c "import os; print(bool(os.environ.get('KEEP_IMAGES')))"
True
```

It'll still work for our purposes, I just wanted to point it out.

AlanCoding: where `distutils.util.strtobool` is the common workaround

AlanCoding (approved):

pabelanger:
Pull-request updated, HEAD is now 0841a92572007be8c40217c4655819b480e40aae

pabelanger:
recheck

.zuul.d/jobs.yaml
```diff
@@ -16,3 +16,5 @@
     vars:
       tox_envlist: integration
       tox_install_siblings: false
+      tox_environment:
+        KEEP_IMAGES: true
```
AlanCoding, line 20: Should this be reversed to `false`? We are running an ephemeral container, we do not want to keep images because it won't matter.

AlanCoding: no, I was wrong, never mind

PR #183, opened 2021-01-15
Delete integration test tags after all tests run

pabelanger:
With pytest, we can use pytest_unconfigure to delete_images() after all
tests have finished running. This will help avoid race conditions when
deleting images.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import subprocess
 
 import tempfile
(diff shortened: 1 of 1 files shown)


test/integration/conftest.py
```diff
@@ -71,29 +72,31 @@ def gen_image_name(request):
...
+    cmd = (
+      f"{container_runtime} images | grep {TAG_PREFIX} | cut -d ' ' -f 1 | xargs {container_runtime} rmi --force")
+    run(cmd, allow_error=True)
+
+
+def pytest_unconfigure(config):
+    if shutil.which('podman'):
+        delete_images('podman')
+    else:
+        delete_images('docker')
```
AlanCoding, line 87: it's possible to run integration tests for _both_ podman and docker

pabelanger: it is, but they will both create tags.  Rather then deleting images with both runtimes, we can just pick one. To delete all tags.

AlanCoding:
I went through a solid handful of revisions of this myself. I am worried about the consequences, and I am extremely partial to our existing solution where we just don't delete the tags for Zuul tests. The problem for local developers is solved by the existing fixture in a way that better handles errors.

There's a lot more exciting things going on, so I hope you don't mind that I go ahead and close

PR #194, opened 2021-03-04
Include collection metadata in LABEL

shanemcd:


ansible_builder/main.py
```diff
@@ -428,12 +430,18 @@ def prepare_build_stage_steps(self):
...
             "",
             "FROM $ANSIBLE_RUNNER_IMAGE"
             "",
         ])
+        if collections:
+            self.steps.extend([
+                "",
+                "LABEL collections='{}'".format(json.dumps(collections)),
+                "",
+            ])
```
pabelanger, line 444: Another idea, is to remove this from the Dockerfile, then use the --label flag on the command line. That will atleast remove this blob from the file, and each time ansible-builder runs, it gets setup.

Then users or docker / podman would still need to pass the format some how.

shanemcd: I think you must have left this comment before we spoke in Slack, but for posterity I'll reiterate here: we don't have control over the invocation of `docker build` in our downstream build system.

ansible_builder/main.py
```diff
@@ -428,12 +430,18 @@ def prepare_build_stage_steps(self):
...
+    def prepare_final_stage_steps(self, collections):
         self.steps.extend([
             "",
             "FROM $ANSIBLE_RUNNER_IMAGE"
             "",
         ])
+        if collections:
+            self.steps.extend([
+                "",
+                "LABEL collections='{}'".format(json.dumps(collections)),
```
newswangerd, line 442: I would recommend using `com.ansible.execution-environment.collections` or `com.ansible.collections` as the label for this, as per the docker recommendations: https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations.

We'll be using these labels to determine if an image is an EE, so making sure that they don't collide with labels other tools might use is important. `collections` is a pretty generic term and it could be used by other software.

Some other metadata should definitely be included:
- schema version, if you have an official schema for EE images, or at the very least the ansible-builder version. That way we can handle cases appropriately if there are any known bugs or format changes between different versions of builder.
- ansible version. Users will want to know which version of ansible is being used in the EE to help identify bugs and resolve CVEs

Some metadata that would be really nice to have but aren't critical:
- collection source: this would allow us to like to the source or documentation for a collection
- other dependencies: knowing what python or other binary deps are in an EE could also be helpful for identifying bugs or security issues

If you don't want to add all of this as different labels, another potential solution would be us use one label such as `com.ansible.execution-environment.metadata` and put all of this information in a json blob there. That would have the added benefit making it so there are fewer values to check when parsing the container config.

ryanpetrello:
👋 @cidrblock you were asking me about this the other day

shanemcd:
This idea has been abandoned.

PR #205, opened 2021-04-15
Explicitly flip to root user during build process

shanemcd:
This fixes an issue where the parent image has set USER to something other than root

See https://github.com/ansible/awx/issues/9917

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -2,7 +2,7 @@
 - job:
     name: ansible-builder-tox-integration
     parent: ansible-buildset-registry-consumer
-    timeout: 3600
+    timeout: 4500
     vars:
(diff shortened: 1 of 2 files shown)


shanemcd:
recheck

AlanCoding (approved):

shanemcd:
regate

.zuul.d/jobs.yaml
```diff
@@ -2,7 +2,7 @@
 - job:
     name: ansible-builder-tox-integration
     parent: ansible-buildset-registry-consumer
-    timeout: 3600
+    timeout: 4500
```
pabelanger, line 5: I actually think we need to set timeout in the job below too. Which is surprising, we might be hitting a bug here.

PR #216, opened 2021-05-20
Change from a two-stage to a single-stage build process with `ansible-builder introspect` layer

AlanCoding:
Up until now, `ansible-builder` would output an EE Containerfile that used 2 parent images and 3 stages. The stages were `galaxy`, `builder`, and unnamed.

This PR would change it to 2 parent images and 2 stages. Galaxy operations are moved into the `builder` stage. To enable this, we introduce a new `quay.io/ansible/ansible-builder` image, which combines:
 - scripts from `ansible/python-builder`
 - everything else from `ansible/ansible-runner`
 - installs `ansible-builder` itself

The `ansible/ansible-builder` image will not be a part of the final EE (as you would expect from any "builder") image.

This renames the `--build-args`
 - `ANSIBLE_RUNNER_IMAGE` to `EE_BASE_IMAGE` (but still references the same image)
 - `PYTHON_BUILDER_IMAGE` to `EE_BUILDER_IMAGE` (and switches from `python-builder` image to `ansible-builder`)

### Links

I resurrected old code from 76154e09b5a7a887884caeaa515fdc626db6c61d

The `ansible/python-builder` scripts are COPYed into the new image, those are from:

https://github.com/ansible/python-builder-image/tree/main/scripts

There was concern in the planning of this that we will need `python38-wheel` from the `ansible/python-builder` [Containerfile](https://github.com/ansible/python-builder-image/blob/main/Containerfile). That would be a maintenance headache. This might still be right, but I don't have a test case that fails it.

----

### Representative set of examples from `test/data` folder:

| Feature of builder | Relevant example |
|--------------------|------------------|
| user bindep        | subversion       |
| user python        | pip              |
| collection bindep  | ansible.posix.at |
| collection python  | pytz             |
 
Result from running cases

#### pytz

```Dockerfile
ARG EE_BASE_IMAGE=quay.io/ansible/ansible-runner:devel
ARG EE_BUILDER_IMAGE=quay.io/ansible/ansible-builder:latest

FROM $EE_BUILDER_IMAGE as builder
USER root

ADD _build /build
WORKDIR /build

ARG ANSIBLE_GALAXY_CLI_COLLECTION_OPTS=
RUN ansible-galaxy role install -r requirements.yml --roles-path /usr/share/ansible/roles
RUN ansible-galaxy collection install $ANSIBLE_GALAXY_CLI_COLLECTION_OPTS -r requirements.yml --collections-path /usr/share/ansible/collections
RUN ansible-builder introspect --write-bindep=/tmp/src/bindep.txt --write-pip=/tmp/src/requirements.txt
RUN assemble

FROM $EE_BASE_IMAGE
USER root

COPY --from=builder /usr/share/ansible /usr/share/ansible

```

#### pip

```Dockerfile
ARG EE_BASE_IMAGE=quay.io/ansible/ansible-runner:devel
ARG EE_BUILDER_IMAGE=quay.io/ansible/ansible-builder:latest

FROM $EE_BUILDER_IMAGE as builder
USER root

ADD _build /build
WORKDIR /build

RUN ansible-builder introspect --user-pip=requirements.txt --write-bindep=/tmp/src/bindep.txt --write-pip=/tmp/src/requirements.txt
RUN assemble

FROM $EE_BASE_IMAGE
USER root
```

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -14,7 +14,7 @@ FROM $ANSIBLE_RUNNER_IMAGE
 # =============================================================================
 
 COPY --from=builder /output/ /output
-RUN /output/install-from-bindep rm -rf /output
+RUN /output/install-from-bindep && rm -rf /output
 
 # move the assemble scripts themselves into this container
 COPY --from=builder /usr/local/bin/assemble /usr/local/bin/assemble
--- a/Containerfile
+++ b/Containerfile
@@ -1,7 +1,7 @@
 ARG ANSIBLE_RUNNER_IMAGE=quay.io/ansible/ansible-runner:devel
 ARG PYTHON_BUILDER_IMAGE=quay.io/ansible/python-builder:latest
 
-FROM $PYTHON_BUILDER_IMAGE as builder
+FROM $PYTHON_BUILDER_IMAGE as python_builder
 # =============================================================================
 ARG ZUUL_SIBLINGS
 
@@ -13,9 +13,12 @@ RUN assemble
 FROM $ANSIBLE_RUNNER_IMAGE
 # =============================================================================
 
-COPY --from=builder /output/ /output
+COPY --from=python_builder /output/ /output
 RUN /output/install-from-bindep && rm -rf /output
 
 # move the assemble scripts themselves into this container
-COPY --from=builder /usr/local/bin/assemble /usr/local/bin/assemble
-COPY --from=builder /usr/local/bin/get-extras-packages /usr/local/bin/get-extras-packages
+COPY --from=python_builder /usr/local/bin/assemble /usr/local/bin/assemble
+COPY --from=python_builder /usr/local/bin/get-extras-packages /usr/local/bin/get-extras-packages
+
+# building EEs require the install-from-bindep script, but not the rest of the /output folder
+COPY --from=python_builder /output/install-from-bindep /output/install-from-bindep
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,21 @@
+ARG ANSIBLE_RUNNER_IMAGE=quay.io/ansible/ansible-runner:devel
+ARG PYTHON_BUILDER_IMAGE=quay.io/ansible/python-builder:latest
+
+FROM $PYTHON_BUILDER_IMAGE as builder
+# =============================================================================
+ARG ZUUL_SIBLINGS
+
+# install this library (meaning ansible-builder)
+COPY . /tmp/src
+RUN assemble
+
+
+FROM $ANSIBLE_RUNNER_IMAGE
+# =============================================================================
+
+COPY --from=builder /output/ /output
+RUN /output/install-from-bindep rm -rf /output
+
+# move the assemble scripts themselves into this container
+COPY --from=builder /usr/local/bin/assemble /usr/local/bin/assemble
+COPY --from=builder /usr/local/bin/get-extras-packages /usr/local/bin/get-extras-packages
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -9,9 +9,9 @@ from . import __version__, constants
 from .colors import MessageColors
 from .exceptions import DefinitionError
 from .main import AnsibleBuilder
-from .introspect import add_introspect_options, process, simple_combine
+from .introspect import process, simple_combine, base_collections_path
 from .requirements import sanitize_requirements
-from .utils import configure_logger
+from .utils import configure_logger, write_file
 
 
 logger = logging.getLogger(__name__)
@@ -37,21 +37,21 @@ def run():
             sys.exit(1)
     elif args.action == 'introspect':
         data = process(args.folder)
-        if args.sanitize or args.raw:
+        if args.sanitize:
             data['python'] = sanitize_requirements(data['python'])
             data['system'] = simple_combine(data['system'])
             logger.info('# Sanitized dependencies for {0}'.format(args.folder))
         else:
             print('# Dependency data for {0}'.format(args.folder))
 
-        if args.raw:
-            if args.type[0] == 'python':
-                print('\n'.join(data['python']))
-            elif args.type[0] == 'system':
-                print('\n'.join(data['system']))
-        else:
-            print('---')
-            print(yaml.dump(data, default_flow_style=False))
+        print('---')
+        print(yaml.dump(data, default_flow_style=False))
+
+        if args.write_pip and data.get('python'):
+            write_file(args.write_pip, simple_combine(data.get('python')) + [''])
+        if args.write_bindep and data.get('system'):
+            write_file(args.write_bindep, simple_combine(data.get('system')) + [''])
+
         sys.exit(0)
 
     logger.error("An error has occured.")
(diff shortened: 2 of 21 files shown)


ansible_builder/main.py
```diff
@@ -383,6 +328,10 @@ def prepare_appended_steps(self):
 
         return False
 
+    def prepare_build_context(self):
+        self.steps.extend(BuildContextSteps())
```
AlanCoding: Maaaybe this step could be conditionally skipped in the case that:
 - there is no user-provided python or system requirements file
 - there is no user-provided galaxy requirements.yml file

But if that's the case, why are you using `ansible-builder` in the first place?

ansible_builder/cli.py
```diff
@@ -134,18 +141,40 @@ def parse_args(args=sys.argv[1:]):
...
+        'folder', default=base_collections_path, nargs='?',
+        help=(
+            'Ansible collections path(s) to introspect. '
+            'This should have a folder named ansible_collections inside of it.'
+        )
+    )
+    # TODO: If we can allow python-builder scripts to be fed multiple files
+    # then we should prefer that over the --user-* options
+    introspect_parser.add_argument(
+        '--user-pip', dest='user_pip',
```
AlanCoding, line 158: @pabelanger sometime in the past, I think you made the point that there's no need to combine `requirements.txt` files and/or `bindep.txt` files from collections with the user files. The scripts like `assemble` and `install-from-bindep` could just pass multiple files to `pip3` or `yum`.

Right now, the logic around [EXTRAS](https://github.com/ansible/python-builder-image/blob/5e8cac7ed3315c7bdc165488670fed27492e2fae/scripts/install-from-bindep#L61) does not seem to be enough to support that. I could be wrong, but I think we would need EXTRAS for assemble/install for both pip/bindep. Right now it looks like its only supported for pip-install.

I don't want to go adding feature requests to python-builder, and without that, these options are needed so that we combine the requirements files without the two-stage build.

pabelanger: ZUUL_SIBLIGS is how we handle it in the assemble script, which will then pass the install-from-bindep script the proper files.

https://github.com/ansible/python-builder-image/blob/5e8cac7ed3315c7bdc165488670fed27492e2fae/scripts/assemble#L112

Basically, we can pass in the list of collection directories, then iterate over that to install things.

AlanCoding: > Basically, we can pass in the list of collection directories

A list of paths under `.zuul-siblings/`, right? And that script puts us in `/tmp/src` before we get there.

So those directories have to be in the absolute location `/tmp/src/.zuul-siblings/`.

To appropriately set the stage, the collections are under `/usr/share/ansible/collections/`, with the subdirs known by `introspect`. The user content is in `/build`, but it may not be there.

But this is ignoring the big showstopper that we have encountered again and again. We could set `ENV ZUUL_SIBLINGS=...`, but how and when can we fill in that `...`? We can't view the contents of the collections until we get past the `ansible-galaxy` layer _in the Containerfile_. This is the fundamental tension of ansible-builder.

Maybe it could be done in a way that looks like this?

https://stackoverflow.com/a/37860420

There are a lot of unanswered questions and I have gone off the deep end speculating.

pabelanger: Yes, that is right. Remember, 'python-builder' was initially designed to build python projects in containers. So, from openstack project we expected requirements.txt / bindep.txt files to be in specific places (top level). So we never added any of the logic to allow projects to use custom locations.

What we likely should do, is update assemble a little to learn more about the project, like if a galaxy.yml file is found, then start looking for metadata files to point to the right locations.

I feel we could add the logic, but honestly I think we likely should create a script, that lives in ansible-builder which sets a lot of that logic up. Then invokes assemble.sh under the covers to minimize the Dockerfile size.

AlanCoding: > What we likely should do, is update assemble a little to learn more about the project, like if a galaxy.yml file is found, then start looking for metadata files to point to the right locations.

There's something here I love, and something I'm hesitant about.

The `requirements.yml` (galaxy requirements file) is not at all useful on its own. We can only do meaningful things with the _installed_ collections, and allow space for things like Automation Hub credentials. This is what ansible-builder does.

In the big picture, I would very much like to combine ansible-builder with python-builder. Taking a very long view, a 3-stage build (with python-builder / ansible-runner image build chain) could possibly satisfy both use cases... but there are valid reasons to think this may never be acceptable.

Within the current context, with the new single-stage, 2 phase build, and a new `ansible-builder` image, I still have some more runway. I could do a skunkworks version of providing ZUUL_SIBLINGS, but it'll have `introspect` doing a lot of ugly file moving operations.

AlanCoding:
I'm more and more thinking that it might actually be safer to keep this as a 3-stage Containerfile, but still get rid of the two-step overall ansible-builder process. It's more that I want to keep this agile, because of the additional changes that have been discussed, I don't have great confidence that I know what the final form will be.

Containerfile
```diff
@@ -0,0 +1,24 @@
...
+
+COPY --from=python_builder /output/ /output
+RUN /output/install-from-bindep && rm -rf /output
+
+# move the assemble scripts themselves into this container
+COPY --from=python_builder /usr/local/bin/assemble /usr/local/bin/assemble
+COPY --from=python_builder /usr/local/bin/get-extras-packages /usr/local/bin/get-extras-packages
+
+# building EEs require the install-from-bindep script, but not the rest of the /output folder
+COPY --from=python_builder /output/install-from-bindep /output/install-from-bindep
```
shanemcd, line 24: Won't we also need the `assemble` script too?

shanemcd: nvm, missed it above.

ansible_builder/exceptions.py
```diff
@@ -1,9 +1,8 @@
-import sys
 
 
 class DefinitionError(RuntimeError):
     # Eliminate the output of traceback before our custom error message prints out
-    sys.tracebacklimit = 0
+    #sys.tracebacklimit = 0
```
AlanCoding: need to clean this up

AlanCoding:
Applying this to the awx-ee

https://github.com/AlanCoding/awx-ee/compare/more_updates...AlanCoding:single_phase?expand=1

AlanCoding:
recheck

ansible_builder/cli.py
AlanCoding: @pabelanger trying to follow up here, the big-picture thing we talked about here is not something we can do in the near-term. But, I was worried that the current PR would not be highly compatible with the directions you and I are interested in taking things in. So I put this up:

https://github.com/ansible/ansible-builder/pull/217

The idea is that, one day, we may have a python-builder image that contains the builder "introspect" logic in addition to the scripts. That could be implemented with relative ease after that PR. It would be difficult to ever change to that after the changes here to the 2-phase Containerfile. At least, that's my primary motivation.

AlanCoding:
The ansible-runner repo runs tests with jobs like `ansible-runner-tox-ansible29`

https://github.com/ansible/ansible-zuul-jobs/blob/46cbb0481a6ae8c9cd3efd8bd19feeefbeac4abb/zuul.d/ansible-runner-jobs.yaml#L20

This is defined in a separate repo, which is why it slipped past me. I'm looking for some critical syntax where the check test job depends on the image build job, and then I need to know where the test job can _find_ the image from the prior step.

but... that definition doesn't have the critical bit of dependencies that I needed

https://github.com/ansible/project-config/blob/ba80198af07f07962bfe4e546e7da5fc4fd3fb22/zuul.d/projects.yaml#L804

doesn't have it either.

So in conclusion, I don't know how ansible-runner integration works, so I don't know how or if I can build something similar here.

AlanCoding:
Discussed and we have agreement that we're abandoning this in favor of the `[option 2]` PR

PR #219, opened 2021-05-26
Add wheel to list of requirements

shanemcd:
Needed for pre-building wheels inside of the ansible-builder image

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -16,6 +16,13 @@ FROM $PYTHON_BASE_IMAGE
 COPY --from=python_builder /output/ /output
 RUN /output/install-from-bindep && rm -rf /output
 
+RUN dnf update -y \
+  && dnf install -y python38-wheel git \
+  && dnf clean all \
+  && rm -rf /var/cache/{dnf,yum} \
+  && rm -rf /var/lib/dnf/history.* \
+  && rm -rf /var/log/*
+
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):

AlanCoding:
We may need to do `dnf install -y python38-wheel` instead. Unsure of specifics.

pabelanger:
Because we do https://github.com/ansible/python-builder-image/blob/main/Containerfile#L38 in python-builder, and those scripts use the packages, we need to also do it here.

We have to duplicate the data in the container file, or we install them at runtime inside the assemble script.

beeankha (approved):

Containerfile
```diff
@@ -16,6 +16,13 @@ FROM $PYTHON_BASE_IMAGE
 COPY --from=python_builder /output/ /output
 RUN /output/install-from-bindep && rm -rf /output
 
+RUN dnf update -y \
+  && dnf install -y python38-wheel git \
+  && dnf clean all \
+  && rm -rf /var/cache/{dnf,yum} \
+  && rm -rf /var/lib/dnf/history.* \
+  && rm -rf /var/log/*
+
```
pabelanger, line 25: actually, this isn't needed. Because we are using python-builder to build this image. We can add them into the bindep.txt file, and things will just work properly.

PR #225, opened 2021-05-26
Try to keep less layer churn in the /output folder

AlanCoding:
I have been encountering a lot of spooky action with local testing.

Basically, the `/tmp/src/requirements.txt` provably has the user content, but then after the assemble script runs, `/output/requirements.txt` contains the requirements for _ansible-builder_ itself, including requirements-parser.

This might not fix it, but it does avoid rewriting some files between multiple layers.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -12,11 +12,9 @@ FROM $PYTHON_BASE_IMAGE
 # =============================================================================
 
 COPY --from=builder /output/ /output
-RUN /output/install-from-bindep && rm -rf /output
+# building EEs require the install-from-bindep script, but not the rest of the /output folder
+RUN /output/install-from-bindep && find /output/* -not -name install-from-bindep -exec rm -rf {} +
 
 # move the assemble scripts themselves into this container
 COPY --from=builder /usr/local/bin/assemble /usr/local/bin/assemble
 COPY --from=builder /usr/local/bin/get-extras-packages /usr/local/bin/get-extras-packages
-
-# building EEs require the install-from-bindep script, but not the rest of the /output folder
-COPY --from=builder /output/install-from-bindep /output/install-from-bindep


Containerfile
```diff
@@ -14,7 +14,8 @@ FROM $PYTHON_BASE_IMAGE
 # =============================================================================
 
 COPY --from=python_builder /output/ /output
-RUN /output/install-from-bindep && rm -rf /output
+# building EEs require the install-from-bindep script, but not the rest of the /output folder
+RUN /output/install-from-bindep && cd /output && rm -rf !('install-from-bindep')
```
shanemcd: I don't think this is valid shell syntax?

But... I think I would prefer to do something like `mv install-from-bindep /usr/local/bin/` and then `rm -rf /output/*`

If you really want to do something like what you have, you could do:

```
$ find /output/* -not -name install-from-bindep -exec rm -rf {} +
``

Comment (AlanCoding) [reply Containerfile]: You're right, this doesn't work

> But... I think I would prefer to do something like `mv install-from-bindep /usr/local/bin/` and then `rm -rf /output/*`

But nothing else does it this way. There is value in consistency with other projects, and those invoke /output/install-from-bindep

https://github.com/ansible/ansible-runner/blob/d4a4b26878e43748938d12170f09b1c94c17bad0/Dockerfile#L32

The next thing I was going to look into was the `find` approach, so you may have already gotten working syntax

File: Containerfile [hunk 2]
```diff
@@ -14,7 +14,8 @@ FROM $PYTHON_BASE_IMAGE
 # =============================================================================
 
 COPY --from=python_builder /output/ /output
-RUN /output/install-from-bindep && rm -rf /output
+# building EEs require the install-from-bindep script, but not the rest of the /output folder
+RUN /output/install-from-bindep && find /output/* -not name install-from-bindep -exec rm -rf {} +
```
shanemcd:
```suggestion
RUN /output/install-from-bindep && find /output/* -not -name install-from-bindep -exec rm -rf {} +
```

beeankha (approved):

beeankha (approved):
Confirming that this makes all tests pass locally for me

AlanCoding:
Update, I have ran these steps locally

```
sudo rm -rf ~/.local/share/containers/storage
rm -rf dist/
python setup.py sdist
podman build --rm=true -t quay.io/ansible/ansible-builder:latest -f Containerfile .
py.test test/integration/test_build.py -k test_python_git_requirement[podman]
```

I repeated that, running on this branch (`gentle_removals`) and then again on `devel`.

I found the exact failure that is happening on CI when running with `devel`, and I saw it resolved with this branch.

@Spredzy's #252 _is_ blocked on merging this. However, it also appears that we're ready to merge this. I will go ahead with that now.

PR #226, opened 2021-05-26
Move requirements to bindep.txt instead of Containerfile

AlanCoding:


AlanCoding:
whoops, Paul beat me to it

PR #229, opened 2021-05-27
Make Flask test permissive to case

AlanCoding:


AlanCoding:
Looks like we do need this for tests to pass.

beeankha (approved):

Issue #228, opened 2021-05-27
Introduce a shared pattern for python test libraries


AlanCoding:
Lots of collections have pip requirements that are needed for testing but not for use. Probably _most_ of them.

Right now we maintain the exclude list:

https://github.com/ansible/ansible-builder/blob/e531bafd06cadb954c7b8aaaeb52b29179d8a127/ansible_builder/requirements.py#L7-L17

A better way to configure this would be to allow the collection to specify a separate requirements file for test requirements.

Someone might also be interested in an ansible-builder option to produce an EE that _includes_ test requirements. We would need to separate our exclude list into "always exclude" and "exclude for non-test".

Ping @thedoubl3j 

Very much related to https://github.com/ansible/ansible-builder/issues/210, and we should think through a pattern that applies equally to both these topics.

AlanCoding:
If we have a "default" location of `build-requirements.txt` top-level in a collection, then `test-requirements.txt` seems like the obvious place for test requirements. Of course, we would need a way to give a special location with the `meta/` EE metadata for collections.

Some versions of this pattern have already occurred in collections.

Issue #207, opened 2021-04-20
Allow for top-level EE denylist


pabelanger:
There are some cases, where we need to override the requirements.txt files and black file packages. There are various reasons for this.

In fact, we do this already internally for ansible-builder, however the list is hard-coded.

I propose we move this logic to an external file, and allow the execution-environment.yaml file to better control it.

AlanCoding:
> I propose we move this logic to an external file

I don't think that you're proposing that the default behavior will change. So some way or the other, we still have to ship the existing ignore list. Saying that it should be in a file, my best solution would be to package it as a data file:

https://packaging.python.org/guides/distributing-packages-using-setuptools/#data-files

Doing this exposes us to some hazards... but it's probably fine.

But we aren't using those old setuptools options, so it might look like this:

https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html

```ini
[options.data_files]
data = ansible_builder/data/ignore_python.txt
```

Talking about adding that to the `setup.cfg`.

Now, assume that we did this. It's not the main code path of `ansible-builder build` that uses this. It's `ansible-builder introspect` that would need this file. That's fine when it's using the default file. Since we have a full install in the ansible-builder image, it can figure it out without a problem. However, if the user customizes the file, then that's information outside of the container that we have to somehow get into the container. That would involve an `ADD` layer, and adding a new option to the introspect command.

All that said, I don't know if allowing this as a separate file is worth it compared to adding a list in the definition for include / exclude libraries.

This is one of those issues that definitely looks easy, but I think it has a lot of gotchas with it. We could easily mess it up.

AlanCoding:
After looking over https://github.com/ansible/ansible-builder/issues/228 I'm majorly hesitating on this idea.

We need a better handle on what categories of requirements are involved. If it basically comes down to test requirements, then I want to know if people want to build an EE _for testing_ or if they really want a test library in EEs with their collection.

The issue is a real problem, but this may be the wrong solution.

pabelanger:
bumping this up again as we are migrating to python39, from python38. Today, there is no way to denylist collections requirements. In this case, underlying layers have migrated to python39. We can pass in our own bindep.txt file with python39 references, however collections will be hard coded to python38.

The only option today, is ask a collection owner to release a new version with the updated references.

PR #230, opened 2021-05-27
Add another python test requirement to skip

AlanCoding:
After looking at https://github.com/RedHatInsights/ansible-collections-insights/pull/20, I _think_ I need to do this.

We can do it better later, filed https://github.com/ansible/ansible-builder/issues/228 for this.

### DIFF
--- a/ansible_builder/requirements.py
+++ b/ansible_builder/requirements.py
@@ -9,7 +9,7 @@ EXCLUDE_REQUIREMENTS = frozenset((
     'ansible', 'ansible-base', 'python', 'ansible-core',
     # general python test requirements
     'tox', 'pycodestyle', 'yamllint', 'pylint',
-    'flake8', 'pytest', 'pytest-xdist', 'coverage', 'mock',
+    'flake8', 'pytest', 'pytest-xdist', 'coverage', 'mock', 'testinfra',
     # test requirements highly specific to Ansible testing
     'ansible-lint', 'molecule', 'galaxy-importer', 'voluptuous',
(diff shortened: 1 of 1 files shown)


AlanCoding:
looks like we may have some test flakiness

pabelanger:
@AlanCoding what do you think of moving this logic out into a user defined denylist moving forward?  That will help greatly, and avoid the hardcoded defaults each time a collection adds something we don't like.

Also, I've proposed https://github.com/RedHatInsights/ansible-collections-insights/pull/26

thedoubl3j:
If there is a way to get to what @pabelanger that would be great. I don't anticipate many more crazy things out there that we would need to ignore as I think a lot of the `testinfra` like packages came from our automated generated exercise but if we can get stuff out in front of the user/creator, that would help a lot I think. Or we can always document somewhere something like "do not included packages needed for testing" might be able to expand this under the idea of having multiple files for requirements and then adding the one that needs to be used for EE building in the `execution-environment.yaml`. We have a few folks already doing that.

thedoubl3j (approved):
:+1:

AlanCoding:
> @AlanCoding what do you think of moving this logic out into a user defined denylist moving forward? That will help greatly, and avoid the hardcoded defaults each time a collection adds something we don't like.

That discussion is in #207. It's not trivial, the design of the schema matters. Yes, a fully feature-complete ansible-builder should have this capability, I agree with that. But there's a lot of feature gaps that we are fairly deliberate about not addressing until after 1.0.0

Issue #201, opened 2021-04-08
Feature to ignore/sanitize some bindep requirements


anshulbehl:
Currently the `sanitize_requirements` only works with python requirements and doesn't work for the binary dependencies.
While working with some partner collections we have noticed below requirements in bindep.txt
```
# This is a cross-platform list tracking distribution packages needed by this collection;
# see https://docs.openstack.org/infra/bindep/ for additional information.

python >=3.6,<=3.8
```
This causes failure while building as below with ubi8 images

```
There are following alternatives for "python": python2, python36, python38
Error: Unable to find a match: python
STEP 12: FROM registry-proxy.engineering.redhat.com/rh-osbs/ansible-automation-platform-ee-minimal:ansible-automation-platform-2.0-rhel-8-containers-candidate-50443-20210329152500
Error: error building at STEP "RUN assemble": error while running runtime: exit status 1
```
python as a requirement should be ignored in the bindep file[python38 is already installed in the base images]  and for that we would probably need to create a feature to sanitise binary dependencies.

AlanCoding:
In cases we are relying on things like `python38-devel`

https://github.com/ansible/awx-ee/blob/devel/_build/bindep_combined.txt#L10

We might leave those untouched, I'm not sure.

anshulbehl:
for this one I think we only need to sanitise when we see `python` in bindep file

AlanCoding:
Could you link to a case of this? This shouldn't be too hard to implement but I want to know a little better what we're targeting.

Issue #90, opened 2020-08-21
Malformed bindep.txt files do not halt execution


AlanCoding:
See https://github.com/ansible/ansible-builder/pull/89

We don't know how to determine of the `bindep` CLI command is successful in finding the missing deps, or if it has erred.

AlanCoding:
Issue was filed upstream for `bindep` here https://storyboard.openstack.org/#!/story/2008049

Ping @pabelanger, this can create situations which are _really_ difficult to debug. I mainly hit it where it would error without a trailing newline at the end of the input bindep.txt file (but there are many other things that can go wrong, from collection bindep.txt files). It keeps running deep into the build process because there's no indicator to alert the code to the error (meaning of return code is overloaded for other purpose), and by the time it does error it is a nightmare to figure out. This is a big part of the reason we write the intermediary files like `bindep_combined.txt` - because this file might contain an error dump from the `bindep` CLI, and there's no way for the user to know. I would say that `assemble` becomes obligated to handle these errors, but I honestly don't think it can because this is a pretty fundamental problem.

AlanCoding:
I'm increasingly thinking that the best solution is to make `ansible-builder` depend on bindep as a python package and call it directly in code.

```python
    if depends is None:
        depends = bindep.depends.get_depends(args.filename)
        if not depends:
            return 1

    if args.profiles:
        logging.info("Platform profiles:")
        for profile in depends.platform_profiles():
            logging.info("%s", profile)
        logging.info("")
        logging.info("Configuration profiles:")
        for profile in depends.profiles():
            logging.info("%s", profile)
        return 0
```

The cases which are actually painful for us look like fairly intentional return 1 or 0 cases in code, and we could error handle just fine if we were doing this in python, IMO.

But now we're calling this via the `assemble` script, and this leads back to my idea that we may be better off to combine those with introspect to create multiple documented tools with a CLI to go with them.

AlanCoding:
I am likely to combine this with https://github.com/ansible/ansible-builder/issues/201

felixfontein:
Hmm, I think I also encountered that: https://github.com/ansible-collections/community.sops/runs/6039420137?check_suite_focus=true

I had URLs in my bindep.txt file, which made bindep barf, but the build process just continued and ended with `Complete!`. (Obviously the dependency wasn't in the EE image and thus the tests expecting it to be there failed...)

Shrews:
[...]

PR #234, opened 2021-05-28
Sanitize bindep content that we get

AlanCoding:
Connect https://github.com/ansible/ansible-builder/issues/201

Connect https://github.com/ansible/ansible-builder/issues/90

This should resolve both of those when complete. Right now it still needs a little more testing via an integration test. The failure cases won't come up until some time into the build process.

This also has more risk than most other things for collections concerned with execution environments. So I would like to see about gathering a list of relevant collections and then running the introspect script to assure I'm getting sane content.

### DIFF
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -11,7 +11,7 @@ from .colors import MessageColors
 from .exceptions import DefinitionError
 from .main import AnsibleBuilder
 from .introspect import process, simple_combine, base_collections_path
-from .requirements import sanitize_requirements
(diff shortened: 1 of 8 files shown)


Akasurde:
@AlanCoding Do you want to rebase this branch? Thanks

AlanCoding:
oof, I don't feel like I have the time to take that on right now, but thanks for the ping. I will close this for now. Love the work you've been doing, and I want to advocate for builder, but my development focus is elsewhere at the moment.

Issue #212, opened 2021-04-30
Add ability to just generate contents of build folder


pabelanger:
It would be helpful to allow of just running the generate build folder command, rather then a full build.

We have a use case downstream, where we want to generate the latest _build folder contents, but build using a different container build system.  Today we run ansible-builder, but ctrl-C when the 2nd stage builds start.

AlanCoding:
Early in this project's history we have 2 primary actions, `ansible-builder create` and `ansible-builder build`. The "create" command would just create the build context without running the podman build command.

https://github.com/ansible/ansible-builder/blob/1c7b5d317399e18596e04500d4b68b2add91b6ad/ansible_builder/cli.py#L49-L56

This was no longer possible after we started running the "introspect" script outside of the container - which we are removing now.

So does the re-introduction of the create command match what you're saying here @pabelanger? That would be a well-defined task.

shanemcd:
This is now possible with #217

AlanCoding:
I'm going to say that this is ready for devel as the old `ansible-builder create` command.

Akasurde:
PR #236 and #250 merged. Closing this issue. Please feel free to re-open if the problem persists. Thanks.

PR #236, opened 2021-06-01
Add Back 'create' Command as an Option

beeankha:
Addressing issue https://github.com/ansible/ansible-builder/issues/212

Instead of requiring users to run a full build, this change will enable the creation of just a `Containerfile`.

### DIFF
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -22,7 +22,19 @@ def run():
     args = parse_args()
     configure_logger(args.verbosity)
 
-    if args.action in ['build']:
+    if args.action in ['create']:
+        logger.debug(f'Ansible Builder is creating a Containerfile for your execution environment image, "{args.tag}".')
+        ab = AnsibleBuilder(**vars(args))
+        action = getattr(ab, ab.action)
+        try:
+            if action():
+                print("Complete! Build context is at: {}".format(ab.build_context))
+                sys.exit(0)
+        except DefinitionError as e:
+            logger.error(e.args[0])
+            sys.exit(1)
+
+    elif args.action in ['build']:
         logger.debug(f'Ansible Builder is building your execution environment image, "{args.tag}".')
         ab = AnsibleBuilder(**vars(args))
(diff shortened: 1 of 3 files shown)


ansible_builder/cli.py
```diff
@@ -22,8 +22,11 @@ def run():
     args = parse_args()
     configure_logger(args.verbosity)
 
-    if args.action in ['build']:
-        logger.debug(f'Ansible Builder is building your execution environment image, "{args.tag}".')
+    if args.action in ['create', 'build']:
+        if args.action == 'build':
+            logger.debug(f'Ansible Builder is building your execution environment image, "{args.tag}".')
```
AlanCoding: I would suggest moving these log statements into the `ansible_builder.main` module inside of the build / create method. This avoids using an `if` and would allow you to order them in a way that makes more sense. We should emit _this_ log until right before running the `podman build` command.

ansible_builder/cli.py
```diff
@@ -98,7 +111,7 @@ def parse_args(args=sys.argv[1:]):
                                       default=constants.default_tag,
                                       help='The name for the container image being built (default: %(default)s)')
 
-    for p in [build_command_parser]:
+    for p in [create_command_parser, build_command_parser]:
```
AlanCoding, line 123: Some time has passed since this look was last functional, and not all of these options look like they should apply to the `create` command.

 - `--container-runtime` would seem to have no effect on `ansible-builder create`, so it would be best to exclusively add it to the build_command_parser
 - `--build-arg` also only gets used on `podman build`, so it's irrelevant for create, so same goes for this

I think the rest of them look right.

beeankha: Done, now the `--help` for `anisble-builder build` shows the following options:

Screenshot from 2021-06-03 11-50-08


and for `create` it shows:

Screenshot from 2021-06-03 11-49-32

AlanCoding (approved):
I am super excited about this. Thanks for getting it knocked out so fast.

Issue #213, opened 2021-05-10
When layering a collection on top of an EE builder introspects the lower layers too


anshulbehl:
When you try to build and EE with a partner collection on top of an already built EE that has its own collection, builder tries to introspect all the collections inside all the layers.
This essentially means that if there are EE/s that are created without using builder or using manual build requirements and binary dependencies, it can probably fail to find some custom requirements.

Usecase1: I have created an EE that has an installed requirement not present on pypi and that requirements is part of the collection requirements, if a consumer tries to layer their collection on my collection the builder process will fail

Usecase2: I have created an EE that has an installed bindep that is coming from a closed source, when a consumer tries to layer their collection on top of my collection, it will fail if builder is not attached to the closed source at the time of building new layer.

Ideally, Lower layers should not be introspected when builder is trying to build an EE. Specific issue is related to trying to layer partner collection on top of RH supported EE that has the rhv/ovirt collection built in.

Let me know if specific logs are needed or more info is needed as well

AlanCoding:
We have never seriously looked into the engineering complexities of layering EEs on top of each other. This would be a major item of work that needs real planning.

PR #238, opened 2021-06-02
Pull request for devel

pabelanger:
Rename EE_BASE_IMAGE to ANSIBLE_RUNNER_IMAGE
Rename EE_BUILDER_IMAGE to ANSIBLE_BUILDER_IMAGE

### DIFF
--- a/ansible_builder/constants.py
+++ b/ansible_builder/constants.py
@@ -14,7 +14,7 @@ base_collections_path = '/usr/share/ansible/collections'
 
 build_arg_defaults = dict(
     ANSIBLE_GALAXY_CLI_COLLECTION_OPTS='',
-    EE_BASE_IMAGE='quay.io/ansible/ansible-runner:devel',
+    ANSIBLE_RUNNER_IMAGE='quay.io/ansible/ansible-runner:devel',
     ANSIBLE_BUILDER_IMAGE='quay.io/ansible/ansible-builder:latest'
 )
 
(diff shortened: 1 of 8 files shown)


ansible_builder/constants.py
```diff
@@ -14,8 +14,8 @@
 
 build_arg_defaults = dict(
     ANSIBLE_GALAXY_CLI_COLLECTION_OPTS='',
-    EE_BASE_IMAGE='quay.io/ansible/ansible-runner:devel',
-    EE_BUILDER_IMAGE='quay.io/ansible/ansible-builder:latest'
+    ANSIBLE_RUNNER_IMAGE='quay.io/ansible/ansible-runner:devel',
```
AlanCoding, line 17: Raising this point out of a sense of obligation:

https://github.com/ansible/ansible-builder/issues/213

If we, in the future, allow building EEs with other EEs as the base image (as opposed to the ansible-runner image), then this naming will age poorly.

How interested are we in that feature? I'm not sure. If we don't intend to support it, then your naming here would help keep people from trying it.

AlanCoding (approved):
If you merge this, would you be willing to close #213? If yes, then merge.

pabelanger:
@AlanCoding I'm not sure, the plan for downstream was to start to build things atop of an EE, to reduce the duplication of dependency with multiple builds over layered approach.

Given I didn't raise it, will have to differ to @anshulbehl

pabelanger:
that said, we really need to understand if we want to change this to redhat product specific naming or more upstream specific images.  Either choice is going to cause confusion for people.

PR #240, opened 2021-06-02
Add podman to tox allowlist_externals for integration tests

beeankha:
When running `tox`, the following warning would pop up:
```
WARNING: test command found but not installed in testenv
  cmd: /usr/bin/podman
  env: /home/bhenderson/Documents/GitHub/ansible-builder/.tox/integration
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above!
```
Adding `podman` to `allowlist_externals` in order to make this warning no longer appear.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -29,9 +29,10 @@ passenv =
   KEEP_IMAGES
 deps =
     pytest-xdist
-whitelist_externals =
+allowlist_externals =
     bash
     mkdir
+    podman
 commands =
     mkdir -p artifacts
     python setup.py sdist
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -29,9 +29,10 @@ passenv =
   KEEP_IMAGES
 deps =
     pytest-xdist
-whitelist_externals =
+allowlist_externals =
     bash
     mkdir
+    podman
```
AlanCoding, line 36: what about docker?

AlanCoding (approved):

PR #263, opened 2021-07-28
Add docs tox target and Zuul job

Shrews:
This adds a documentation build tox target that will build HTML formatted docs locally. E.g.,

`tox -e docs`

All doc build warnings are treated as errors. Because of this, a few warnings are fixed in the current docs.

This also adds a Zuul check and gate job to test documentation building.

### DIFF
--- a/.zuul.d/project.yaml
+++ b/.zuul.d/project.yaml
@@ -3,6 +3,9 @@
     check:
       jobs:
         # - ansible-builder-tox-integration
+        - ansible-tox-docs:
+            vars:
+                sphinx_build_dir: docs/build
         - ansible-builder-build-container-image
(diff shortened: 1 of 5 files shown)


.zuul.d/project.yaml
```diff
@@ -3,6 +3,7 @@
     check:
       jobs:
         # - ansible-builder-tox-integration
+        - ansible-tox-docs
```
pabelanger: you should add this to gate below also.

Shrews: yup, just wanted it working in check first. done

Shrews:
recheck

shanemcd (approved):

PR #274, opened 2021-08-24
Add py38 job

Shrews:
In preparation for removing the py36 and py37 jobs from `ansible/project-config`, add a py38 job in-repo. This is necessary because the py37 job occasionally chooses an AWS Fedora34 image that does not have python 3.7 installed, and will fail.

### DIFF
--- a/.zuul.d/project.yaml
+++ b/.zuul.d/project.yaml
@@ -10,6 +10,7 @@
         - github-workflows:
             files:
               - .github/workflows/.*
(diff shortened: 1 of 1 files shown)


shanemcd (approved):

Shrews:
recheck

pabelanger:
@Shrews @shanemcd lets drop ansible/check as required branch protection. This way, 'recheck' or 'gate' labels will directly enqueue into gate now.

PR #276, opened 2021-08-25
Do not merge: possible factory pattern

Shrews:
Just an example


samdoran:
I finally took the time to read through this and it's very helpful. I haven't worked with ABCs myself so the example is helpful. I am going to try for a factory method in the my implementation rather than using a simple factory function. We'll see how it goes.🤞

PR #284, opened 2021-09-15
Run tests using GitHub Actions

samdoran:
Add tests that run in parallel using a matrix. Each test run corresponds to a `tox` environment. A custom container image is used for all tests except integration tests to improve test reliability and prevent having to install Python and `tox` for every test.

I haven't figure out how to run `podman` in a  Docker container so the integration tests run directly in the VM.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,10 +72,13 @@ jobs:
         with:
           python-version: ${{ matrix.py_version.name }}
 
-      - name: Setup test environment
+      - name: Install tox
         run: |
-          ${{ format('python{0}', matrix.py_version) }} -m pip install --upgrade pip
-          ${{ format('python{0}', matrix.py_version) }} -m pip install tox
+          python3 -m pip install --upgrade pip
+          python3 -m pip install tox
+
+      - name: Create tox environment
+        run: |
           tox --notest
 
       - name: Run integration tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,17 +2,7 @@ name: CI
 
 on:
   pull_request:
-    branches:
-      - devel
-      - release_*
-
   push:
-    branches:
-      - devel
-      - release_*
-
-  schedule:
-    - cron: '00 04 * * *'
 
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,14 @@ jobs:
       - name: Run tests
         run: tox -e ${{ matrix.test.env }}
 
+      - name: Save docs
+        uses: actions/upload-artifact@v2
+        if: ${{ success() && matrix.test.name == 'Docs' }}
+        with:
+          name: docs
+          path: docs/build/html
+          retention-days: 3
+
   unit:
     name: ${{ matrix.test.name }} - ${{ matrix.py_version.name}}
     runs-on: ubuntu-20.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,26 @@ on:
 jobs:
   integration:
     runs-on: ubuntu-20.04
-    name: Integration
-    steps:
-      - uses: actions/checkout@v2
+    name: Integration - ${{ matrix.py_version.name }}
 
-      - uses: actions/setup-python@v2
+    strategy:
+      fail-fast: false
+      matrix:
+        py_version:
+          - name: 3.8
+            tox: py38
+
+          - name: 3.9
+            tox: py39
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install Python ${{ matrix.py_version.name }}
+        uses: actions/setup-python@v2
         with:
-          python-version: "3.8"
+          python-version: ${{ matrix.py_version.name }}
 
       - name: Install dependencies
         run: |
@@ -27,7 +40,7 @@ jobs:
 
       - name: Run integration tests
         run: |
-          tox -e integration
+          tox -e integration-${{ matrix.py_version.tox }}
 
   sanity:
     name: ${{ matrix.test.name }}
@@ -96,3 +109,21 @@ jobs:
 
       - name: Run tests
         run: tox -e ${{ matrix.test.env }}-${{ matrix.py_version.tox}}
+
+  images:
+    name: Container Build - ${{ matrix.runtime }}
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runtime:
+          - docker
+          - podman
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build Container Image
+        run: ${{ matrix.runtime }} build --rm=true -t quay.io/ansible/ansible-builder -f Containerfile .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,14 +41,6 @@ jobs:
       - name: Run tests
         run: tox -e ${{ matrix.test.env }}
 
-      - name: Save docs
-        uses: actions/upload-artifact@v2
-        if: ${{ success() && matrix.test.name == 'Docs' }}
-        with:
-          name: docs
-          path: docs/build/html
-          retention-days: 3
-
   integration:
     runs-on: ubuntu-20.04
     name: Integration - ${{ matrix.py_version.name }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,10 +4,12 @@ on:
   pull_request:
     branches:
       - devel
+      - release_*
 
   push:
     branches:
       - devel
+      - release_*
 
   schedule:
     - cron: '00 04 * * *'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,25 @@ jobs:
         run: tox
 
 
+  images:
+    name: Build Image - ${{ matrix.runtime }}
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runtime:
+          - docker
+          - podman
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build Container Image
+        run: ${{ matrix.runtime }} build --rm=true -t quay.io/ansible/ansible-builder -f Containerfile .
+
+
   integration:
     runs-on: ubuntu-20.04
     name: Integration - ${{ matrix.py_version.name }}
@@ -106,22 +125,3 @@ jobs:
 
       - name: Run tests
         run: tox
-
-
-  images:
-    name: Build Image - ${{ matrix.runtime }}
-    runs-on: ubuntu-20.04
-
-    strategy:
-      fail-fast: false
-      matrix:
-        runtime:
-          - docker
-          - podman
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Build Container Image
-        run: ${{ matrix.runtime }} build --rm=true -t quay.io/ansible/ansible-builder -f Containerfile .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ on:
   schedule:
     - cron: '00 04 * * *'
 
+
 jobs:
   sanity:
     name: ${{ matrix.test.name }}
@@ -20,41 +21,47 @@ jobs:
       image: quay.io/ansible/ansible-builder-test-container:1.0.0
       env:
         PIP_CACHE_DIR: ${{ runner.temp }}/.cache/pip
+        TOXENV: ${{ matrix.test.tox_env }}
 
     strategy:
       fail-fast: false
       matrix:
         test:
           - name: Lint
-            env: linters
+            tox_env: linters
 
           - name: Docs
-            env: docs
+            tox_env: docs
 
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Setup test environment
-        run: tox -e ${{ matrix.test.env }} --notest
+      - name: Create tox environment
+        run: tox --notest
 
       - name: Run tests
-        run: tox -e ${{ matrix.test.env }}
+        run: tox
+
 
   integration:
     runs-on: ubuntu-20.04
     name: Integration - ${{ matrix.py_version.name }}
     needs: images
 
+    env:
+      TOXENV: ${{ matrix.py_version.tox_env }}
+
     strategy:
       fail-fast: false
       matrix:
         py_version:
           - name: 3.8
-            tox: py38
+            tox_env: py38
 
           - name: 3.9
-            tox: py39
+            tox_env: py39
+
 
     steps:
       - name: Checkout
@@ -67,13 +74,14 @@ jobs:
 
       - name: Setup test environment
         run: |
-          ${{ format('python{0}', matrix.py_version.name) }} -m pip install --upgrade pip
-          ${{ format('python{0}', matrix.py_version.name) }} -m pip install tox
-          tox -e integration-${{ matrix.py_version.tox }} --notest
+          ${{ format('python{0}', matrix.py_version) }} -m pip install --upgrade pip
+          ${{ format('python{0}', matrix.py_version) }} -m pip install tox
+          tox --notest
 
       - name: Run integration tests
         run: |
-          tox -e integration-${{ matrix.py_version.tox }}
+          tox
+
 
   unit:
     name: ${{ matrix.test.name }} - ${{ matrix.py_version.name}}
@@ -82,6 +90,7 @@ jobs:
       image: quay.io/ansible/ansible-builder-test-container:1.0.0
       env:
         PIP_CACHE_DIR: ${{ runner.temp }}/.cache/pip
+        TOXENV: ${{ matrix.py_version.tox_env }}
 
     strategy:
       fail-fast: false
@@ -92,20 +101,21 @@ jobs:
 
         py_version:
           - name: 3.8
-            tox: py38
+            tox_env: py38
 
           - name: 3.9
-            tox: py39
+            tox_env: py39
 
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Setup test environment
-        run: tox -e ${{ matrix.test.env }}-${{ matrix.py_version.tox}} --notest
+      - name: Create tox environment
+        run: tox --notest
 
       - name: Run tests
-        run: tox -e ${{ matrix.test.env }}-${{ matrix.py_version.tox}}
+        run: tox
+
 
   images:
     name: Build Image - ${{ matrix.runtime }}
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -1,34 +1,32 @@
...
+description = Run all tests with {basepython}
 usedevelop = True
-setenv =
-  VIRTUAL_ENV={envdir}
 deps =
-    pytest
-commands =
-    py.test -v test -k "not integration" {posargs}
+    -r {toxinidir}/test/requirements.txt
+commands = pytest {posargs:test}
```
webknjaz: Instead of setting "test" here, it's usually better to set it in `test_paths =` in `pytest.ini`

samdoran: Does that let you override it with positional args?

webknjaz: Yep

Shrews (commented):
Is there any advantage to having the docs job in GHA rather than Zuul? Losing the intermediate doc artifact to view the doc results is really a big hit here, and I'm not really seeing the advantage of moving that job. Is it a lot faster or something?

webknjaz:
Personally, I'd recommend enabling the PR integration in RTD. It integrates nicely with the PRs meaning that you get preview link in the checks every time and you don't need to go anywhere to find it.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,131 @@
...
+
+          - name: Docs
+            env: docs
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup test environment
+        run: tox -e ${{ matrix.test.env }} --notest
```
webknjaz: Instead of `-e` in each place, consider just adding
```yaml
env:
    TOXENV: ${{ matrix.env }}
```
to this job. It's a widely used practice across the ecosystem.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,131 @@
...
+
+      - name: Run tests
+        run: tox -e ${{ matrix.test.env }}
+
+      - name: Save docs
```
webknjaz: You probably don't need this if you just enable RTD. It'll have PR previews out of the box, with no expiration.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,131 @@
...
+        if: ${{ success() && matrix.test.name == 'Docs' }}
+        with:
+          name: docs
+          path: docs/build/html
+          retention-days: 3
+
   integration:
     runs-on: ubuntu-20.04
-    name: Integration
+    name: Integration - ${{ matrix.py_version.name }}
```
webknjaz, line 60: Note that the GH UI is quite narrow in some places. This is why verify that this fits and try to go for short job names.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,131 @@
...
-          python-version: "3.8"
+          python-version: ${{ matrix.py_version.name }}
 
-      - name: Install dependencies
+      - name: Setup test environment
         run: |
-          python -m pip install --upgrade pip
-          pip install tox
+          ${{ format('python{0}', matrix.py_version.name) }} -m pip install --upgrade pip
+          ${{ format('python{0}', matrix.py_version.name) }} -m pip install tox
```
webknjaz: I wouldn't put this many separate commands in one step because it's hard to parse where one starts and the other one ends in the logs. Unless you add runner commands for collapsing the log groups around each.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,131 @@
...
+          tox -e integration-${{ matrix.py_version.tox }} --notest
```
webknjaz: I don't understand why you'd need to explicitly pass the py version here. Can't you just set up the default python and use `TOXENV=integration-py` to use it?

test/unit/test_utils.py
```diff
@@ -0,0 +1,104 @@
...
+    assert write_file(path, text)  # does not exist, write
+    assert not write_file(path, text)  # already correct, do not write
+
+
+@pytest.fixture
+def source_file(tmpdir):
+    source = os.path.join(tmpdir, 'bar.txt')
+    with open(source, 'w') as f:
+        f.write('foo\nbar\n')
+    return source
```
webknjaz: Why use tmp_dir?

```suggestion
def source_file(tmp_path):
    source = tmp_path / 'bar.txt'
    source.write_text('foo\nbar\n')
    return str(source)
```

.github/workflows/ci.yml
samdoran: I was trying to duplicate the functionality in Zuul that previews docs builds for PRs. I didn't know that was something that Read the Docs could do.

.github/workflows/ci.yml
samdoran: That's a good idea.

.github/workflows/ci.yml
samdoran: Currently this is run in a VM, so I probably don't need to be explicit about the Python executable since the only Python version will be the one installed by the `setup-python@v2` action. I was mainly thinking about when I get these tests moved to running within the test container, which has multiple python versions installed. In that case `pip` is ambiguous.

But, inside the container image, `tox` is already installed, so I would be able to delete the `pip` commands.

In general, I prefer to be explicit about the Python executable that's invoking `pip` rather than just calling `pip` directly.

test/unit/test_utils.py
samdoran: I'm not sure I fully understand the question. Why use a temporary directory at all here? Are you saying this should just be a permanent test fixture rather than creating the same file each time? Or should I return a `Path` object if I'm going to bother changing this to `tmp_path`?

I didn't look at the broader use of the fixture and how it's used it tests, so there is most likely more room for improvement.

.github/workflows/ci.yml
webknjaz: Here's an example of how it looks in the checks section: https://github.com/ansible/pylibssh/pull/250.

test/unit/test_utils.py
webknjaz: This is mostly about the `tmpdir` fixture being deprecated while `tmp_path` is the thing that replaces it and should be always used instead.

.github/workflows/ci.yml
webknjaz: What I observe across the ecosystem is that there's a `setup-python` step. Then, there's some `python -m pip install tox --user` (that uses the previously set up python interpreter). And then, there's `python -m tox --notest --skip-missing-interpreters false` (for provisioning) and `python -m tox --skip-missing-interpreters false` for running the actual testing.

And then, there would be a matrix of envs/pythons/OSs that makes up jobs that GHA would know better how to orchestrate.

webknjaz: Why do you want to put this in a container, anyway?

.github/workflows/ci.yml
```diff
@@ -1,30 +1,139 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        test:
+          - name: Lint
+            tox_env: linters
+
+          - name: Docs
+            tox_env: docs
```
webknjaz, line 26: looking at this: do you really need separate name and tox env? It's usually more than enough to reuse envs as job names:
```suggestion
        tox_env:
        - linters
        - docs
```

.github/workflows/ci.yml
```diff
@@ -1,30 +1,139 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        py_version:
+          - name: 3.8
+            tox_env: py38
+
+          - name: 3.9
+            tox_env: py39
```
webknjaz: Once you select the current Python version, just using a static string `python` or `py` for `$TOXENV` is enough to get it picked up
```suggestion
      TOXENV: python

    strategy:
      fail-fast: false
      matrix:
        py_version:
        - 3.8
        - 3.9
```

.github/workflows/ci.yml
```diff
@@ -1,30 +1,139 @@
...
+      env:
+        PIP_CACHE_DIR: ${{ runner.temp }}/.cache/pip
+        TOXENV: ${{ matrix.py_version.tox_env }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        test:
+          - name: Unit
+            env: unit
```
webknjaz: this is unused

test/unit/test_utils.py
samdoran: Oh man, I had a total Monday moment. Sorry. I didn't realize that was a code suggestion. I thought it was a diff from my PR. 🤦‍♂️ 

I have switched to using `tmp_path` in _so many_ places I thought that was a change I had made here as well.

.github/workflows/ci.yml
samdoran: I could do that, I just like the way it looks having the job name capitalized. GHA doesn't provide an easy mechanism for manipulating the tox environment string into a name that I think looks nice, so that's the reason for separate entries. It's purely cosmetic.

.github/workflows/ci.yml
samdoran: > Once you select the current Python version

How do you "select" the Python version? By only installing one Python version with the `setup-python` action?

It seems like using `python` is ambiguous in an environment where multiple Python versions are installed, which is the case in the test image. Currently the integration tests run in the VM and I'm installing Python using the `setup-python` action. I would like to move away from that in the future once I figure out how to run Podman inside a Docker container.

In the test container image, all the Python versions are already installed as well as `tox`. I would think setting `TOXENV=python` in that case would pick up whichever Python version is sym linked to `/user/bin/python`, which may or may not be the version I want to test, unless there is some other tox magic I don't know about.

.github/workflows/ci.yml
samdoran: It seems it was used to create these entries in the test matrix:
 - [Unit - 3.8](https://github.com/ansible/ansible-builder/pull/284/checks?check_run_id=3793391339)  
 -  [Unit - 3.9](https://github.com/ansible/ansible-builder/pull/284/checks?check_run_id=3793391405)

test/unit/test_utils.py
samdoran: These fixtures are all original code I just moved. That's why they are showing up as "new" changes in this PR.

Ok, now I think I've un-confused myself. 😄

.github/workflows/ci.yml
```diff
@@ -1,30 +1,135 @@
----
-
 name: CI
 
 on:
   pull_request:
-    branches: [devel]
+    branches:
+      - devel
```
Shrews: I think if you just leave off the `branches` part, it will trigger for a pull request on _any_ branch, which would be what we want for backports to be tested properly.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,135 @@
...
 
   push:
-    branches: [devel]
+    branches:
+      - devel
+
+  schedule:
+    - cron: '00 04 * * *'
```
Shrews: Not sure I see the need for scheduled CI runs. Is there a benefit I'm not seeing?

.github/workflows/ci.yml
samdoran: This evaluates the base branch, not the source branch name. Limiting this workflow to only devel and release branches would prevent it  from running against other branches that get pushed to the repo. There shouldn't be many (any?) of those, though.

From the [GHA docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags):
> For a pull_request event, only branches and tags on the base are evaluated.

.github/workflows/ci.yml
samdoran: There's no real _need_ for it. I just figured it would be good to have at least some regular CI runs happening. Maybe weekly instead of nightly?

Shrews: My reading of the GHA docs seems to indicate that the person that added this would get notified about the run status. If that's correct, then you probably shouldn't be the one adding this. I could add it later.

.github/workflows/ci.yml
webknjaz: Yep, Sam is correct: the `pull_request` event happens whenever there's an interaction with a PR. If there's a branch pushed to the upstream repo, it'll generate a `push` event, and only after a PR gets created there's *additionally/separately* a `pull_request` happening. If somebody would want to send a PR against a non-matching branch, this would not cause GHA to react to this specific event if the branch is not included.
Although, it's hard to imagine this happening often (or ever) so I don't see a strong argument for maintaining the list of branches for this event.

.github/workflows/ci.yml
webknjaz: @Shrews it's usually used for nightlies, as Sam mentioned. There's one quirk, though. In the event a repo stops being active, GitHub disables the workflows with "cron". An unobvious side-effect is that it does disable the whole workflow, not just the `schedule` trigger. So if there's other event triggers in the same workflow, this will result in new jobs not being started if somebody sends in a contribution after the disabling happens.

webknjaz: > My reading of the GHA docs seems to indicate that the person that added this would get notified about the run status. If that's correct, then you probably shouldn't be the one adding this. I could add it later.

IIRC that is true, the person who pushed it gets notified (or is it the author of the last commit on the branch?)

samdoran: That is very interesting and true. I'll go ahead and remove it for now.

.github/workflows/ci.yml
webknjaz: FWIW I'm working on a reusable workflow for tox and here's what it'll look like: https://github.com/webknjaz/tox-gha-test/actions/runs/1450588254 — with the exact env names.

.github/workflows/ci.yml
samdoran: I'll remove the branch list.

.github/workflows/ci.yml
webknjaz: `tox -e py` or `tox -e python` will select whatever currently is run if you just type in `python`. In GHA that would be whatever the last step running `setup-python` activated.

This is the approach you'd find in any (most?) of the big community projects (like setuptools, or tox, or pytest). It's been the de-facto approach for as long as I can remember. Definitely since the times Travis was popular, which then migrated to Azure and GHA.

Another approach you may see in such CI setups is running `tox --notest` first, for venv provisioning and then, running `tox` in a subsequent job. The reason is that it's nicer when you see separate steps in the CI output, it's more obvious which one failed (there was a bug in tox in some version that made people think that running tests failed but it was actually the pre-test stage, this was the case where the distraction was also useful). In some CIs, specifically Travis, there's different "stages". and depending on which stage errors out, it'd show different status for that job — error for the installation stage and failure for actual testing (or was it vice versa? I don't remember exactly) — this was another DX thingy that may be important.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,127 @@
...
-
 name: CI
 
 on:
   pull_request:
-    branches: [devel]
-
   push:
-    branches: [devel]
+
```
webknjaz, line 6: Did you intentionally add this new line?

.github/workflows/ci.yml
```diff
@@ -1,30 +1,127 @@
----
```
webknjaz, line 1: Any reason you didn't like this?

.github/workflows/ci.yml
```diff
@@ -1,30 +1,127 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        py_version:
+          - name: 3.8
+            tox_env: unit-py38
+
+          - name: 3.9
+            tox_env: unit-py39
```
webknjaz, line 117: have you considered just using
```suggestion
            tox_env: unit

          - name: 3.9
            tox_env: unit
```

.github/workflows/ci.yml
samdoran: I think the document start marker should only exist if it's being used or if there are directives above the marker. Since this is a single document with no directives, it's not adding any value.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,127 @@
...
+        uses: actions/checkout@v2
+
+      - name: Create tox environment
+        run: tox --notest
+
+      - name: Run tests
+        run: tox
+
+
+  images:
```
webknjaz: Since this job is needed by others, maybe put it above them so it's logically ordered?

.github/workflows/ci.yml
samdoran: Yes, I have two blank lines between sections in the document so I added another here to match.

.github/workflows/ci.yml
```diff
@@ -1,30 +1,127 @@
...
+
+          - name: Docs
+            tox_env: docs
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Create tox environment
+        run: tox --notest
```
webknjaz, line 33: A common theme you may notice across the popular project CI setups is that they use `--skip-missing-interpreters false` in every tox invocation, as a sanity check.
```suggestion
        run: tox --skip-missing-interpreters false --notest
```

.github/workflows/ci.yml
samdoran: But that may not use the correct Python version.

.github/workflows/ci.yml
samdoran: That makes sense.

.github/workflows/ci.yml
webknjaz: Ah, right, you don't use the setup action. The typical way is to invoke `setup-python` and then tox. Then it'll choose the right one.

samdoran: The container image where the test is running has multiple Python versions installed.

webknjaz: Oh, I missed that it's a container. Makes sense, then.

webknjaz: I couldn't see why it diverged from what I typically used to be seeing.

tox.ini
```diff
@@ -1,39 +1,33 @@
...
-commands=
+description = Run code linters
+commands =
     flake8 --version
     flake8 ansible_builder test
     yamllint --version
     yamllint -s .
 
-[testenv:integration]
+[testenv:unit{,-py38, -py39}]
```
webknjaz: Is the whitespace needed here?

samdoran: I don't think so.

Shrews (approved):

test/integration/test_build.py
```diff
@@ -148,6 +148,7 @@ def test_has_pytz(self, cli, container_runtime, pytz):
         r = cli(f'{container_runtime} run --rm {ee_tag} pip3 show pytz')
         assert 'World timezone definitions, modern and historical' in r.stdout, r.stdout
 
+    @pytest.mark.skip("Concurrency issues plus output parsing is using the incorrect format")
```
webknjaz, line 151: @samdoran for future, it's better to use xfail per https://pganssle-talks.github.io/xfail-lightning/ + https://pypi.org/p/pytest-rerunfailures exists.

samdoran: Ah, that's good to know. This test is intermittent plus poorly written. I didn't know about `pytest-rerunfailures`.

webknjaz: If it's flaky, use `xfail` + `strict=False` unless you want retries.

PR #302, opened 2021-10-29
Set the version in the docs conf

samdoran:
In order to get the version, the package needs to be installed since the version
string only exists after the package is installed.

Minor change to author and copyright info


shanemcd (approved):

samdoran:
recheck

samdoran:
recheck

PR #330, opened 2021-12-16
Mark unreliable tests

samdoran:
These tests fail on podman sometimes due to dnf issues, sometimes due to various errors from podman.


Shrews (changes requested):
Let's just skip them if `runtime == 'podman'` rather than disable them completely.

Shrews (approved):
Oh, I misunderstood what `xfail` does. Nevermind last comment.

Issue #323, opened 2021-12-10
Option to disable --sanitize / override collection dependencies


jplitza:
We have the situation that a collection we use depends on a module `foo` (without version). We would like to patch that version, and included a line in our `requirements.txt` like
```
foo @ git+https://github.com/...@deadbeefdeadbeefdeadbeefdeadbeef
```
But since `ansible-builder introspect` has a hardcoded `--sanitize` argument, our VCS specification is completely ignored and merged to only `foo` with the collection requirements.

SirUli:
So that is about a python requirements.txt which has special settings which are dropped.

The problem here is likely related to #319 since the library that performs this parsing is more than 3 years unmaintained and doesn't understand many of the pip features.

jplitza:
I saw that issue but doubt it's related. It may be a dependency to fix this  issue, but the builder code explicitly acts upon the requirement's name only:
https://github.com/ansible/ansible-builder/blob/1ad38e40f5a73834625073117852c7735dad016a/ansible_builder/requirements.py#L31-L36

SirUli:
Yep, tested that now. It is partly due to that. While the logic in the ansible-builder code only compares names and appends specs then, the library requirements-parser also doesn't yield the required feedback:

```python
lines = ['foo @ git+https://github.com/@deadbeefdeadbeefdeadbeefdeadbeef']
for req in requirements.parse('\n'.join(lines)):
    print(req.line)
    print(req.editable)
    print(req.local_file)
    print(req.specifier)
    print(req.vcs)
    print(req.revision)
    print(req.name)
    print(req.uri)
    print(req.subdirectory)
    print(req.path)
    print(req.hash_name)
    print(req.extras)
    print(req.specs)
```

Leads to:

```
foo @ git+https://github.com/@deadbeefdeadbeefdeadbeefdeadbeef
False
False
True
None
None
foo
None
None
None
None
[]
[]
```

The library simply cannot deal with this format. It however would understand the format as follows:

``git+https://git.myproject.org/SomeProject#egg=something``

But here ansible-builder doesn't move the req.uri forward. So this parsing of requirements needs a bit more attention to detail or the usage of req.line instead.

Best Regards, Uli

jplitza:
BTW, I worked around the issue by adding
```yaml
additional_build_steps:
  append:
    - RUN pip install --no-deps --force git+https://github.com/...@deadbeefdeadbeefdeadbeefdeadbeef
```

Shrews:
[...]

PR #336, opened 2021-12-21
Do not sanitize user python requirements

### LINKED ISSUE #334: ansible-builder introspect --sanitize removes yamllint and ansible-lint from requirements.txt
I want yamllint and ansible-lint to be available inside of my Execution Environment, so I added it to my requirements.txt file, but when the builder step runs `ansible-builder introspect --sanitize...` it strips this out from the resulting requirements.txt and is not available in the resulting EE.

I can work around this by adding a step to the append: section of additional_build_steps: to install it manually
```
    - RUN pip3 install yamllint ansible-lint
```

but this seems to go against the idea of listing all required python modules in the requirements.txt when they are removed unexpectedly by the builder stage

Shrews:
Fixes #334 

User Python requirements, specified via the `--user-pip` CLI option, will have duplicates removed, but will not go through the Python package exclusion process.

### DIFF
--- a/ansible_builder/requirements.py
+++ b/ansible_builder/requirements.py
@@ -18,9 +18,22 @@ EXCLUDE_REQUIREMENTS = frozenset((
 
 
 def sanitize_requirements(collection_py_reqs):
+    """
+    Cleanup Python requirements by removing duplicates and excluded packages.
+
+    The user requirements file will go through the deduplication process, but
+    skips the special package exclusion process.
+
+    :param dict collection_py_reqs: A dict of lists of Python requirements, keyed
+        by fully qualified collection name. The special key `user` holds requirements
+        from the user specified requirements file from the ``--user-pip`` CLI option.
+
+    :returns: A finalized list of sanitized Python requirements.
+    """
     # de-duplication
     consolidated = []
     seen_pkgs = set()
+
     for collection, lines in collection_py_reqs.items():
         try:
             for req in requirements.parse('\n'.join(lines)):
@@ -38,12 +51,13 @@ def sanitize_requirements(collection_py_reqs):
                 consolidated.append(req)
                 seen_pkgs.add(req.name)
         except Exception as e:
-            logger.warning('Warning: failed to parse requirments from {}, error: {}'.format(collection, e))
+            logger.warning('Warning: failed to parse requirements from {}, error: {}'.format(collection, e))
 
     # removal of unwanted packages
     sanitized = []
     for req in consolidated:
-        if req.name and req.name.lower() in EXCLUDE_REQUIREMENTS:
+        # Exclude packages, unless it was present in the user supplied requirements.
+        if req.name and req.name.lower() in EXCLUDE_REQUIREMENTS and 'user' not in req.collections:
             logger.debug(f'# Excluding requirement {req.name} from {req.collections}')
             continue
         if req.name is None and req.vcs:
--- a/docs/collection_metadata.rst
+++ b/docs/collection_metadata.rst
@@ -52,6 +52,10 @@ not be included in the combined file. These include test packages and
 packages that provide Ansible itself. The full list can be found in
 ``EXCLUDE_REQUIREMENTS`` in the ``ansible_builder.requirements`` module.
 
+Any requirements supplied in the user requirements file, via the ``--user-pip``
+option to the ``introspect`` command, will not be processed against the list of
+excluded Python packages.
+
 System-level Dependencies
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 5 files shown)


Shrews:
Likely need an option to `build` to pass this through.

docs/collection_metadata.rst
```diff
@@ -26,6 +26,29 @@ The short-hand form of the command is also supported::
 
     ansible-builder introspect --sanitize ~/.ansible/collections/
 
+The ``--sanitize`` option will go through all of the collection requirements and
+remove duplicates, as well as remove some Python requirements that should normally
+be excluded. For instance, these Python requirements will *always* be excluded
+since they should be a part of execution environment definition itself:
+
+   - ansible
```
dmsimard: I guess this depends on the version of ansible.

After 2.9, the ansible package isn't strictly necessary and this is reflected in the published images for ansible-runner, for example:
```
# for version in 2.12 2.11 2.10 2.9
> do
> podman run -it quay.io/ansible/ansible-runner:stable-${version}-devel pip freeze | grep ansible
> done
ansible-core @ file:///output/wheels/ansible_core-2.12.1.post0-py3-none-any.whl
ansible-runner @ file:///output/wheels/ansible_runner-2.1.2.dev2-py3-none-any.whl
ansible-core @ file:///output/wheels/ansible_core-2.11.7.post0-py3-none-any.whl
ansible-runner @ file:///output/wheels/ansible_runner-2.1.2.dev2-py3-none-any.whl
ansible-base @ file:///output/wheels/ansible_base-2.10.16.post0-py3-none-any.whl
ansible-runner @ file:///output/wheels/ansible_runner-2.1.1.dev14-py3-none-any.whl
ansible @ file:///output/wheels/ansible-2.9.27.post0-py3-none-any.whl
ansible-runner @ file:///output/wheels/ansible_runner-2.1.2.dev2-py3-none-any.whl
```

After 2.10, the ``ansible`` package on pypi moved to a model where it only includes a set of collections.
My use case was to mirror the UX of installing 'ansible' from PyPI and then bumped into [that issue](https://github.com/ansible/ansible-builder/issues/323#issuecomment-1000003661) (thanks for pointing me here)

I don't mind too much because there's simple workarounds like mentioned in that issue in addition to installing collections "manually" with a galaxy [requirements.yml](https://github.com/ansible-community/ansible-build-data/blob/main/5/galaxy-requirements.yaml).

I could have missed it but it would be great to print a message that says a package was "sanitized" to avoid the effect of surprise.

AlanCoding: > I could have missed it but it would be great to print a message that says a package was "sanitized" to avoid the effect of surprise.

I think that was attempted as a comment in the resultant requirements file, but with subsequent refactoring with the assemble script and other stuff, I think that file may wind up getting trashed, as more things were moved inside of ephemeral build stages.

https://github.com/ansible/ansible-builder/blob/87ae1d6a57cd6fca6d483929959f67527432af58/ansible_builder/requirements.py#L47

ansible_builder/requirements.py
```diff
@@ -43,7 +54,7 @@ def sanitize_requirements(collection_py_reqs):
     # removal of unwanted packages
     sanitized = []
     for req in consolidated:
-        if req.name and req.name.lower() in EXCLUDE_REQUIREMENTS:
+        if (req.name and (req.name.lower() in EXCLUDE_ALWAYS_REQUIREMENTS or (req.name.lower() in EXCLUDE_OPTIONAL_REQUIREMENTS and exclude))):
```
AlanCoding: Let me put forward my alternative suggestion here - instead of documenting a new thing, we could avoid sanitizing anything from the user's requirements.

I find it hazardous to turn off sanitization of a collection's dependencies. I also don't see why we should ever sanitize the user's requirements. This would be much easier for documentation, probably just a mention in the --sanitize option help text.

To implement this, it might be adding something like `and "user" not in req.collections`

Just an idea. I'm always in favor of adding less configuration options if it will accomplish the same goal.

Shrews: Hrm, I have to consider this some more, but I sort of like this solution at first read. I think we would still want to sanitize the user requirements to remove duplicates, but skip the part where we remove the excluded packages.

eqrx (approved):

Shrews:
recheck

AlanCoding (approved):
This looks great to me, thanks for incorporating the feedback.

docs/collection_metadata.rst
dmsimard: @AlanCoding would it make sense to bump the verbosity of that from debug to warning, perhaps ?

Shrews:
Since this is a bit of a behavior change, I don't think I'm going to backport this to `release_1.0`.

PR #337, opened 2022-01-11
Support multiple tags

### LINKED ISSUE #295: feature request: multiple tags on build
Currently specifying multiple tags like so: `ansible-builder build -t tag1 -t tag2`, results in `docker build -f context/Dockerfile -t tag2 context`

The underlying tools support multiple tags so it would be nice if specifying multiple tags with `ansible-builder` was possible as well.

Shrews:
Fixes #295

Supports multiple image tags. Example:

`ansible-builder build -t TAG1 -t TAG2 ...`

or

`ansible-builder build -t TAG1 TAG2 ...`

### DIFF
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -36,7 +36,7 @@ class AnsibleBuilder:
                  filename=constants.default_file,
                  build_args=None,
                  build_context=constants.default_build_context,
-                 tag=constants.default_tag,
+                 tag=[constants.default_tag],
                  container_runtime=constants.default_container_runtime,
                  output_filename=None,
                  no_cache=False,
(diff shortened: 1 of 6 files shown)


ansible_builder/main.py
```diff
@@ -119,7 +120,7 @@ def build_command(self):
         return command
 
     def build(self):
-        logger.debug(f'Ansible Builder is building your execution environment image, "{self.tag}".')
+        logger.debug('Ansible Builder is building your execution environment image. Tags: %s' % (', '.join(self.tags)))
```
eqrx: Why did you drop the format string? Not that concerned but curious:)

Shrews: Oops, I meant to go back and correct that before I pushed up the change. I'll correct it.

eqrx (approved):

PR #338, opened 2022-01-25
Allow VCS requirements with egg name

### LINKED ISSUE #323: Option to disable --sanitize / override collection dependencies

eqrx:
Current version converts requirements that are defined by a VCS
URI into their plain form. This change lets runner copy the full
line of said definitions to the destination file

Fixes #323

### DIFF
--- a/ansible_builder/requirements.py
+++ b/ansible_builder/requirements.py
@@ -46,7 +46,7 @@ def sanitize_requirements(collection_py_reqs):
         if req.name and req.name.lower() in EXCLUDE_REQUIREMENTS:
             logger.debug(f'# Excluding requirement {req.name} from {req.collections}')
             continue
-        if req.name is None and req.vcs:
+        if req.vcs:
             # A source control requirement like git+, return as-is
             new_line = req.line
         elif req.name:


eqrx:
recheck

eqrx:
recheck

sivel:
I've tested this and for the simple case seems to work correctly.  Just needs some tests.  Some examples:

```
>>> collection_py_reqs = {'user': ['git+https://github.com/baz/baz.git#egg=baz']}
>>> sanitize_requirements(collection_py_reqs)
['git+https://github.com/baz/baz.git#egg=baz  # from collection user']
```

The only problem is with something like this:

```
>>> collection_py_reqs = {'user': ['git+https://github.com/baz/baz.git#egg=baz'], 'foo.bar': ['baz==1.2.3']}
>>> sanitize_requirements(collection_py_reqs)
['git+https://github.com/baz/baz.git#egg=baz  # from collection user,foo.bar']
```

If we do that with just straight packages:

```
>>> collection_py_reqs = {'user': ['baz'], 'foo.bar': ['baz==1.2.3']}
>>> sanitize_requirements(collection_py_reqs)
['baz==1.2.3  # from collection user,foo.bar']
```

So the behavior with `#egg=` is backwards from plain package names, in that it drops the explicit version, instead of keeping the explicit version.

PR #279, opened 2021-08-27
Add CLI type subcommand

Shrews:
Preparation for adding new type subcommands (e.g., `srpm`, `ostree`, etc).

The following example invocations (still being supported):

- `ansible-build build -h`
- `ansible-build create -h`
- `ansible-build introspect -h`

are equivalent to the new format:

- `ansible-build container build -h`
- `ansible-build container create -h`
- `ansible-build container introspect -h`

PR #341, opened 2022-02-02
Revert PR #279

Shrews:
Too many changes have merged between PR #279 and now, so this is a manual revert. The feature we needed this change for is no longer going to be developed, so keeping this change would just add confusion.

This is not a complete revert as some code structural changes are left in place, as well as some tests.


Shrews:
Integration tests are currently failing due to issue fixed by PR #342. Will need to rebase this after that fix merges.

eqrx (approved):

PR #353, opened 2022-02-15
Use `devel` tag for the development image

Shrews:
The `devel` and `release_1.0` branches are publishing conflicting `latest` tagged images. Last merge wins.


eqrx:
recheck

eqrx (approved):

PR #357, opened 2022-03-03
Support collection signing verification

eqrx:
It was requested that builder supports the opportunistic signature
verification feature of ansible-galaxy. this commit adds one ~~two~~ new
flags. ~~`disable-gpg-verify` lets galaxy always ignore signatures
and is simply added verbatim to the ansible-galaxy invocation in
the container file.~~ `keyring` is the path of a GPG keyring file
that shall contain public keys that are used to verify signatures
of collections. This is forwarded to the ansible-galaxy command
by copying the given path into the build directory and passing
the path to the copied file from within the container to the
ansible-galaxy invocation

### DIFF
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -80,6 +80,19 @@ def add_container_options(parser):
         )
     )
 
+    # Keyring option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
+    create_command_parser.add_argument(
+        '--keyring',
+        help='The keyring used during signature verification when installing collections from a Galaxy server')
+    # Disable-gpg-verify option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
+    create_command_parser.add_argument(
+        '--disable-gpg-verify',
+        dest='disable_gpg_verify',
+        action='store_true',
+        help='Disable GPG signature verification when installing collections from a Galaxy server')
+    # Would also have copied over the signature option from ansible-galaxy but since it can't be specified
+    # with requirements-file set (we set that in any case) it is left out.
+
     build_command_parser = parser.add_parser(
         'build',
         help='Builds a container image.',
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -143,10 +143,16 @@ def add_container_options(parser):
                                 ' and '.join([' for '.join([v, k]) for k, v in constants.runtime_files.items()]))
                        )
 
-        # Keyring option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
-        # If this is not supplied, no verification is performed.
-        p.add_argument('--keyring',
-                       help='The keyring used for collection signature verification during installation from Galaxy.')
+        p.add_argument('--galaxy-keyring',
+                       help='Keyring for collection signature verification during installs from Galaxy. '
+                            'Will be copied into images. Verification is disabled if unset.')
+        p.add_argument('--galaxy-ignore-signature-status-codes',
+                       action="append",
+                       help='A gpg status code to ignore during signature verification when installing with '
+                       'ansible-galaxy. May be specified multiple times. See ansible-galaxy doc for more info.')
+        p.add_argument('--galaxy-required-valid-signature-count',
+                       help='The number of signatures that must successfully verify collections from '
+                       'ansible-galaxy ~if there are any signatures provided~. See ansible-galaxy doc for more info.')
 
     introspect_parser = parser.add_parser(
         'introspect',
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -80,19 +80,6 @@ def add_container_options(parser):
         )
     )
 
-    # Keyring option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
-    create_command_parser.add_argument(
-        '--keyring',
-        help='The keyring used during signature verification when installing collections from a Galaxy server')
-    # Disable-gpg-verify option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
-    create_command_parser.add_argument(
-        '--disable-gpg-verify',
-        dest='disable_gpg_verify',
-        action='store_true',
-        help='Disable GPG signature verification when installing collections from a Galaxy server')
-    # Would also have copied over the signature option from ansible-galaxy but since it can't be specified
-    # with requirements-file set (we set that in any case) it is left out.
-
     build_command_parser = parser.add_parser(
         'build',
         help='Builds a container image.',
@@ -155,6 +142,11 @@ def add_container_options(parser):
                             '(default depends on --container-runtime, {0})'.format(
                                 ' and '.join([' for '.join([v, k]) for k, v in constants.runtime_files.items()]))
                        )
+
+        # Keyring option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
+        # If this is not supplied, no verification is performed.
+        p.add_argument('--keyring',
+                       help='The keyring used for collection signature verification during installation from Galaxy.')
 
     introspect_parser = parser.add_parser(
         'introspect',
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -314,7 +314,8 @@ class Containerfile:
             filename = output_filename
         self.path = os.path.join(self.build_context, filename)
         self.container_runtime = container_runtime
-        self.keyring = keyring
+        self.original_keyring = keyring
+        self.keyring_copy = None
 
         # Build args all need to go at top of file to avoid errors
         self.steps = [
@@ -343,8 +344,9 @@ class Containerfile:
                 self.build_context, constants.user_content_subfolder, new_name)
             copy_file(requirement_path, dest)
 
-        if self.keyring:
-            copy_file(self.keyring, os.path.join(self.build_outputs_dir, 'keyring.gpg'))
+        if self.original_keyring:
+            self.keyring_copy = constants.default_keyring_name
+            copy_file(self.original_keyring, os.path.join(self.build_outputs_dir, self.keyring_copy))
 
         if self.definition.ansible_config:
             copy_file(
@@ -384,7 +386,7 @@ class Containerfile:
 
     def prepare_galaxy_install_steps(self):
         if self.definition.get_dep_abs_path('galaxy'):
-            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy'], self.keyring))
+            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy'], self.keyring_copy))
         return self.steps
 
     def prepare_introspect_assemble_steps(self):
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -40,10 +40,8 @@ class AnsibleBuilder:
                  output_filename=None,
                  no_cache=False,
                  verbosity=constants.default_verbosity,
-                 disable_gpg_verify=False,
                  keyring=None):
         """
-        :param bool disable_gpg_verify: Always disable signature verification of collections installed by ansible-galaxy.
         :param str keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
         """
         self.action = action
@@ -61,8 +59,7 @@ class AnsibleBuilder:
             build_context=self.build_context,
             container_runtime=self.container_runtime,
             output_filename=output_filename,
-            keyring=keyring,
-            disable_gpg_verify=disable_gpg_verify)
+            keyring=keyring)
         self.verbosity = verbosity
 
     @property
@@ -303,10 +300,8 @@ class Containerfile:
                  build_context=None,
                  container_runtime=None,
                  output_filename=None,
-                 disable_gpg_verify=False,
                  keyring=None):
         """
-        :param bool disable_gpg_verify: Always disable signature verification of collections installed by ansible-galaxy.
         :param str keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
         """
         self.build_context = build_context
@@ -320,7 +315,6 @@ class Containerfile:
         self.path = os.path.join(self.build_context, filename)
         self.container_runtime = container_runtime
         self.keyring = keyring
-        self.disable_gpg_verify = disable_gpg_verify
 
         # Build args all need to go at top of file to avoid errors
         self.steps = [
@@ -390,7 +384,7 @@ class Containerfile:
 
     def prepare_galaxy_install_steps(self):
         if self.definition.get_dep_abs_path('galaxy'):
-            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy'], self.keyring, self.disable_gpg_verify))
+            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy'], self.keyring))
         return self.steps
 
     def prepare_introspect_assemble_steps(self):
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -39,7 +39,13 @@ class AnsibleBuilder:
                  container_runtime=constants.default_container_runtime,
                  output_filename=None,
                  no_cache=False,
-                 verbosity=constants.default_verbosity):
+                 verbosity=constants.default_verbosity,
+                 disable_gpg_verify=False,
+                 keyring=None):
+        """
+        :param bool disable_gpg_verify: Always disable signature verification of collections installed by ansible-galaxy.
+        :param str keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
+        """
         self.action = action
         self.definition = UserDefinition(filename=filename)
 
@@ -54,7 +60,9 @@ class AnsibleBuilder:
             definition=self.definition,
             build_context=self.build_context,
             container_runtime=self.container_runtime,
-            output_filename=output_filename)
+            output_filename=output_filename,
+            keyring=keyring,
+            disable_gpg_verify=disable_gpg_verify)
         self.verbosity = verbosity
 
     @property
@@ -294,8 +302,13 @@ class Containerfile:
     def __init__(self, definition,
                  build_context=None,
                  container_runtime=None,
-                 output_filename=None):
-
+                 output_filename=None,
+                 disable_gpg_verify=False,
+                 keyring=None):
+        """
+        :param bool disable_gpg_verify: Always disable signature verification of collections installed by ansible-galaxy.
+        :param str keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
+        """
         self.build_context = build_context
         self.build_outputs_dir = os.path.join(
             build_context, constants.user_content_subfolder)
@@ -306,6 +319,9 @@ class Containerfile:
             filename = output_filename
         self.path = os.path.join(self.build_context, filename)
         self.container_runtime = container_runtime
+        self.keyring = keyring
+        self.disable_gpg_verify = disable_gpg_verify
+
         # Build args all need to go at top of file to avoid errors
         self.steps = [
             "ARG EE_BASE_IMAGE={}".format(
@@ -332,6 +348,9 @@ class Containerfile:
             dest = os.path.join(
                 self.build_context, constants.user_content_subfolder, new_name)
             copy_file(requirement_path, dest)
+
+        if self.keyring:
+            copy_file(self.keyring, os.path.join(self.build_outputs_dir, 'keyring.gpg'))
 
         if self.definition.ansible_config:
             copy_file(
@@ -371,7 +390,7 @@ class Containerfile:
 
     def prepare_galaxy_install_steps(self):
         if self.definition.get_dep_abs_path('galaxy'):
-            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy']))
+            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy'], self.keyring, self.disable_gpg_verify))
         return self.steps
 
     def prepare_introspect_assemble_steps(self):
--- a/ansible_builder/main.py
+++ b/ansible_builder/main.py
@@ -40,10 +40,18 @@ class AnsibleBuilder:
                  output_filename=None,
                  no_cache=False,
                  verbosity=constants.default_verbosity,
-                 keyring=None):
+                 galaxy_keyring=None,
+                 galaxy_required_valid_signature_count=None,
+                 galaxy_ignore_signature_status_codes=()):
         """
-        :param str keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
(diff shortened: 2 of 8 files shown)


ansible_builder/cli.py
```diff
@@ -80,6 +80,19 @@ def add_container_options(parser):
         )
     )
 
+    # Keyring option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
+    create_command_parser.add_argument(
```
Shrews: Both of these new options should be present for both `create` and `build` subcommands. There is a `for` loop down below that adds some common options.

ansible_builder/main.py
```diff
@@ -39,7 +39,13 @@ def __init__(self,
                  container_runtime=constants.default_container_runtime,
                  output_filename=None,
                  no_cache=False,
-                 verbosity=constants.default_verbosity):
+                 verbosity=constants.default_verbosity,
+                 disable_gpg_verify=False,
```
Shrews: This is a bit odd that the default does not match the CLI option default of `True`. Same in `Containerfile` init.

ansible_builder/steps.py
```diff
@@ -43,14 +43,19 @@ def __init__(self):
...
 
 class GalaxyInstallSteps(Steps):
-    def __init__(self, requirements_naming):
+    def __init__(self, requirements_naming, keyring, disable_gpg_verify):
         """Assumes given requirements file name has been placed in the build context
         """
+
+        install_opts = f"-r {requirements_naming} --collections-path {constants.base_collections_path}"
+        if keyring:
+            install_opts += " --keyring ./keyring.gog"
```
Shrews: `.gog` != `.gpg`  :)

Shrews (changes requested):
Some changes noted. Since I'm seeing about adding tests, I'll make the changes that I suggested.

ansible_builder/cli.py
```diff
@@ -80,6 +80,19 @@ def add_container_options(parser):
...
+        '--keyring',
+        help='The keyring used during signature verification when installing collections from a Galaxy server')
+    # Disable-gpg-verify option copied over from ansible-galaxy. Gets passed to the invocations of it verbatim.
+    create_command_parser.add_argument(
+        '--disable-gpg-verify',
+        dest='disable_gpg_verify',
+        action='store_true',
```
Shrews: We want the default to be to NOT verify collection signatures. This forces the user to use this option in order to prevent that.

If we want the user to specifically opt-in to verification, I think we'd have to go with an option like `--enable-collection-verification` (bikeshed the name) which defaults to `False`. If that option is not given, we use `--disable-gpg-verify` during collection installation. If it is given, then we avoid using that option and also pass the keyring value along. We also probably want to verify that a keyring value is actually given if we enable signature verification.

(edited for some wrong assumptions and confusing wording)

Shrews: Resolved this by turning on sig verify if `--keyring` is given. No need for the other option.

ansible_builder/steps.py
```diff
@@ -43,14 +43,19 @@ def __init__(self):
...
-    def __init__(self, requirements_naming):
+    def __init__(self, requirements_naming, keyring):
         """Assumes given requirements file name has been placed in the build context
         """
+
+        install_opts = f"-r {requirements_naming} --collections-path {constants.base_collections_path}"
+        if keyring:
+            install_opts += " --keyring ./keyring.gpg"
+        else:
+            install_opts += " --disable-gpg-verify"
```
Shrews: Looks like we'll have to use the [`ANSIBLE_GALAXY_DISABLE_GPG_VERIFY`](https://docs.ansible.com/ansible-core/devel/reference_appendices/config.html#galaxy-disable-gpg-verify) env var to control this since this option isn't going to be supported on older versions of Ansible.

Shrews:
There might still be a discussion around whether or not keyring should be specified in the EE definition file vs. a command line option. I personally don't see a reason why it couldn't just be a EE file value.

Shrews:
FYI, I'm uncertain if the relative path to the copied keyring (./keyring.gpg) is correct. Looking into it. 

Edit: tests modified to test that the keyring is copied to the user data dir (`_build`) in the build context dir.

Shrews:
FYI, builder is affected by https://github.com/ansible/ansible/issues/77349 since it does `role install` before `collection install`.

ansible_builder/steps.py
eqrx: Oh no, https://www.gog.com/ if branded into my brain, sorry xD

ansible_builder/main.py
eqrx: Actually a good point, thank you

eqrx (commented):
Reviewing PR created by me since @Shrews did a lot of work on it :) -> lgtm

Shrews:
> Reviewing PR created by me since @Shrews did a lot of work on it :) -> lgtm

FYI, this isn't exactly done yet. PR #https://github.com/ansible/ansible/pull/77026 was recently merged which adds a couple of new signature verification options (`--required-valid-signature-count` and `--ignore-signature-status-code`) which we will need to handle.

test/integration/test_create.py
```diff
@@ -50,3 +52,81 @@ def test_create_streams_output_with_invalid_verbosity(cli, build_dir_and_ee_yml)
...
+    ]
+    req = [
+        'collections:',
+        '  - name: community.general',
+    ]
+    tmpdir, eeyml = build_dir_and_ee_yml("\n".join(ee))
+    reqyml = tmpdir / "requirements.yml"
+    reqyml.write_text("\n".join(req))
+    cli(f'ansible-builder create -c {tmpdir} -f {eeyml} --output-filename Containerfile '
+        '--galaxy-ignore-signature-status-code 500 --galaxy-required-valid-signature-count 3')
```
Shrews, line 127: Are these options valid without specifying a keyring? If not, we probably should have a check for that early on in builder that errors out quickly for the user.

test/unit/test_steps.py
```diff
@@ -16,3 +17,25 @@ def test_additional_build_steps(verb):
...
+        f"RUN ansible-galaxy role install -r requirements.txt --roles-path {constants.base_roles_path}",
+
+        f"RUN ANSIBLE_GALAXY_DISABLE_GPG_VERIFY=1 ansible-galaxy collection install "
+        f"$ANSIBLE_GALAXY_CLI_COLLECTION_OPTS -r requirements.txt --collections-path {constants.base_collections_path}"
+    ]
+    assert steps == expected
+
+
+def test_galaxy_install_steps_with_keyring():
+    steps = list(GalaxyInstallSteps("requirements.txt", constants.default_keyring_name, None, None))
```
Shrews: Would be nice to see a unit test where these two values are not `None`.

ansible_builder/cli.py
```diff
@@ -143,6 +143,18 @@ def add_container_options(parser):
...
                        )
 
+        p.add_argument('--galaxy-keyring',
+                       help='Keyring for collection signature verification during installs from Galaxy. '
+                            'Will be copied into images. Verification is disabled if unset.')
+        p.add_argument('--galaxy-ignore-signature-status-code',
+                       help='A gpg status code to ignore during signature verification when installing with '
+                       'ansible-galaxy(for example, NO_PUBKEY).')
+        p.add_argument('--galaxy-required-valid-signature-count',
+                       help='The number of signatures that must successfully verify collections and rolesfrom '
```
Shrews: `roles` are not verified. There's no support for that.

I almost feel like it would be better if we didn't describe this too much and leave the galaxy documentation to fully explain. Your first sentence might be enough, plus "see docs for ansible-galaxy for further explanation" or similar.

ansible_builder/main.py
```diff
@@ -371,7 +401,11 @@ def prepare_build_context(self):
 
     def prepare_galaxy_install_steps(self):
         if self.definition.get_dep_abs_path('galaxy'):
-            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy']))
+            self.steps.extend(GalaxyInstallSteps(CONTEXT_FILES['galaxy'],
+                                                 self.copied_galaxy_keyring,
+                                                 self.galaxy_ignore_signature_status_code,
+                                                 self.galaxy_required_valid_signature_count))
```
Shrews, line 230: Mentioned in a comment elsewhere, but feels like we should probably verify `galaxy-keyring` is given if one or both of these other options are supplied. I don't think they're valid without it.

docs/usage.rst
```diff
@@ -75,6 +75,21 @@ To use a definition file named something other than
 
    $ ansible-builder build --file=my-ee.yml
 
+``--galaxy-keyring``
```
Shrews, line 78: Need some docs for the other new options.

eqrx:
retest

ansible_builder/main.py
```diff
@@ -294,7 +309,16 @@ class Containerfile:
...
                  container_runtime=None,
-                 output_filename=None):
+                 output_filename=None,
+                 keyring=None,
+                 galaxy_keyring=None,
+                 galaxy_required_valid_signature_count=None,
+                 galaxy_ignore_signature_status_code=None):
+        """
+        :param str galaxy_keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
+        :param str galaxy_required_valid_signature_count: Number of sigs (prepend + to disallow no sig( required for ansible-galaxy to accept collections.
```
Shrews:
```suggestion
        :param str galaxy_required_valid_signature_count: Number of sigs (prepend + to disallow no sig) required for ansible-galaxy to accept collections.
```

ansible_builder/steps.py
```diff
@@ -43,14 +43,36 @@ def __init__(self):
 
 
 class GalaxyInstallSteps(Steps):
-    def __init__(self, requirements_naming):
-        """Assumes given requirements file name has been placed in the build context
+    def __init__(self, requirements_naming, galaxy_keyring, galaxy_ignore_signature_status_code, galaxy_required_valid_signature_count):
+        """Assumes given requirements file name and keyring has been placed in the build context.
+
+        :param str galaxy_keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
+        :param str galaxy_required_valid_signature_count: Number of sigs (prepend + to disallow no sig( required for ansible-galaxy to accept collections.
```
Shrews:
```suggestion
        :param str galaxy_required_valid_signature_count: Number of sigs (prepend + to disallow no sig) required for ansible-galaxy to accept collections.
```

docs/usage.rst
```diff
@@ -75,6 +75,47 @@ To use a definition file named something other than
...
+
+.. code::
+
+   $ ansible-builder create --galaxy-keyring=/path/to/pubring.kbx
+   $ ansible-builder build --galaxy-keyring=/path/to/pubring.kbx
+
+``--galaxy-ignore-signature-status-code``
+*****************************************
+
+With --galaxy-keyring set it is possible to ignore certain errors happening while
```
Shrews:
```suggestion
With ``--galaxy-keyring`` set, it is possible to ignore certain errors that may occur while
```

docs/usage.rst
```diff
@@ -75,6 +75,47 @@ To use a definition file named something other than
...
+verifying collections. It is passed unmodified to ansible-galaxy calls via the option
```
Shrews:
```suggestion
verifying collections. It is passed unmodified to ``ansible-galaxy`` calls via the option
```

docs/usage.rst
```diff
@@ -75,6 +75,47 @@ To use a definition file named something other than
...
+--ignore-signature-status-code. See the ansible-galaxy doc for more info.
```
Shrews:
```suggestion
``--ignore-signature-status-code``. See the ``ansible-galaxy`` documentation for more information.
```

docs/usage.rst
```diff
@@ -75,6 +75,47 @@ To use a definition file named something other than
...
+
+.. code::
+
+   $ ansible-builder create --galaxy-keyring=/path/to/pubring.kbx --galaxy-ignore-signature-status-code 500
+   $ ansible-builder build --galaxy-keyring=/path/to/pubring.kbx --galaxy-ignore-signature-status-code 500
+
+``--galaxy-required-valid-signature-count``
+*******************************************
+
+With --galaxy-keyring set it is possible specify how many valid signatures must present for each collection item.
```
Shrews:
```suggestion
With ``--galaxy-keyring`` set, it is possible to specify how many valid signatures must be present for each collection item.
```

docs/usage.rst
```diff
@@ -75,6 +75,47 @@ To use a definition file named something other than
...
+Bear in mind that if this is set to an number collections having no signatures at all still pass.
```
Shrews:
```suggestion
```

I'd prefer to leave discussion of the details of what ansible-galaxy does with this value to the galaxy docs.

docs/usage.rst
```diff
@@ -75,6 +75,47 @@ To use a definition file named something other than
...
+The value is passed unmodified to ansible-galaxy calls via the option
```
Shrews:
```suggestion
The value is passed unmodified to ``ansible-galaxy`` calls via the option
```

docs/usage.rst
```diff
@@ -75,6 +75,47 @@ To use a definition file named something other than
...
+--galaxy-required-valid-signature-count. See the ansible-galaxy doc for more info.
```
Shrews:
```suggestion
``--required-valid-signature-count``. See the ``ansible-galaxy`` documentation for more information.
```

test/integration/test_create.py
```diff
@@ -50,3 +52,83 @@ def test_create_streams_output_with_invalid_verbosity(cli, build_dir_and_ee_yml)
...
+    text = containerfile.read_text()
+
+    keyring_copy = tmpdir / constants.user_content_subfolder / constants.default_keyring_name
+    assert keyring_copy.exists()
+
+    assert "RUN ANSIBLE_GALAXY_DISABLE_GPG_VERIFY=1 ansible-galaxy" not in text
+    assert f"--keyring {constants.default_keyring_name}" in text
+
+
+def test_galaxy_signing_extra_args(cli, build_dir_and_ee_yml):
```
Shrews, line 109: Let's add a similar test to `integration/test_build.py`.

Shrews (changes requested):
Looks pretty good to me. Just a few doc-related suggested changes, and requesting an additional test for the `build` portion.

ansible_builder/main.py
```diff
@@ -39,7 +39,19 @@ def __init__(self,
...
                  output_filename=None,
                  no_cache=False,
-                 verbosity=constants.default_verbosity):
+                 verbosity=constants.default_verbosity,
+                 galaxy_keyring=None,
+                 galaxy_required_valid_signature_count=None,
+                 galaxy_ignore_signature_status_code=None):
+        """
+        :param str galaxy_keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
+        :param str galaxy_required_valid_signature_count: Number of sigs (prepend + to disallow no sig( required for ansible-galaxy to accept collections.
```
Shrews, line 31: Still have a ( here that should be a ).

docs/usage.rst
```diff
@@ -75,6 +75,46 @@ To use a definition file named something other than
...
+
+.. code::
+
+   $ ansible-builder create --galaxy-keyring=/path/to/pubring.kbx
+   $ ansible-builder build --galaxy-keyring=/path/to/pubring.kbx
+
+``--galaxy-ignore-signature-status-code``
+*****************************************
+
+With ``--galaxy-keyring ``set it is possible to ignore certain errors that may occur while while verifying collections.
```
Shrews:
```suggestion
With ``--galaxy-keyring`` set it is possible to ignore certain errors that may occur while while verifying collections.
```

test/integration/test_build.py
```diff
@@ -187,3 +189,48 @@ def test_build_layer_reuse(cli, runtime, data_dir, ee_tag, tmp_path):
...
+    result = cli(f'ansible-builder build --galaxy-keyring {keyring} -c {tmp_path} -f {ee_def} -t {ee_tag} --container-runtime {runtime} -v 3', allow_error=True)
+
+    keyring_copy = tmp_path / constants.user_content_subfolder / constants.default_keyring_name
+    assert keyring_copy.exists()
+
+    assert "RUN ANSIBLE_GALAXY_DISABLE_GPG_VERIFY=1 ansible-galaxy" not in result.stdout
+    assert f"--keyring {constants.default_keyring_name}" in result.stdout
+
+
+#@pytest.mark.test_all_runtimes
```
Shrews: Did you mean to leave this commented out?

eqrx: yes :< I was unsure about fixing a part of it, working on it now

ansible_builder/main.py
eqrx: oh .... did not notice I copy pasted that in 5 locations, thanks!

Shrews (approved):
LGTM. Not sure if @nitzmahone wants to review or not.

ansible_builder/cli.py
```diff
@@ -143,6 +143,16 @@ def add_container_options(parser):
                                 ' and '.join([' for '.join([v, k]) for k, v in constants.runtime_files.items()]))
                        )
 
+        p.add_argument('--galaxy-keyring',
+                       help='Keyring for collection signature verification during installs from Galaxy. '
+                            'Will be copied into images. Verification is disabled if unset.')
+        p.add_argument('--galaxy-ignore-signature-status-code',
```
nitzmahone: This is a multi-value `type='append'` option in the Galaxy CLI, so this one probably needs to be as well- then we'll need to generate one `--ignore-signature-status-code` for each value we got. Could probably also update the description to match Galaxy CLI re: specifying multiple values.

ansible_builder/steps.py
```diff
@@ -43,14 +43,36 @@ def __init__(self):
...
+
+        :param str galaxy_keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
+        :param str galaxy_required_valid_signature_count: Number of sigs (prepend + to disallow no sig) required for ansible-galaxy to accept collections.
+        :param str galaxy_ignore_signature_status_code: GPG Status code to ignore when validating galaxy collections.
         """
+
+        env = ""
+        install_opts = f"-r {requirements_naming} --collections-path {constants.base_collections_path}"
+
+        if galaxy_ignore_signature_status_code:
```
nitzmahone: multi-value support needed (see above)

ansible_builder/steps.py
```diff
@@ -43,14 +43,36 @@ def __init__(self):
...
+    def __init__(self, requirements_naming, galaxy_keyring, galaxy_ignore_signature_status_code, galaxy_required_valid_signature_count):
+        """Assumes given requirements file name and keyring has been placed in the build context.
+
+        :param str galaxy_keyring: GPG keyring file used by ansible-galaxy to opportunistically validate collection signatures.
+        :param str galaxy_required_valid_signature_count: Number of sigs (prepend + to disallow no sig) required for ansible-galaxy to accept collections.
+        :param str galaxy_ignore_signature_status_code: GPG Status code to ignore when validating galaxy collections.
         """
+
+        env = ""
+        install_opts = f"-r {requirements_naming} --collections-path {constants.base_collections_path}"
```
nitzmahone: We should probably quote these values in the generated cmdline string (eg, paths with spaces)

ansible_builder/steps.py
```diff
@@ -43,14 +43,36 @@ def __init__(self):
...
+            # the Containerfile since we need it only during the build and not
+            # in the final image.
+            env = "ANSIBLE_GALAXY_DISABLE_GPG_VERIFY=1 "
+
         self.steps = [
-            "RUN ansible-galaxy role install -r {0} --roles-path {1}".format(
-                requirements_naming, constants.base_roles_path),
-            "RUN ansible-galaxy collection install $ANSIBLE_GALAXY_CLI_COLLECTION_OPTS -r {0} --collections-path {1}".format(
-                requirements_naming, constants.base_collections_path),
+            f"RUN ansible-galaxy role install -r {requirements_naming} --roles-path {constants.base_roles_path}",
```
nitzmahone: also quoted arg values

docs/usage.rst
```diff
@@ -75,6 +75,46 @@ To use a definition file named something other than
...
+
+.. code::
+
+   $ ansible-builder create --galaxy-keyring=/path/to/pubring.kbx
+   $ ansible-builder build --galaxy-keyring=/path/to/pubring.kbx
+
+``--galaxy-ignore-signature-status-code``
+*****************************************
+
+With ``--galaxy-keyring`` set it is possible to ignore certain errors that may occur while while verifying collections.
```
nitzmahone: s/while while/while/ ;)

docs/usage.rst
```diff
@@ -75,6 +75,46 @@ To use a definition file named something other than
...
+
+.. code::
+
+   $ ansible-builder create --galaxy-keyring=/path/to/pubring.kbx --galaxy-ignore-signature-status-code 500
+   $ ansible-builder build --galaxy-keyring=/path/to/pubring.kbx --galaxy-ignore-signature-status-code 500
+
+``--galaxy-required-valid-signature-count``
+*******************************************
+
+With ``--galaxy-keyring`` set it is possible specify constrainsts of collection validation. The value is
```
nitzmahone:
```suggestion
When ``--galaxy-keyring`` is set, the number of required valid collection signatures can be overridden. The value is
```

test/integration/test_build.py
```diff
@@ -187,3 +189,52 @@ def test_build_layer_reuse(cli, runtime, data_dir, ee_tag, tmp_path):
...
+    Test that all extr asigning args for gpg are passed into the container file.
+    """
+    pytest.xfail("failing configuration (but should work)")
+
+    keyring = tmp_path / "mykeyring.gpg"
+    keyring.touch()
+    ee_def = data_dir / 'ansible.posix.at' / 'execution-environment.yml'
+
+    result = cli(f'ansible-builder build -c {tmp_path} -f {ee_def} -t {ee_tag} --container-runtime {runtime} -v 3 '
+                 f'--galaxy-keyring {keyring} --galaxy-ignore-signature-status-code 500 '
```
nitzmahone, line 236: might as well use legit ignore codes in here as well (see above)

test/integration/test_build.py
```diff
@@ -187,3 +189,52 @@ def test_build_layer_reuse(cli, runtime, data_dir, ee_tag, tmp_path):
...
+                 f'--galaxy-required-valid-signature-count 3', allow_error=True)
+
+    assert "--galaxy-ignore-signature-status-code 500" in result.stdout
+    assert "--galaxy-required-valid-signature-count 3" in result.stdout
```
nitzmahone, line 240: Should probably add a multi-value test for `--ignore-signature-status-code` as well

nitzmahone (changes requested):
Sorry for the slow re-review- a couple little things to adjust around the new ignore code behavior, but looks good otherwise.

nitzmahone:
I've also inquired about maybe changing Galaxy CLI's parse action on that arg to `extend` instead of `append`, which would let us generate a slightly less verbose cmdline if we set ours to match.

nitzmahone (approved):

PR #370, opened 2022-04-18
Cleanup UserDefinition class

Shrews:
Doing some cleanup of the `UserDefinition` class:

- Combine `UserDefinition` with the unnecessary `BaseDefinition` class.
- Convert all strings to f-strings.
- Use more descriptive temp var names instead of `f` and `y`.
- Add missing inline method/property documentation.
- Replace use of `.items()` to iterate through dict keys.
- `ansible_config` property now returns `None` directly instead of just using `pass`.
- Correct spelling error: `occured` -> `occurred`

### DIFF
--- a/ansible_builder/user_definition.py
+++ b/ansible_builder/user_definition.py
@@ -15,53 +15,38 @@ ALLOWED_KEYS = [
 ]
 
 
-class BaseDefinition:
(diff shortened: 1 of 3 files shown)


eqrx (approved):
Very nice!

PR #371, opened 2022-04-18
Container image signature validation

Shrews:
This implements container image signature validation with a new execution environment schema format and new CLI options to control sig validation policy.

Changes:
- Assume version 1 of the EE def if a version is not specified. This ensures backward compatibility since we didn't actually enforce a version to be present.
- Support for a version 2 of the EE def file, allowing for base and builder images to be signature verified.
- New `--container-policy` and `--container-keyring` options for `build` action.

The pulp-integration jobs are successful against my fork, [seen here](https://github.com/Shrews/ansible-builder/runs/7414461352?check_suite_focus=true).


ansible_builder/main.py
```diff
@@ -59,6 +64,63 @@ def __init__(self,
...
+        """
+        Process the container_policy and container_keyring arguments.
+
+        :param str policy: The container_policy value.
+        :param str keyring: The container_keyring value.
+
+        The container_policy and container_keyring arguments come from the CLI
+        and work together to help build or use a podman policy.json file used
+        do image validation. Depending on the policy being used, the keyring
+        may or may not be necessary.
```
Akasurde:
```suggestion
        The container_policy and container_keyring arguments come from the CLI
        and work together to help build or use a podman policy.json file used
        to do image validation. Depending on the policy being used, the keyring
        may or may not be necessary.
```

ansible_builder/main.py
```diff
@@ -59,6 +64,63 @@ def __init__(self,
...
+                raise ValueError('--container-keyring requires --container-policy')
+            elif resolved_policy != PolicyChoices.SIG_REQ:
+                raise ValueError(f'--container-keyring is not valid with --container-policy={resolved_policy.value}')
+
+            # Set the keyring to an absolute path to be referenced in the policy file.
+            if not os.path.exists(keyring):
+                raise ValueError('--container-keyring error: file does not exist')
+            if not os.path.isfile(keyring):
+                raise ValueError('--container-keyring error: not a file')
+            resolved_keyring = os.path.abspath(keyring)
```
Akasurde, line 121:
```suggestion
            resolved_keyring = os.path.abspath(keyring)
            logger.debug('Using Container Keyring from file %s', resolved_keyring)
```

Useful, not required.

Shrews: This is already logged in `build_command()` a bit further down in this file, line 196.

ansible_builder/main.py
```diff
@@ -59,6 +64,63 @@ def __init__(self,
...
+
+            # Require keyring if we write a policy file
+            if resolved_policy == PolicyChoices.SIG_REQ and keyring is None:
+                raise ValueError(f'--container-policy={resolved_policy.value} requires --container-keyring')
+
+            # Do not allow images to be defined with --build-arg CLI option if
+            # any sig policy is defined.
+            for key, _ in self.build_args.items():
+                if key in ('EE_BASE_IMAGE', 'EE_BUILDER_IMAGE'):
+                    raise ValueError(f'{key} not allowed in --build-arg option with --container-policy={resolved_policy.value}')
```
Shrews: This exception message should probably not reference container policy since defining these images is invalid for version 2 format. Something more like:

```
EE_BASE_IMAGE not allowed to be defined with version 2 format
```

ansible_builder/user_definition.py
```diff
@@ -37,41 +82,35 @@ def __init__(self, filename):
...
-        version = self.raw.get('version')
-        if not version:
-            raise ValueError("Expected top-level 'version' key to be present.")
+        """
+        Version of the EE file.
+
+        If no version is specified, assume version 1 (for backward compat).
+        """
+        version = self.raw.get('version', 1)
         return str(version)
```
nitzmahone, line 114: Hrm, seems like that should be an `int` for easier comparison, but it looks like it's always been `str`... Weird.

eqrx (approved):

nitzmahone (approved):
Yep, looks good- one question about the return type of the `version` property on `UserDefinition` that I never noticed, but it doesn't look like this PR introduced that change, so we can probably ignore it here.

ansible_builder/user_definition.py
Shrews: Yeah. We can always change it later.

PR #386, opened 2022-06-02
Remove unused Containerfile parameter

Shrews:
The `keyring` parameter in `Containerfile.__init__()` is unused (added in PR #357).


eqrx (approved):
My bad, thanks for finding it! :)

PR #400, opened 2022-06-29
Add pulp_integration framework

Shrews:
This adds a new `pulp_integration` test framework that will allow us to have a local image repository that supports signed images. The PR that adds support for container image signature validation (PR #371) will be able to use this to add integration tests.

Note that the Pulp Integration job fails in this PR because it is adding a new GHA job that uses a secret. It will not have access to this secret until this is merged, but I verified that it does work as a PR against my own fork, seen here: ~~https://github.com/Shrews/ansible-builder/runs/7134536766?check_suite_focus=true~~
~~https://github.com/Shrews/ansible-builder/runs/7201566704?check_suite_focus=true~~
https://github.com/Shrews/ansible-builder/runs/7254569323?check_suite_focus=true


.github/workflows/ci.yml
```diff
@@ -36,6 +36,53 @@ jobs:
...
+            tox_env: pulp-integration-py310
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install Python ${{ matrix.py_version.name }}
+        uses: actions/setup-python@v2
+        with:
+          py-version: ${{ matrix.py_version.name }}
```
nitzmahone: This should be `python-version`- note the warning at https://github.com/Shrews/ansible-builder/runs/7134536766?check_suite_focus=true#step:3:1 . It's currently doing the right thing, but only because the default `python-version: 3.x` the action declares currently resolves to `3.10`.

setup_pulp.sh
```diff
@@ -0,0 +1,106 @@
...
+           pulp/pulp
+
+##############################################################################
+# Iteratively query the REST API until we get a JSON response (http code will
+# be 200) which will indicate the system is ready.
+##############################################################################
+
+set +e
+iter=0
+httpcode=$(curl -o /dev/null -s -w "%{http_code}\n" localhost:8080/pulp/api/v3/status/)
```
nitzmahone: Any reason to repeat this outside the loop instead of just initing to empty (or not at all, which IIRC it won't care about anyway)?

setup_pulp.sh
```diff
@@ -0,0 +1,106 @@
...
+
+until [ "$httpcode" = "200" ]
+do
+  iter=$((iter+1))
+  if [ $iter == 20 ]; then
+    exit 1
```
nitzmahone: Given core's past experience with the pulp container DB init just going to lunch and never coming back once in awhile, might be good to add an error message if this exit trips.

setup_pulp.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
```
nitzmahone: This file should probably be under .github, not at the root of the repo- maybe `.github/test-scripts` or something?

test/pulp_integration/test_policies.py
```diff
@@ -0,0 +1,2 @@
+def test_policy():
+    assert 1 == 1
```
nitzmahone, line 3: I assume this is basically `# FIXME: implement`, or ?

nitzmahone (commented):
A couple suggestions, but I don't see anything terribly off-the-rails here for your first GHA experience ;)

setup_pulp.sh
eqrx: Question from me: Is is still common to have the interpreter under `/bin`? Have my bash under `/usr/bin` since I remember :) (`/bin` links to `/usr/bin`)

setup_pulp.sh
```diff
@@ -0,0 +1,106 @@
...
+
+#
+# This script sets up a pulp all-in-one container (per the instructions at
+# https://pulpproject.org/pulp-in-one-container/) that the builder tests can
+# then use as an image repository that supports image signatures.
+#
+
+set -ex
```
eqrx: Would recommend setting `-u` to catch unset variables.

Cherry on top: `set -eu -o pipefail`

But since this script has neither vars nor I would find it understandable if we keep it like that

eqrx (changes requested):

setup_pulp.sh
Shrews: I dunno. Less typing this way? I can modify it.

.github/workflows/ci.yml
Shrews: Oopsie. Done

setup_pulp.sh
Shrews: Done

setup_pulp.sh
Shrews: Done

setup_pulp.sh
Shrews: Done

test/pulp_integration/test_policies.py
Shrews: Yup. This is where the next PR would pick up work, modifying this test file. Added a comment for clarity.

setup_pulp.sh
Shrews: Added `-u`

Shrews:
Lastest working run against my fork: https://github.com/Shrews/ansible-builder/runs/7201566704?check_suite_focus=true

eqrx (approved):

.github/test-scripts/setup_pulp.sh
```diff
@@ -0,0 +1,107 @@
...
+# See https://github.com/containers/skopeo/issues/942 for more info.
+##############################################################################
+
+export XDG_RUNTIME_DIR=/tmp/pulptests
+mkdir $XDG_RUNTIME_DIR
+
+skopeo login --username admin --password password localhost:8080 --tls-verify=false
+skopeo copy docker://registry.access.redhat.com/ubi9/ubi-micro:latest docker://localhost:8080/testrepo/ubi-micro --dest-tls-verify=false
+
+podman login --username '6340056|ansible-runner-gha' --password "$1" registry.redhat.io
```
eqrx: Maybe guard against `$1` being empty? Would require an if block, decreases readability of the code but would report something more obvious than ENOTTY when the password secret is not set. Not sure if I would actually implement that since this is a shell script and bound to fail anyway :D

Shrews: Added a check to exit early if pwd is empty (should never happen once merged, really).

.github/test-scripts/setup_pulp.sh
```diff
@@ -0,0 +1,117 @@
...
```
nitzmahone: Thinking about this one a little more- the username should probably also be an arg provided by a GHA secret- not so much that it's a secret, but if someone wanted to run this from a fork, they'd have to provide their own token, which would have a different username. Since GHA doesn't really have a general concept of external "vars" that aren't secrets, a secret it'd have to be. :(

Shrews: Done.

Shrews: Seen as secrets here: https://github.com/Shrews/ansible-builder/runs/7254569323?check_suite_focus=true#step:6:149

nitzmahone (approved):

nitzmahone:
Merging, should go green once it has access to local secrets. We may need to retarget the pulp job to only run on maintainer-sourced PRs or merge jobs only so it doesn't fail on all future PRs.

Issue #196, opened 2021-03-29
Allow specifying secrets to download collections from custom locations


ghost:
A general use case here would be to download collections from a private git repository, there should be a way to specify a token/key to be mounted onto the container for this purpose

JustInVTime:
:+1:

AlanCoding:
Going to close as duplicate of https://github.com/ansible/ansible-builder/issues/195

If there's any specific means of passing secrets, please contribute to the discussion there. Thanks.

Issue #195, opened 2021-03-12
need to support custom CA certificates and proxy settings


ChetRHosey:
We use a transparent proxy which does MITM inspection of some HTTPS traffic. This means that some HTTPS connections are re-signed by an organizational CA rather than the original CA.

This interferes with ansible-builder, as it's not clear how we would add the custom CA root(s) to the builder image, and it fails to validate the connection to Galaxy:

~~~
#8 0.851 ERROR! Unknown error when attempting to call Galaxy at 'https://galaxy.ansible.com/api/': <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1108)>
#8 ERROR: executor failed running [/bin/sh -c ansible-galaxy collection install $ANSIBLE_GALAXY_CLI_COLLECTION_OPTS -r requirements.yml --collections-path /usr/share/ansible/collections]: exit code: 1
~~~

Full output follows.

Some of our sites require explicit HTTP/HTTPS proxy settings, so ideally it would be possible to set those as well.

~~~
$ ansible-builder --version
0.4.0
$ ansible-builder build
Running command:
  docker build -f context/Dockerfile -t ansible-execution-env:latest context
#1 [internal] load build definition from Dockerfile
#1 sha256:da308b69c518bf933c2fe118954bb242105f66dbab8e01c524ce9c0149d68625
#1 transferring dockerfile: 517B done
#1 DONE 0.0s

#2 [internal] load .dockerignore
#2 sha256:5a456dc77426d604b098c2d1d21b47a02ba4e7b7ec117389ab91b07a62a31324
#2 transferring context: 2B done
#2 DONE 0.0s

#3 [internal] load metadata for quay.io/ansible/ansible-runner:devel
#3 sha256:a2ee052f444f48b483ef7462f9f15dc496066ca125add16c92e38bb837b6dd2b
#3 DONE 0.6s

#4 [1/5] FROM quay.io/ansible/ansible-runner:devel@sha256:9873fcce94e9ac5eea49a6a8be051271d1eaf37269bcb7cde0e43a045ef7bbbb
#4 sha256:753a06865a330ab0144695c33d3db1821431ab90998e5e86162a89f396accf74
#4 DONE 0.0s

#9 [internal] load build context
#9 sha256:6b5546cd88b98a90418e016afd6fcc2e1b9e7db9cd9abf195d0aee0db43c718e
#9 transferring context: 70B done
[...]

PR #404, opened 2022-07-12
Add support of build mounts

stanislav-zaprudskiy:
* https://docs.docker.com/build/building/secrets/
* https://docs.podman.io/en/latest/markdown/podman-run.1.html#secret-secret-opt-opt

A possible implementation of https://github.com/ansible/ansible-builder/issues/196 (but is not related to https://github.com/ansible/ansible-builder/issues/195)

### DIFF
--- a/src/ansible_builder/containerfile.py
+++ b/src/ansible_builder/containerfile.py
@@ -24,7 +24,8 @@ class Containerfile:
                  output_filename: str | None = None,
                  galaxy_keyring: str | None = None,
                  galaxy_required_valid_signature_count: int | None = None,
-                 galaxy_ignore_signature_status_codes: list | None = None
+                 galaxy_ignore_signature_status_codes: list | None = None,
+                 mounts: list | None = None,
                  ) -> None:
         """
         Initialize a Containerfile object for instruction file creation.
@@ -53,6 +54,7 @@ class Containerfile:
         self.copied_galaxy_keyring = None
         self.galaxy_required_valid_signature_count = galaxy_required_valid_signature_count
         self.galaxy_ignore_signature_status_codes = galaxy_ignore_signature_status_codes
+        self.mounts = " ".join(f"--mount={m}" for m in mounts or [])
         self.steps: list = []
 
     def prepare(self) -> None:
@@ -207,6 +209,7 @@ class Containerfile:
         """
         with open(self.path, 'w') as f:
             for step in self.steps:
+                step = self._add_mounts_for_run_step(step)
                 f.write(step + self.newline_char)
 
     def _insert_global_args(self, include_values: bool = False) -> None:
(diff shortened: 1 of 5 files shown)


stanislav-zaprudskiy:
I thought of extending the user definition file schema to allow passing mounts and secrets too. Thus, these lengthy arguments would persist in file, instead of needing to re-type them every time.

However, this doesn't really come along with _somewhat_ similar `build_arg_defaults` - which only allows passing the `ansible-builder` provided arguments, and not any random ones.

stanislav-zaprudskiy:
Another thing is that perhaps updating the integration tests and docs would be needed.

stanislav-zaprudskiy:
I force-pushed to incorporate latest `ansible:devel` changes causing merge conflicts previously.

Kaezon:
I just did some testing, and I ran into a problem where ansible-builder was selecting the podman runtime while I was specifying `--ssh`. Since podman doesn't support secret mounts, there should probably be a check to bail out if the selected runtime is podman.

Bit of a side-note, I noticed that the changed files have since moved. Some code was moved into `containerfile.py` as well.
It only required a few minutes of copy-pasting to correct though.

stanislav-zaprudskiy:
I plan to update the PR soon, just haven't found time yet to catch up with the recent `ansible-builder` changes.

stanislav-zaprudskiy:
@Shrews, @Akasurde, would you give this a review?

Shrews:
I'll take a look at this shortly. Thanks for keeping it updated.

src/ansible_builder/containerfile.py
```diff
@@ -207,6 +209,7 @@ def write(self) -> None:
         """
         with open(self.path, 'w') as f:
             for step in self.steps:
+                step = self._add_mounts_for_run_step(step)
```
Shrews, line 212: This is adding the list of mounts to _every_ `RUN` command. Definitely do not like that as it's very much overkill. I think if we are going to allow secrets, then we should assume that the user is going to manually add `RUN` statements that will use those secrets to the EE definition file under `additional_run_steps` where they will manually specify the mounts, which means we could eliminate the `--mounts` option entirely.

Shrews (changes requested):
So, we definitely need some documentation, ideally with examples, about these new features/options.

How the mounts is being handled isn't very clean, IMO. Not every RUN command needs those mounts. I can only see manually inserted RUN commands (via `additional_build_steps`) ever needing to access the secrets. Unless I've missed something?

I'm not sure about the `--ssh` option, but for the `--secret` option, I believe we could create an integration test for that. It could add a secret from a file, cat that secret from the RUN command we add to the EE, and we could check the build output for the secret contents. This would validate that it works all the way through, unlike the unit tests.

Shrews:
Come to think of it, I think we could wrap both of these new options up in a more generic --extra-cli-opts argument (or similar) that we just pass through to the build command.

Shrews:
I was able to quickly hack up a proof of concept for my suggestion, with a working test proving that secrets work in PR #677. I may need to clean that up a bit with some comments, error checking, unit tests, adding docs, etc (just quickly vomited some code to see if it would work).

Shrews:
Now that PR #677 is merged and released in version `3.1`, I'm going to go ahead and close this. I added a [scenario doc](https://ansible.readthedocs.io/projects/builder/en/3.1.0/scenario_guides/scenario_secret_passing/) that shows how to use this new feature to pass a secret.

Issue #293, opened 2021-10-06
feature request: ability to execute commands BEFORE ansible-galaxy run


fibbs:
In execution_environment.yml, you can provide additional commands to be executed to build the actual ee image. This way, I was able to install extra software packages ("figlet" in my case, for example, which I use to generate lovely looking motds for my servers).
A problem I came across is that I wanted to install Ansible roles from a corporate git server (gitlab) within the requirements.yml file. This was not possible because the base EE image used by default of course does not trust my company's CA certificate structure.
I have solved this by creating my own base image which I then use to build the ee image, but from my point of view this is not optimal. It would be great to have a possibility right inside the execution_environment.yml to supply "pre-requirements-installation customizations", so that I can copy my CA certificate, eventually needed SSH priv keys for auth, ... or whatever I would need to be able to work through the requirements.yml and requirements.txt.

nwerker:
+1 

This would be very helpful for any OS Dependencies!

lucab85:
I'm looking for a way to to specify some commands to apply on the `$EE_BUILDER_IMAGE` image.

At the moment it's possible to specify commands on `$EE_BASE_IMAGE` before (`prepend`) or after (`append`) via `additional_build_steps` parameter in `execution-environment.yml`.

Same use-case about SSL certificates.

lucab85:
+1

SROBER28:
+1

jskacel:
@lucab85 `prepend` is actually not running before galaxy install command.. :-/

```
---
version: 0.1

build_arg_defaults:
  EE_BASE_IMAGE: 'registry.redhat.io/ansible-automation-platform-22/ee-29-rhel8'

additional_build_steps:
  prepend: |
    COPY RootCA.crt /etc/pki/ca-trust/source/anchors/
    RUN /usr/bin/update-ca-trust

dependencies:
  galaxy: requirements.yml
```

Ends with:
```
# ansible-builder build -t jskacel/ee-image --prune-images --no-cache -v3
Ansible Builder is building your execution environment image. Tags: jskacel/ee-image
File context/_build/requirements.yml is already up-to-date.
Rewriting Containerfile to capture collection requirements
Running command:
  podman build -f context/Containerfile -t jskacel/ee-image context --no-cache
---
[1/3] STEP 1/7: FROM registry.redhat.io/ansible-automation-platform-22/ee-29-rhel8 AS galaxy
[...]

PR #411, opened 2022-08-24
ansible-builder 3.0 feature branch

nitzmahone:
A 3.0 evolution of `ansible-builder` to address a number of pain points with current `ansible-builder`:

* Allows use of vanilla RHEL-ish container images instead specially-crafted images- check out `demo/v3_demo/execution-environment.yml` for a working sample tested with Fedora, CentOS Stream 8/9, UBI9...
* Allows injection of custom before/after build steps at every build stage.
* Allows copying arbitrary files from the builder host into the generated container build context (which can then be copied into intermediate/final container images as needed with custom build steps).
* Allows declarative bootstrapping of Python, Ansible Core, and Ansible Runner (in addition to roles/collections, of course).
* Allows inline definition of EE Python/bindep/role/collection requirements.
* Allows customization of `ENTRYPOINT` and `CMD`.


Akasurde:
Since we are here, can we think about - 

1. Adding custom scripts specified by users to perform actions which can be before/after galaxy command, before/after specific workflow like start/end of the build - https://github.com/ansible/ansible-builder/issues/293
2. Possibility of specifying Custom CA certificates (required for private galaxy/automation hub) either by copying or mounting volume mounts - https://github.com/ansible/ansible-builder/issues/195

cidrblock:
3 questions while I think of them, 

Can the build def be left in the EE?
Can we produce a 2nd EE with dev tools?
Can the results of a deep introspection of the EE be left in the EE post build (docstrings, python packages, system packages)?


Here is how navigator introspects the image today: https://github.com/ansible/ansible-navigator/blob/main/share/ansible_navigator/utils/image_introspect.py

Collection/plugin are "cached" but here is how they are cataloged, note ansible-doc wasn't used here due to the need to show shadowed collections: https://github.com/ansible/ansible-navigator/blob/main/share/ansible_navigator/utils/catalog_collections.py

nitzmahone:
@cidrblock 

> Can the build def be left in the EE?

eg the source `execution-environment.yml` that builder consumed? Sure, should be pretty easy. I assume no problem there if an EE is used as the base for another EE, that the resultant build def in the final EE would be "last writer wins"- that's technically accurate anyway I guess, since that's what you'd use to re-create that EE... :laughing: 

Need to figure out where to dump it (and the related deep introspection below, assuming we do it). Something like `/etc/ansible/ee-def.yml`?

> Can we produce a 2nd EE with dev tools?

Technically yes, but it seems like a slippery slope. `include_devtools: yes` turns into "I want to install a custom build of the devtools from $private_git_repo that I need to auth to", and generating two copies of the build context, accepting two copies of the tags for `build` and munging them appropriately- I just have trouble seeing how the UX doesn't go south in a big hurry. Also, not sure which all tools you're looking to include, but if navigator is one of them, it's likely going to need to be isolated (eg, installed in a venv or something) to avoid trashing Python deps...


> Can the results of a deep introspection of the EE be left in the EE post build (docstrings, python packages, system packages)?

Yes, I think that would be a reasonable thing, though we'd need to talk through how we'd version that format and where the thing that creates it will live...

nitzmahone:
@Akasurde 

> Adding custom scripts specified by users to perform actions which can be before/after galaxy command, before/after specific workflow like start/end of the build - https://github.com/ansible/ansible-builder/issues/293

Yes, I've been assuming we'd add pre/post hook steps for the Galaxy and Build stages (and maybe more)- I'd probably just do arbitrary steps to replace the existing `additional_build_steps`, but a related need there is to be able to add arbitrary files to the context... eg, `awx-ee` already does something that I can't see how it possibly works: https://github.com/ansible/awx-ee/blob/devel/execution-environment.yml#L14-L15 - `run.sh` is in the root of the project, so builder has no idea it needs to copy that file to the context- it's not parsing the steps (nor should it). Maybe it works because runner already put /run.sh there? Dunno, but having a way to get runner to add files to the context and then specify arbitrary pre/post steps for the different build stages would allow for scripts, commands, even private stages if someone was clever enough (or builder forced the issue by explicitly switching back to the previous stage after executing those hooks).


>Possibility of specifying Custom CA certificates (required for private galaxy/automation hub) either by copying or mounting volume mounts - https://github.com/ansible/ansible-builder/issues/195

I think this could also be accommodated for build-time stuff with the pre/post Galaxy steps I mentioned above- the implementation would likely be highly dependent on knowing where the base OS image hangs its certs, so we probably don't want to provide a high-level interface for doing so, but having the ability to inject extra steps and files into the context should cover that need.

ffirg:
The biggest user experience pain point challenge we get is using builder tooling in a disconnected environment. Can we consider this in the architect up front? We'll need options to be able to NOT pull from the internet for deps etc. 

I'd also like to get any container builds resulting from this put through the RH deptopia mill as this is or is becoming the SoT for container imaging introspection and is currently presenting a gap challenge for productization @Spredzy FYI

ryanmerolle:
I'm happy to test this when ready in a working state.

ansible_builder/_target_scripts/assemble
```diff
@@ -0,0 +1,174 @@
...
+        if [ -z $PKGMGR_OPTS ]; then
+            # NOTE(pabelanger): skip install docs and weak dependencies to
+            # make smaller images. Sadly, setting these in dnf.conf don't
+            # appear to work.
+            PKGMGR_OPTS="--nodocs --setopt install_weak_deps=0"
+        fi
+    fi
+fi
+
+# NOTE(pabelanger): Ensure all the direcotry we use exists regardless
```
Akasurde, line 53:
```suggestion
# NOTE(pabelanger): Ensure all the directory we use exists regardless
```

AlanCoding:
How is this going? It's likely that in the next 1-2 months we'll have to add some hacks to the awx-ee in order to start moving new features into the community that require new Ansible / runner, and @shanemcd has approaches that would work, but we would prefer to work this into it instead.

ryanmerolle:
I hope this will allow the usage of overriding the default yum repos.  The assemble function overwrites all my internal yum proxy details which would likely be an issue for those with air gaped environments.

nitzmahone:
@ryanmerolle yes, it will be possible (and much easier) to do so with explicit pre/post stage tasks for each build stage, but we probably won't add an explicit first-class feature for declarative repo management (starts to duplicate a lot of Ansible itself there :laughing: )

ryanmerolle:
Well I know you guys have been hard at work.  I appreciate the update!

ansible_builder/containerfile.py
```diff
@@ -0,0 +1,365 @@
...
+
+        ######################################################################
+        # Zero stage: prep base image
+        ######################################################################
+
+        # 'base' (possibly customized) will be used by future build stages
+        self.steps.extend([
+            "# Base build stage",
+            "FROM $EE_BASE_IMAGE as base",
+        ])
```
sivel, line 68: Should we include a `USER root` here?  There is no guarantee that the `base_image` will use a default user of `root`.  An example image `registry.access.redhat.com/ubi8/python-39:latest` uses a default non-root user, of `1001`, and the below `dnf` commands fail with:

```
Error: This command has to be run with superuser privileges (under the root user on most systems).
Error: building at STEP "RUN dnf install $PYPKG -y && dnf clean all": while running runtime: exit status 1
```

Shrews: Yep, I think we're going to add that to the base image and just let it propagate to the other images.

Shrews: https://github.com/nitzmahone/ansible-builder/pull/12

PR #415, opened 2022-09-26
Check minimum Python version

### LINKED ISSUE #407: restrict min Python version when installing
According to #379 and #381, ansible-builder 1.1.0 needs Python 3.8  or higher.

Not only documentation, I would like to restrict when installing.

Like this:  https://github.com/ansible/ansible/blob/v2.13.2/setup.cfg#L39

Akasurde:
##### SUMMARY

* Ansible builder supports Python 3.8 and above. Added a check
  for minimum Python version checking.

Fixes: #407

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,4 +5,5 @@ from setuptools import setup
 setup(
     setup_requires=['pbr'],
     pbr=True,
+    python_requires=">=3.8",
 )


Shrews:
Could we not just use `python_requires`, discussed [here](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires)?

Akasurde:
@Shrews Can you please check the new change? Thanks

Shrews (approved):
Looks fine, but seems a new release of `yamllint` version 1.29.0 is causing linting errors. See PR https://github.com/ansible/ansible-builder/pull/440 for the fix.

PR #460, opened 2023-03-29
Normalize package names with lowercase

### LINKED ISSUE #408: ERROR: Double requirement given: pyvmomi
### issue

ansible-builder fails with this error.

```
+ pip3 install --cache-dir=/output/wheels -r /tmp/src/requirements.txt
WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3 install --user` instead.
ERROR: Double requirement given: pyvmomi (from -r /tmp/src/requirements.txt (line 5)) (already in pyVmomi>=6.7 (from -r /tmp/src/requirements.txt (line 1)), name='pyvmomi')
Error: error building at STEP "RUN assemble": error while running runtime: exit status 1

An error occured (rc=1), see output line(s) above for details.
```

references: #45, #351

### environment

```
# cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.6 (Ootpa)

# dnf list installed | grep ansible-builder
ansible-builder.noarch                        1.1.0-3.el8ap                               @ansible-automation-platform-2.2-for-rhel-8-x86_64-rpms

# ansible-builder --version
1.1.0

# podman --version
podman version 4.1.1
```

### step to reproduce

I created definitions files as below.

1. execution-environment.yml 

```
---
build_arg_defaults:
  EE_BASE_IMAGE: 'registry.redhat.io/ansible-automation-platform-22/ee-29-rhel8:latest'

dependencies:
  galaxy: requirements.yml
  python: requirements.txt

additional_build_steps:
```

2. requirements.txt

```
ansible==2.9.27
psutil
pyvmomi
```

3. requirements.yml

```
---
collections:
- name: community.vmware
```

After that, I downloaded the images in advance.

```
# podman images
REPOSITORY                                                               TAG         IMAGE ID      CREATED        SIZE
registry.redhat.io/ansible-automation-platform-22/ee-supported-rhel8     latest      e992a75cc062  8 days ago     1.45 GB
registry.redhat.io/ansible-automation-platform-22/ee-minimal-rhel8       latest      0ba765fbe331  8 days ago     286 MB
registry.redhat.io/ansible-automation-platform-22/ansible-builder-rhel8  latest      5f854afaaaa4  8 days ago     311 MB
registry.redhat.io/ansible-automation-platform-22/ee-29-rhel8            latest      45fa22cc1c56  8 days ago     805 MB
```

I performed the ansible-builder command.

```
# ansible-builder build -f execution-environment.yml -v 3 -t localhost/my-custom-ee-29
```

I will attach the whole output.

### workaround

When I commented out the line in requirements.txt, it worked as expected.

```
# cat requirements.txt 
ansible==2.9.27
psutil
#pyvmomi
```

I guess when the `pyvmomi` is already installed in the EE_BASE_IMAGE, it seems the builder (introspect?) didn't calculate the duplication in the procedure.

### background

My customer wants to migrate the virtual environments from Ansible Tower 3.8 to Ansible Automation Platform 2.2 according to this step described in this product documentation.

https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.2/html/red_hat_ansible_automation_platform_upgrade_and_migration_guide/upgrading-to-ees

However the output of `awx-manage export_custom_venv /path/to/venv` cannot be applied to `ansible-builder` command like this issue, it makes confused.

dsavineau:
When a python package is listed multiple time but with uppercase/lowercase changes then introspect sanitization won't handle it. In db9e74a, only the underscore/dash scenario has been solved. As a result, the assemble script will fail with a double requirement error.

```console
ERROR: Double requirement given: Jinja2==3.0.3 (from -r /tmp/src/requirements.txt (line 40)) (already in jinja2>=2.8 (from -r /tmp/src/requirements.txt (line 12)), name='Jinja2')
```

While this isn't an issue for recent pip version, older versions with python 3.8 and 3.9 runtimes on RHEL 8 based distributions are affected by this.

This replaces the setuptools runtime usage by importlib.metadata which provides builtin method to normalize package names.

Unlike pkg_resources.safe_name, the dash character is changed to underscore.

Closes: #408

https://packaging.python.org/en/latest/specifications/name-normalization

### DIFF
--- a/ansible_builder/_target_scripts/introspect.py
+++ b/ansible_builder/_target_scripts/introspect.py
@@ -5,7 +5,7 @@ import sys
 import yaml
 
 import requirements
-from pkg_resources import safe_name
+import importlib.metadata
 
 base_collections_path = '/usr/share/ansible/collections'
 default_file = 'execution-environment.yml'
@@ -330,7 +330,7 @@ def sanitize_requirements(collection_py_reqs):
         try:
             for req in requirements.parse('\n'.join(lines)):
                 if req.specifier:
-                    req.name = safe_name(req.name)
+                    req.name = importlib.metadata.Prepared(req.name).normalized
                 req.collections = [collection]  # add backref for later
                 if req.name is None:
                     consolidated.append(req)
--- a/ansible_builder/requirements.py
+++ b/ansible_builder/requirements.py
@@ -1,6 +1,6 @@
 import logging
 import requirements
-from pkg_resources import safe_name
+import importlib.metadata
 
 
 logger = logging.getLogger(__name__)
@@ -39,7 +39,7 @@ def sanitize_requirements(collection_py_reqs):
         try:
             for req in requirements.parse('\n'.join(lines)):
                 if req.specifier:
-                    req.name = safe_name(req.name)
+                    req.name = importlib.metadata.Prepared(req.name).normalized
                 req.collections = [collection]  # add backref for later
                 if req.name is None:
                     consolidated.append(req)
--- a/test/data/ansible_collections/test/reqfile/requirements.txt
+++ b/test/data/ansible_collections/test/reqfile/requirements.txt
@@ -1,3 +1,4 @@
 pytz
 python-dateutil>=2.8.2    # intentional dash
+jinja2>=3.0    # intentional lowercase
 -r extra_req.txt
--- a/ansible_builder/cli.py
+++ b/ansible_builder/cli.py
@@ -2,7 +2,7 @@ import argparse
 import logging
 import sys
 import os
-import pkg_resources
+import importlib.metadata
 
 from . import constants
 
@@ -43,7 +43,7 @@ def run():
 
 
 def get_version():
-    return pkg_resources.get_distribution('ansible_builder').version
+    return importlib.metadata.version('ansible_builder')
 
 
 def add_container_options(parser):
--- a/test/data/ansible_collections/other/reqfile/requirements.txt
+++ b/test/data/ansible_collections/other/reqfile/requirements.txt
@@ -1,1 +1,2 @@
 python_dateutil    # intentional underscore
+Jinja2    # intentional uppercase
--- a/test/integration/test_introspect_cli.py
+++ b/test/integration/test_introspect_cli.py
@@ -36,6 +36,7 @@ def test_introspect_write_python(cli, data_dir, tmp_path):
         'pyvcloud>=14  # from collection test.metadata',
         'pytz  # from collection test.reqfile',
         'python-dateutil>=2.8.2  # from collection test.reqfile',
+        'jinja2>=3.0  # from collection test.reqfile',
         'tacacs_plus  # from collection test.reqfile',
         'pyvcloud>=18.0.10  # from collection test.reqfile',
         '',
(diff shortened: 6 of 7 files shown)


ansible_builder/requirements.py
```diff
@@ -39,7 +39,7 @@ def sanitize_requirements(collection_py_reqs):
         try:
             for req in requirements.parse('\n'.join(lines)):
                 if req.specifier:
-                    req.name = safe_name(req.name)
+                    req.name = safe_name(req.name).lower()
```
nitzmahone: Fix LGTM, though while we're "in the neighborhood", we should probably either switch to `Prepared.normalize()` from `importlib.metadata` (which handles the character replacements as well as the case internally) or just embed [its impl](https://github.com/python/cpython/blob/main/Lib/importlib/metadata/__init__.py#L691-L695) directly, since `Prepared` doesn't seem to be documented :frowning_face:. `pkg_resources` is deprecated, and part of `setuptools`, which IIRC will not be installed by default anymore in new virtualenvs and other places, so we can't necessarily rely on its presence (nor should we declare a runtime dep on it).

test/data/ansible_collections/test/reqfile/requirements.txt
```diff
@@ -1,3 +1,4 @@
 pytz
 python-dateutil>=2.8.2    # intentional dash
+jinja2>=3.0    # intential lowercase
```
nitzmahone:
```suggestion
jinja2>=3.0    # intentional lowercase
```

nitzmahone (changes requested):
Just need to decide what to do about the actual normalizing impl we're using- @shrews / @Akasurde, any preference there?

ansible_builder/requirements.py
Akasurde: I agree, let us switch to `importlib.metadata` just like https://github.com/ansible/ansible-runner/pull/1224

ansible_builder/_target_scripts/introspect.py
```diff
@@ -5,7 +5,7 @@
 import yaml
 
 import requirements
-from pkg_resources import safe_name
+import importlib.metadata
```
Akasurde, line 8: Can you please create a backward-compatible shim just like https://github.com/ansible/ansible-runner/pull/1224/files#diff-fe262d49ef98e3d8bda61b35f5546ef3f425fc52cfcb301be8e204a0b08d5707R1-R6?

Akasurde (changes requested):

ansible_builder/_target_scripts/introspect.py
nitzmahone: Can't do it that way (without a lot of pain anyway)- `introspect` gets copied into the target container and run standalone during the build, so it won't be able to access any other builder guts. It'll just need to do the import dance inline. (this is one of the big changes for 3.0 that no longer requires builder itself to be installed in the builder container)

nitzmahone: Thinking about it a little more, so long as the target EE container Python is >= 3.9, what's already here should work fine. If we want to be able to build with <3.9, we should probably just inline the PEP503 + `_` normalization ala https://github.com/python/cpython/blob/3.10/Lib/importlib/metadata/__init__.py#L866-L871

dsavineau: I don't think we have to deal with `importlib_metadata` here (unlike runner) because the Prepared class is already part of stdlib even back in 3.8
This would allows a safer backport to 1.2.x (needed by AAP 2.3) rather than adding a new runtime dependency.

https://github.com/python/cpython/blob/3.9/Lib/importlib/metadata.py#L473-L491
https://github.com/python/cpython/blob/3.8/Lib/importlib/metadata.py#L434-L452

nitzmahone (approved):

Akasurde (approved):

dsavineau:
@nitzmahone @Akasurde Do you agree to backport this to 3.0 and 1.2 branches so current AAP 2.3 release and next AAP 2.4 will be fixed ?

PR #464, opened 2023-04-04
Change ansible_core/ansible_runner schema types

Shrews:
Changes `dependencies.ansible_core` and `dependencies.ansible_runner` from string type to a dict. This dict can/will contain keys specific to different installation types. Currently, only pip installation (`package_pip` subkey) is supported.


nitzmahone (commented):
Yep, this all looks good, but immediately reminded me that `package_name` for the Python package dep is now inconsistent with `package_pip`... So do we name `package_os`, `package_name_os`, or ... ?

nitzmahone (approved):

Issue #211, opened 2021-04-30
Support PIP_OPTs for python-builder-image


pabelanger:
We added support to python-builder-image, which allows a user to pass pip options via PIP_OPTs:

  https://github.com/ansible/python-builder-image/pull/34
  
In the case of downstream builds, we set --no-build-isolation for reproducible builds. Additionally, it would allow upstream users to pass:

  --no-binary :all:

Which disabled use of pre-compiled wheels, again getting upstream builds closer to downstream / offline supported.

AlanCoding:
How would users pass this option to ansible-builder? Would it be in the execution environment definition?

pabelanger:
I imagined it working the same way we did the galaxy opts setting

AlanCoding:
so a build arg.

The assemble script also takes, in some sense, CONSTRAINTS, PACKAGES, and others. I don't fully understand those options.

In my imagination, this would follow an `ARG` and then `ENV` pattern in the Containerfile. That wouldn't get messy if we add more later.

tanganellilore:
@AlanCoding @Shrews I created a PR that will add this param. In the PR I wrote why we need this tipe of arg

Thanks

Shrews:
I'm going to go ahead and close this. We just merged support for a new EE format that has features that would allow you highly customize any of the builder build phases (e.g., the new [`additional_build_steps` ](https://ansible-builder.readthedocs.io/en/latest/definition/#additional-build-steps) section).

PR #465, opened 2023-04-05
Add PIP_OPTS on build_arg_defaults

tanganellilore:
Team,

this pull request add support to `PIP_OPTS ` on `build_arg_defaults`, as suggested in this #211.

So with this implementation we can fix it and allow us to use `PIP_OPTS ` option.

One things related why we need this.
I notice, as the same in the AWX matrix channel shared, that after upgrade of PIP to >21 the resolve dependecies is very slow and pip try to download each package release for each package passed on `requirements.txt`.

One of the workaround at this moment is to use the param `--use-deprecated=legacy-resolver` that will use "old style" of repo dependecies.

Other way to resolve this is to put a release of every package passed (that can be under our control) and all package requirement of each collection, but we need manually resolve dependecies.

In any case this is an option present in `assemble` that can be usefull for everybody tht requires special opts on pip.

### DIFF
--- a/ansible_builder/constants.py
+++ b/ansible_builder/constants.py
@@ -16,7 +16,8 @@ build_arg_defaults = dict(
     ANSIBLE_GALAXY_CLI_COLLECTION_OPTS='',
     ANSIBLE_GALAXY_CLI_ROLE_OPTS='',
     EE_BASE_IMAGE='quay.io/ansible/ansible-runner:latest',
-    EE_BUILDER_IMAGE='quay.io/ansible/ansible-builder:latest'
+    EE_BUILDER_IMAGE='quay.io/ansible/ansible-builder:latest',
+    PIP_OPTS=''
 )
 
 user_content_subfolder = '_build'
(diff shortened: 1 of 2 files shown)


Shrews:
Thanks for the PR!

However, the new v3 EE format that we just merged into the `devel` branch will support this type of feature using the new `additional_build_steps` keyword where you can set any `ENV` or `ARG` value you like before each build phase (see the [docs for that here](https://ansible-builder.readthedocs.io/en/latest/definition/#additional-build-steps)). We expect a beta release of the new builder soon.

I don't see us adding new features for the older EE formats without a very good reason, so my suggestion is to move to the new v3 format if you want to do this.

Shrews:
Closing per above comment.

tanganellilore:
@Shrews clear, I just use old docs and old release. Do you know where is planned v3 release on pip?
We can found somewhere a roadmap or similar?

Thanks

Shrews:
@tanganellilore We hope to have a beta release very soon, with an official release coming within the next few weeks.

tanganellilore:
@Shrews i tried to use new release, and i notice that `install-from-bindep` not have `PIP_OPTS` implemented...and this can be an issue...

PR #468, opened 2023-04-06
Add pipopts on install-from-bindep

tanganellilore:
Team,

this pull request add support to `PIP_OPTS  ` on install-from-bindep, as same of `assemble `file.

So with this implementation we can add it and allow us to use PIP_OPTS  option with new v3 format.

### DIFF
--- a/ansible_builder/_target_scripts/install-from-bindep
+++ b/ansible_builder/_target_scripts/install-from-bindep
@@ -22,6 +22,7 @@ PKGMGR_OPTS="${PKGMGR_OPTS:-}"
 
 PYCMD="${PYCMD:=/usr/bin/python3}"
 PIPCMD="${PIPCMD:=$PYCMD -m pip}"
+PIP_OPTS="${PIP_OPTS-}"
 
 $PYCMD -m ensurepip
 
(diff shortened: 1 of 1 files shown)


tanganellilore:
@Shrews this is reasonable?

Shrews:
> @Shrews this is reasonable?

It does seem reasonable. The thing we need to put some thought into is tests. We only just incorporated these scripts into this repo from a separate repo (they would normally be expected to exist in the base image, but we get rid of that requirement in this release). Any changes to these scripts, just like the rest of the code, must include a test to validate the change. Unfortunately for you, this is the first PR to one of these scripts, and we do not yet have a testing strategy in place. If you have a good idea for testing this change, I encourage you to add it to the PR and let us evaluate it. Otherwise, the team will need to find time to put our heads together on a testing strategy.

Shrews:
@tanganellilore Is this change even really necessary? I understand wanting to use PIP_OPTS in `assemble`, but builder's use of `install-from-bindep` script is really just to install the wheel packages that were built by `assemble`. Do we actually gain anything by adding PIP_OPTS support here?

tanganellilore:
Hi team,
sorry for delay but i was focused on other stuff and other issue on awx project.
What your ad saying @Shrews should be  correct, in the assemble we should have all packages that is requested in final step, but this seems not happen in my test env build.

As you can see on [this pastebin](https://pastebin.com/sy4ztn2b), after some library that is fetched from "cache" (the output dir of assemble), some new library are requested and pip try to loop on each library to find the correct one.

I try to find why this happen, but anyway, adding an option parameter, not brake anything and increase the possibility to customize pip installation.

Thanks

tanganellilore:
Hi team,
one more point.
After some discussion with @shanemcd on matrix, I follow him suggestion to upgrade pip in base imge, and seems to work without pipopts in the final stage.

Record log is here: [pastebin](https://pastebin.com/dCUp9aB3)

I'm still using `PIP_OPTS=--use-deprecated=legacy-resolver` on other layers

nitzmahone (approved):

Shrews (approved):

PR #477, opened 2023-04-17
Update 1.2 CI testing

Shrews:
- Removes use of the test container
- Updates codecov portion to use GH Action
- Use quay.io containers for integration tests instead of building our own
- Fixes to `test_build_command` test


Akasurde (approved):
LGTM

PR #479, opened 2023-04-17
Pass full path while copying

### LINKED ISSUE #476: de-dupe modification warnings
Multiple updates on directory-based copies cause repeated warning messages- consider de-duping these messages and/or increasing their verbosity. eg:
```
(tmp-736775c7cf96623) [mdavis@mdavis-p1 ee-awx-derived]$ ansible-builder build
File context/_build/scripts had modifications and will be rewritten
File context/_build/scripts had modifications and will be rewritten
File context/_build/scripts had modifications and will be rewritten
File context/_build/scripts had modifications and will be rewritten
File context/_build/scripts had modifications and will be rewritten
File context/_build/scripts had modifications and will be rewritten
Running command:
  podman build -f context/Containerfile -t ansible-execution-env:latest --squash context
```

Akasurde:
copy_file API compares and copies files, passing file's full path while copying.

Fixes: #476

### DIFF
--- a/test/unit/test_utils.py
+++ b/test/unit/test_utils.py
@@ -55,6 +55,21 @@ def test_copy_touched_file(dest_file, source_file):
     assert not copy_file(source_file, dest_file)
 
 
+def test_copy_file_with_destination_directory(dest_file, source_file):
+    # Change source file to trigger copy_file
+    source_file.write_text('foo\nbar\nzoo')
+
+    with pytest.raises(Exception) as err:
+        copy_file(source_file, '/tmp')
+    assert "can not be a directory" in str(err.value.args[0])
+
+    with pytest.raises(Exception) as err:
+        copy_file('/tmp', dest_file)
+    assert "can not be a directory" in str(err.value.args[0])
+
+    assert copy_file(source_file, dest_file)
+
+
 @pytest.mark.run_command
 def test_failed_command(mocker):
     mocker.patch('ansible_builder.utils.subprocess.Popen.wait', return_value=1)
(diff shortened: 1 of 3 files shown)


Shrews (commented):
This change is actually fixing a bug where `copy_file()` was being passed a file for the source and a directory for the destination. That method is not designed to handle that, so I'm thinking maybe that should be a programming error we catch (maybe raise an Exception if src and dest are not of the same type?)

Akasurde:
> This change is actually fixing a bug where `copy_file()` was being passed a file for the source and a directory for the destination. That method is not designed to handle that, so I'm thinking maybe that should be a programming error we catch (maybe raise an Exception if src and dest are not of the same type?)

Make sense to me. Let me work on it.

Shrews (changes requested):
Ideally, I'd like to see a hard error here to catch a progamming error (e.g., throw an exception for a hard stop). If someone adds a new use of `copy_file()`, this isn't going to do anything to let them know they are using it incorrectly unless they are carefully watching logging.

Also, we need to valid that both parameters are not directories, not just `dest`.

test/unit/test_utils.py
```diff
@@ -55,6 +55,19 @@ def test_copy_touched_file(dest_file, source_file):
...
+def test_copy_file_with_destination_directory(dest_file, source_file):
+    # Change source file to trigger copy_file
+    source_file.write_text('foo\nbar\nzoo')
+
+    with pytest.raises(Exception):
+        copy_file(source_file, '/tmp')
+
+    with pytest.raises(Exception):
+        copy_file('/tmp', dest_file)
+
```
Shrews, line 69: Looks great! But since Exception is pretty broad, let's add an assert to make sure we are getting the Exception message we expect. If you capture the exception in a var (say, `error`), you can check the message, like:

```
assert "can not be a directory" in str(error.value.args[0])
```

Shrews (approved):

PR #483, opened 2023-04-19
Include data files from _target_scripts

### LINKED ISSUE #482: Package data files are not included when installed with source files without .git
The bash files added to `_target_scripts` directory in #411 don't get included when installed from source with git info stripped out (we set `PBR_VERSION=xxx` during build) as pbr won't be able to generate the file list.

This can be reproduced by running `PBR_VERSION=xxx python3 setup.py build` without .git directory.

Akasurde:
pbr ignores the data files when .git directory is missing from the source. Use `MANIFEST.in` to include data files from _target_scripts directory while building source distribution.

Fixes: #482

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include ansible_builder/_target_scripts/*


sivel (approved):
WFM

PR #484, opened 2023-04-20
Fix shell script syntax

Akasurde:
* Quote the values to prevent globbing and word splitting

### DIFF
--- a/ansible_builder/_target_scripts/assemble
+++ b/ansible_builder/_target_scripts/assemble
@@ -20,7 +20,7 @@
 # at build time.
 set -ex
 
-RELEASE=$(source /etc/os-release; echo $ID)
+RELEASE=$(source /etc/os-release; echo "${ID}")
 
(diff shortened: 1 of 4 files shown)


ansible_builder/_target_scripts/assemble
```diff
@@ -106,22 +106,22 @@ function install_wheels {
...
     fi
 
     # Install each of the extras so that we collect all possibly
     # needed wheels in the wheel cache. get-extras-packages also
     # writes out the req files into /output/$extra/requirements.txt.
 
     # FIXME: this doesn't error out the build when it fails, yay
-    for req in $(/build/get-extras-packages) ; do
-        $PIPCMD install $CONSTRAINTS $PIP_OPTS --cache-dir=/output/wheels "$req"
+    for req in $(/output/scripts/get-extras-packages) ; do
```
Akasurde, line 123: This path was wrong. `get-extras-packages` is available in `/output/scripts`.

PR #485, opened 2023-04-20
Fix path for in assemble

Akasurde:
* Fixed path for get-extra-packages script


Akasurde:
Before this patch - 

```
++ /build/get-extras-packages
/output/scripts/assemble: line 123: /build/get-extras-packages: No such file or directory
+ /usr/bin/dnf clean all
```

Shrews:
I think we can just remove this script. It doesn't do anything for builder.

Akasurde:
@Shrews I removed the `get-extras-packages` via https://github.com/ansible/ansible-builder/pull/492.

PR #486, opened 2023-04-22
entrypoint script support, passwd permission relax

### LINKED ISSUE #463: Determine default upstream/downstream USER/WORKDIR/homedir behavior
Currently, builder 3.0 is defaulting to `USER root`, and no `WORKDIR` (which ends up being `/`) - this causes various issues when the container is run as an arbitrary non-existent non-root user under some container runtimes. The ansible-runner container base images (and some downstream derivatives) used various `ENTRYPOINT` script magic to paper over these issues, but at the cost of hardcoded/broken homedirs and a number of other issues. The need for these tweaks has significantly decreased now that runner no longer requires knowledge of the container user's homedir.

I'm guessing we're going to at least need to set `WORKDIR` to something like `/ansible-ee` or whatever that's `root/root` and `rwx` for the `root` group to accommodate container runtimes that want to run it as arbitrary bogus UIDs. Permissions on `/` set at build-time don't appear to be preserved in the final container image, at least under podman, so with the default `WORKDIR` of `/`, runtime-injected users that use `WORKDIR` as `HOME` are non-functional.

Need to get an actual documented set of requirements on this one to be sure we can cover all the bases.

nitzmahone:
fixes #463 

* use a boilerplate entrypoint script by default to ensure usable and correct HOME semantics for all users
* optional permisssion relaxtion on /etc/passwd (default to true) for compatibility with old/broken container runtimes that don't set it properly for ephemeral users

### DIFF
--- a/ansible_builder/containerfile.py
+++ b/ansible_builder/containerfile.py
@@ -156,6 +156,12 @@ class Containerfile:
         self._prepare_galaxy_copy_steps()
         self._prepare_system_runtime_deps_steps()
 
+        if self.definition.version >= 3 and self.definition.options['relax_passwd_permissions']:
+            self._relax_etc_passwd_permissions()
+
+        if self.definition.version >= 3 and (final_workdir := self.definition.options['workdir']):
+            self._prepare_final_workdir(final_workdir)
+
         # install init package if specified
         # FUTURE: could move this into the pre-install wheel phase
         if init_pip_pkg := self.definition.container_init.get('package_pip'):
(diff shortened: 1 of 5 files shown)


ansible_builder/containerfile.py
```diff
@@ -307,6 +313,24 @@ def _insert_custom_steps(self, section: str):
                     lines = section_steps
                 self.steps.extend(lines)
 
+    def _relax_etc_passwd_permissions(self):
+        if self.definition.version < 3:
```
Shrews: This check is unnecessary (done at the call point).

Shrews (dismissed):
Doc updates, please.

PR #487, opened 2023-04-24
Add support for Package manager options

### LINKED ISSUE #418: Option to use additional PKGMGR_OPTS through ansible-builder
Feature request - Option to use PKGMGR_OPTS with Ansible Builder to disable repository using microdnf.
While creating a custom image there is no option to pass microdnf parameters.

Akasurde:
* added support for providing additional package manager options

Fixes: #418

### DIFF
--- a/ansible_builder/containerfile.py
+++ b/ansible_builder/containerfile.py
@@ -196,6 +196,7 @@ class Containerfile:
 
         if self.definition.version >= 3:
             global_args['PKGMGR'] = self.definition.options['package_manager_path']
+            global_args['PKGMGR_OPTS'] = self.definition.options['package_manager_options']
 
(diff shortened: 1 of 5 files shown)


Shrews:
This change shouldn't be necessary. You could just add this in the new v3 EE format:

```
additional_build_steps:
   prepend_builder: ENV PKMGR_OPTS="--whatever"
```

We have the option to set PKMGR in the EE because that is used directly by the code building the Containerfile. That code cannot programmatically deteremine which pkg manager is present, unlike the `assemble` script, which can run inside the container. I don't think we should try to expose the other script env options with an EE equivalent.

Akasurde:
Makes sense. I will go ahead and close this PR then.

PR #503, opened 2023-04-27
Change squash default from 'new' to 'off'

Shrews:


nitzmahone (approved):
LGTM, :crossed_fingers: the tests pass - thanks! 

Does make me realize we don't have a good place to communicate breaking changes though, which this technically is...

PR #505, opened 2023-04-28
Improve test runtime detection

Shrews:
This changes runtime detection (by shelling out to call `which`)  from happening for every test to only once per test session/worker.


Shrews:
Closing. This change is included in #508.

PR #509, opened 2023-05-01
detect `podman-docker` and skip `docker` runtime tests

nitzmahone:
* only detects runtimes once, but still injects static `skipif` entries to surface the fact that we're not testing everything we could
* works for most configurations, but fails to detect:
  - docker client against podman socket (incorrectly detected as `podman`)
  - `podman-docker` v3.x with the default docker client emulation warning disabled


Akasurde:
superseded by https://github.com/ansible/ansible-builder/pull/508

PR #515, opened 2023-05-03
support blind-translated YAML mapping for galaxy deps

nitzmahone:
* simplify docs accordingly

### DIFF
--- a/demo/execution-environment.yml
+++ b/demo/execution-environment.yml
@@ -13,25 +13,26 @@ dependencies:
     package_system: python39  # (optional) name of a Python interpreter OS package to install
     python_path: /usr/bin/python3.9  # (optional) path to the Python interpreter to use
 
-  ansible_core:
+  ansible_core:  # (optional) install ansible-core if the base image doesn't already contain it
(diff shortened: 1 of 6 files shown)


nitzmahone:
If the "blind-translated" in the title wasn't enough, we should probably be explicit somewhere in the docs about the fact that by design, we do *zero* validation of the structure of this YAML beyond requiring that it's a dictionary- we just splat it into a `requirements.yml` in the root of the build context.

We don't builder to be tightly-coupled to Galaxy CLI's schema, so the fact that it's "YAML-in-YAML" is merely a convenience- we shouldn't rely on this going forward (eg, even though inlining things like ignores here might be *really* easy and tempting, we should probably resist the urge since it would require us to more fully understand and rewrite the contents of the `galaxy` dict)...

Shrews (changes requested):
If this is going to be the preferred format, we should change `demo/execution-environment.yml` too.

cidrblock:
TY.  Even without validation against the galaxy expected syntax, the simple fact this is now a structured list will help people spot issues.

Shrews (approved):

PR #516, opened 2023-05-03
Remove forced `dnf upgrade`, speed up tests

nitzmahone:
supersedes #503 (included here)

* removes the forced (and often very slow) `dnf upgrade` that occurred in several stages
* speeds up integration tests by enabling build cache usage for most, and converting to lighter-weight EE v3 definitions
* new optional build arg `PKGMGR_PRESERVE_CACHE`:
  * default none- clears package manager cache frequently
  * any value but `always` zaps the cache only at the end of `install-from-bindep` (good for CI, squashed images)
  * `always` never clears the package manager cache (good for CI, testing base images where a pre-populated package cache is handy)


test/conftest.py
```diff
@@ -226,15 +225,17 @@ def delete_image(runtime, image_name):
...
 
 
 @pytest.fixture
 @pytest.mark.test_all_runtimes
 def ee_tag(request, runtime):
     image_name = gen_image_name(request)
     yield image_name
-    delete_image(runtime, image_name)
+    # FIXME: defer to end?
+    # delete_image(runtime, image_name)
```
Shrews, line 238: I don't know that I'm really onboard with this change. Why is it necessary? Without this, it leaves 30-some images around during local testing that have to be manually cleaned up.

Out of curiosity, I compared testing time with these deletes commented out, and re-including them. My test actually ran _faster_ with the deletes added back, although two tests did fail:

```
test/integration/test_build.py::test_base_image_build_arg[podman]
test/integration/test_build.py::test_collection_system_requirement[podman]
```

FWIW, I didn't test docker.

test/data/prepend_steps/execution-environment.yml
```diff
@@ -1,5 +1,14 @@
...
+
+images:
+  base_image:
+    name: quay.io/centos/centos:stream9
+
 additional_build_steps:
-  prepend: |
-    RUN whoami
+  prepend_base: |
+    RUN echo "hello from prepend_base" > /baseout.txt
```
Shrews: If we do not intend to test v1 `prepend` anymore (we'd be losing coverage if we change this), then we don't need this test data (or the test that uses it) anymore since all of the prepend/append steps are tested in `test/data/v3/pre_and_post/ee.yml`.

test/integration/test_build.py
```diff
@@ -167,32 +166,46 @@ def test_base_image_build_arg(cli, runtime, ee_tag, tmp_path, data_dir):
...
+    no_cache_result = cli(build_cmd + ' --no-cache')
 
-    build_cmd = f'ansible-builder build -c {tmp_path} -f {ee_def} -t {ee_tag} --container-runtime {runtime} -v 3 --squash off'
-    cli(build_cmd + ' --no-cache')
-    result = cli(build_cmd)
+    pass1_containerfile = (tmp_path / containerfile_name).read_text()
+
+    print(f"no_cache_result stdout: \n{no_cache_result.stdout}")
+    print(f"no_cache_result stderr: \n{no_cache_result.stderr}")
+    print(f"containerfile contents for no-cache: \n{pass1_containerfile}")
```
Shrews: Probably want to remove these `print` calls?

test/pulp_integration/test_policies.py
```diff
@@ -39,6 +39,8 @@ def setup_class(cls):
             target = Path(BACKUP_POLICY_PATH).expanduser()
             user_policy.rename(target)
 
+            # cmd = 'podman image prune --force'
+            # subprocess.run(cmd.split())
```
Shrews: Local testing leftovers?

test/integration/test_build.py
```diff
@@ -98,9 +98,11 @@ def test_user_python_requirement(cli, runtime, ee_tag, tmp_path, data_dir):
 
 @pytest.mark.test_all_runtimes
 def test_python_git_requirement(cli, runtime, ee_tag, tmp_path, data_dir):
+    # FIXME: is this test even necessary, since we don't (currently) special-case git in builder?
+
```
Shrews: I think you're right and we don't need this test anymore. The older releases have `git` in the `bindep.txt` file for the builder repo (and we still do, actually). Since we aren't building and publishing images anymore, this is a pointless test. Let's remove it (and the test data).

test/conftest.py
nitzmahone: Yeah, this fixme definitely needs to be resolved- the problem is that doing the image deletions inline can break the build cache, as IIUC at least some versions of podman will try to clear dangling intermediate layers as part of the deletion, but when multiple builds are occurring in parallel, it can nuke entries that are being used.

I *think* the right way to fix this will be to defer the deletions to the session teardown. IIRC there's only one session under pytest-xdist- if not, we'll have to figure something else out...

And yeah, the fact that it ran faster when you re-enabled this is because one of the longest-running tests failed early, likely because of what I mentioned above. We can definitely clean them up, but doing them inline will almost definitely introduce instability as the test concurrency increases.

test/data/prepend_steps/execution-environment.yml
nitzmahone: I think we can probably move most of the v1 tests like that to unit tests and/or create tests- v1 builds are by definition going to be *much* more expensive, and the value of a full build integration test that ensures that a line appears in the output probably isn't worth the time spent on it.

test/conftest.py
Shrews: Gotcha. We gotta do what we gotta do then. We can noodle another way to do the cleanup at a later point.

test/data/prepend_steps/execution-environment.yml
nitzmahone: (moved to a `create` test and killed the test data)

test/integration/test_build.py
nitzmahone: removed

Shrews (approved):

test/conftest.py
nitzmahone: Made a new issue to address in #519

PR #517, opened 2023-05-03
Parse build-args for empty values

### LINKED ISSUE #514: empty `build-arg` passthru is broken
The custom `--build-arg` splitter logic in the builder CLI doesn't properly preserve args with an empty value; they are conflated with args without a value. eg:

```
ansible-builder build --build-arg PKGMGR_PRESERVE_CACHE=
```

should force an empty value, but since the trailing `=` is dropped by the splitter logic (`BuildArgAction` in `cli.py`), once passed to the underlying build CLI, it leaves the existing value (if any) or enables it to come from the environment.

Need to update the splitter logic to support preserving empty strings, as well as only splitting on the first `=`.

### LINKED ISSUE #520: `--build-arg` eats embedded `=` in args
Sometimes it's necessary to embed `=` in the value side of a `build-arg`; the current `BuildArgAction` argparse action ignores everything after the first `=`. eg:

```
ansible-builder --build-arg 'PYTHON_CONFIG_SETTINGS=--config-setting=foo=bar'
```

`BuildArgAction` uses an unbounded `split`, but discards everything but the first item after the first `=`, so the case above gets passed to the builder as `--build-arg PYTHON_CONFIG_SETTINGS=--config-setting`.

It should probably just be using `partition` instead and passing along everything in the final element.

Akasurde:
* Split for first equal sign in build-args
* Honor empty values in build-args

fixes #514, fixes #520

### DIFF
--- a/src/ansible_builder/main.py
+++ b/src/ansible_builder/main.py
@@ -154,7 +154,9 @@ class AnsibleBuilder:
             command.extend(["-t", tag])
 
         for key, value in self.build_args.items():
-            if value:
+            if value == '':
+                build_arg = f"--build-arg={key}=''"
+            elif value:
                 build_arg = f"--build-arg={key}={value}"
             else:
                 build_arg = f"--build-arg={key}"
--- a/src/ansible_builder/main.py
+++ b/src/ansible_builder/main.py
@@ -154,9 +154,7 @@ class AnsibleBuilder:
             command.extend(["-t", tag])
 
         for key, value in self.build_args.items():
-            if value == '':
-                build_arg = f"--build-arg={key}=''"
-            elif value:
+            if value:
                 build_arg = f"--build-arg={key}={value}"
             else:
                 build_arg = f"--build-arg={key}"
--- a/test/integration/test_build.py
+++ b/test/integration/test_build.py
@@ -107,6 +107,16 @@ def test_build_args_basic(cli, runtime, ee_tag, tmp_path, data_dir):
 
 
 @pytest.mark.test_all_runtimes
+def test_build_args_empty_value(cli, runtime, ee_tag, tmp_path, data_dir):
+    bc = tmp_path
+    ee_def = data_dir / 'build_args' / 'execution-environment.yml'
+    result = cli(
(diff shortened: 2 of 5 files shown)


src/ansible_builder/main.py
```diff
@@ -154,7 +154,9 @@ def build_command(self):
             command.extend(["-t", tag])
 
         for key, value in self.build_args.items():
-            if value:
+            if value == '':
```
nitzmahone: This logic should be wrapped up inside the `BuildArgAction` below, and probably just use `partition` to split the string- if the last element has no value but the second element is `=`, you know you need to tack an `=` onto the end of the key.

Also, as a general rule, embedding empty quotes in something that's going to get round-tripped to another CLI is probably going to cause a lot more problems than it solves; just having a trailing `=` on the arg should be sufficient in this case.

test/integration/test_build.py
```diff
@@ -135,6 +135,16 @@ def test_build_args_basic(cli, runtime, ee_tag, tmp_path, data_dir):
     assert 'FOO=bar' in result.stdout
 
 
+@pytest.mark.test_all_runtimes
```
nitzmahone: We can probably cover this adequately with decent unit tests on either the CLI parser or maybe even just the `BuildArgsAction` itself- another dedicated integration test is incredibly expensive for this case.

nitzmahone (changes requested):

nitzmahone:
Can also probably fix #520 easily in this same PR while you're "in the neighborhood" :wink:

src/ansible_builder/main.py
nitzmahone: On further reflection, the placement here is fine, but we definitely don't want the `''` in the final command for that case (just `{key}=`). The logic should probably be more explicit though, as the YAML layer *could* technically provide an integer `0` value (the JSON schema might currently prevent it, but I don't know for sure), which would erroneously trigger the `else`. 

So probably just to be super safe have `BuildArgAction` provide a value of `None` for eg `--build-arg FOO`, `''` for `--build-arg FOO=` and of course the real value for anything else... Then the logic here should explicitly check for value being `None` and `''`, then the unqualified `else` case is the `{key}={value}` one.

nitzmahone (approved):

Shrews (approved):

PR #518, opened 2023-05-03
Update labeler and issue-labeler actions

Akasurde:
* Nodejs 12 is deprecated, use Nodejs 16


Akasurde:
FYI https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

nitzmahone (approved):

PR #521, opened 2023-05-04
Allow to set user value for the container image

### LINKED ISSUE #506: `USER root` vs `USER 1000`
When using ansible-builder, including v3, the user is set as root in the container.

When I looked at the config for the Execution Environment for AWX, I see they add a step at the end to set `USER 1000` which changes behavior inside of the container
https://github.com/ansible/awx-ee/blob/devel/execution-environment.yml

Is there a reason that ansible-builder runs as root and awx-ee runs as 1000? Is one more secure than the other or more "correct"?  I would have expected similar behavior from both.

Akasurde:
* Added options.user setting
* Updated documentation with an example
* Added unit tests

Fixes: #506

### DIFF
--- a/demo/execution-environment.yml
+++ b/demo/execution-environment.yml
@@ -9,7 +9,7 @@ images:
     # name: registry.access.redhat.com/ubi9/ubi:latest  # vanilla image!
 
 options:
-  user: 1001  # (optional) sets the username or UID
+  user: 1000  # (optional) sets the username or UID
 
 dependencies:
   python_interpreter:
--- a/demo/execution-environment.yml
+++ b/demo/execution-environment.yml
@@ -9,7 +9,7 @@ images:
     # name: registry.access.redhat.com/ubi9/ubi:latest  # vanilla image!
 
 options:
-  user: 1000  # (optional) sets the username or UID
+  user: '1000'  # (optional) sets the username or UID
 
 dependencies:
   python_interpreter:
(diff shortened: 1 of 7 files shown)


demo/execution-environment.yml
```diff
@@ -8,6 +8,9 @@ images:
     # name: registry.fedoraproject.org/fedora:38  # vanilla image!
     # name: registry.access.redhat.com/ubi9/ubi:latest  # vanilla image!
 
+options:
+    user: 1001  # (optional) sets the username or UID
```
shanemcd: This doesn't match the docs below.

nitzmahone (commented):
looking good; might want to add a usage of `options->user` to the `test/data/v3/complete` EE def and one of the associated `create` integration tests to validate the right `USER` directive appearance in the final image stage.

test/integration/test_create.py
```diff
@@ -394,3 +395,23 @@ def test_v3_no_workdir(cli, build_dir_and_ee_yml):
...
+          user: bob
+        """
+    )
+    cli(f'ansible-builder create -c {tmpdir} -f {eeyml} --output-filename Containerfile')
+
+    containerfile = tmpdir / "Containerfile"
+    assert containerfile.exists()
+    text = containerfile.read_text()
+
+    assert "USER bob" not in text
```
nitzmahone: I assume this should be:
```suggestion
    assert "USER bob" in text
```

nitzmahone (approved):

PR #524, opened 2023-05-05
Added missing underscore in target_scripts path

Akasurde:
* Without this project data files are not copied in the installation which are required.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ python_requires = >=3.9
 
 
 [options.package_data]
-ansible_builder.target_scripts =
+ansible_builder._target_scripts =
     *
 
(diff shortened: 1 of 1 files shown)


nitzmahone:
While testing this fix, it appears that this option is just ignored under *almost* all circumstances- the only way I could actually get it to break in a way where this option was actually consulted was when directly `pip install` ing a `git clone` with the `.git` dir removed. `python -m build`, and `pip install` with a GH ref archive or a functional git worktree all basically seem to ignore this option and package everything.

I don't see any harm to having it actually describe the right thing, but we may end up needing to go back to `MANIFEST.in` at some point anyway. :man_shrugging: 

Thanks!

PR #528, opened 2023-05-08
Validate build_steps for USER directive

### LINKED ISSUE #527: Custom `USER` directives in `additional_build_steps` can break things...
We might want to consider issuing a warning during context generation if a `USER` directive is encountered in `additional _build_steps`. Changing user identity mid-build can break some of the cleanup tasks- eg, https://github.com/ansible/awx-ee/pull/205. In this particular case, the step is no longer necessary anyway, but since the `output` dir cleanup is currently  occurring after `append_final`, a `USER` directive may break the final cleanup stages of the build.

We could also defensively reset `USER` or move the cleanup earlier into one of the more "atomic" areas, but (for better or worse) that would prevent users from accessing the staging directory we copied over.

Akasurde:
* Providing USER details may lead to failure in the image building process. Validating presence of USER setting before hand may avoid this situation.

Fixes: #527

### DIFF
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -146,7 +146,7 @@ string, or a list of strings.
 
     Please make sure that you do not specify `USER` directives in these build steps. This may lead to
     failures while building the image.
-    If you want to override `USER` setting, consider using `options.user` setting mentioned below.
+    If you want to override the `USER` setting, consider using the `options.user` setting mentioned below.
 
 build_arg_defaults
 ^^^^^^^^^^^^^^^^^^
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -142,6 +142,12 @@ string, or a list of strings.
     ``append_final``
       Commands to insert after building of the final image.
 
+.. note::
+
+    Please make sure that you do not specify `USER` directives in these build steps. This may lead to
+    failures while building the image.
+    If you want to override `USER` setting, consider using `options.user` setting mentioned below.
+
 build_arg_defaults
 ^^^^^^^^^^^^^^^^^^
 
--- a/src/ansible_builder/user_definition.py
+++ b/src/ansible_builder/user_definition.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import textwrap
 import tempfile
@@ -10,6 +11,7 @@ from . import constants
 from .exceptions import DefinitionError
 from .ee_schema import validate_schema
 
+logger = logging.getLogger(__name__)
 
 # HACK: manage lifetimes more carefully
 _tempfiles: list[Callable] = []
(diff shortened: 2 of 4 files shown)


src/ansible_builder/user_definition.py
```diff
@@ -243,3 +243,13 @@ def validate(self):
                     self.build_arg_defaults['EE_BUILDER_IMAGE'] = self.builder_image.name
 
             self._validate_additional_build_files()
+
+            if self.version == 3:
+                build_steps = self.raw.get('additional_build_steps', {})
+                for step_name, steps in build_steps.items():
+                    for directive in steps:
+                        if directive.startswith('USER '):
+                            raise DefinitionError(
```
Shrews: I'm not certain that a hard error is the thing to do. As @nitzmahone suggested in the issue, we probably just want to output a warning message instead. Could be a valid reason for someone wanting to change the user.

docs/definition.rst
```diff
@@ -125,6 +125,12 @@ string, or a list of strings.
     ``append_final``
       Commands to insert after building of the final image.
 
+.. note::
+
+    Please make sure that you do not specify `USER` directives in these build steps. This may lead to
+    failures while building the image.
+    If you want to override `USER` setting, consider using `options.user` setting mentioned below.
```
samccann:
```suggestion
    If you want to override the `USER` setting, consider using the `options.user` setting mentioned below.
```

samccann (approved):
docs portion lgtm

src/ansible_builder/user_definition.py
```diff
@@ -243,3 +245,13 @@ def validate(self):
                     self.build_arg_defaults['EE_BUILDER_IMAGE'] = self.builder_image.name
 
             self._validate_additional_build_files()
+
+            if self.version == 3:
```
Shrews:
```suggestion
            if self.version >= 3:
```

src/ansible_builder/user_definition.py
```diff
@@ -243,3 +245,13 @@ def validate(self):
...
+                build_steps = self.raw.get('additional_build_steps', {})
+                for step_name, steps in build_steps.items():
+                    for directive in steps:
+                        if directive.startswith('USER '):
+                            logging.warning(
+                                f"Found USER directive in '{step_name}' in 'additional_build_steps'. "
+                                f"Remove this directive from '{step_name}', "
+                                f"since this may cause failure in the build process.")
```
Shrews: The `Remove this directive` part sounds more like a demand rather than an informative suggestion which would seem more appropriate for a warning message. Maybe something more like "Including this directive may cause failures in the build process."

Also, can we add a test that captures the output (maybe with `capsys`?) and verifies this gets output to the user?

Shrews (changes requested):

Shrews (approved):

PR #529, opened 2023-05-09
Add test image cleanup

### LINKED ISSUE #519: Fix image deletion in integration tests
#516 disabled the inline image deletion from integration tests, as it would sometimes corrupt the build cache. Test images will just pile up now- not a problem for CI, but definitely a problem for local execution. 

We don't want to just blindly `prune` or otherwise blast images that our test might not have created (at least not without also marking such tests as `destructive`), but the distributed nature of `pytest-xdist` complicates the necessary bookkeeping to defer explicit image deletion until the end of the session, since only the workers know what images they created, but the main session teardown hook runs in the main controller long after the required data is lost.

One idea:
* Have `delete_image` record the image names in a per-runtime `set` that's tied to a session-scoped fixture. At fixture teardown, merely append the data to well-known filenames in the session tempdir (using a lockfile/mutex or atomic write, since multiple workers may hit it at once). Then have the main session or unload hook read all the names for each runtime into a set (as some may be duplicated across workers) and best-effort delete all the images in those files, producing warnings if there are problems.

Shrews:
Removes any images created for tests.

Note that intermediary images are not removed. These can be cleaned up with a simple `podman image prune` or `docker image prune` command.

Fixes #519

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,9 +1,11 @@
+import filelock
 import os
 import pathlib
 import pytest
 import re
 import shutil
 import subprocess
+import tempfile
 import uuid
 import yaml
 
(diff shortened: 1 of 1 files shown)


Shrews:
Closed because GH was not updating this PR. Replaced with https://github.com/ansible/ansible-builder/pull/531.

PR #538, opened 2023-05-17
Added scenarios for version 3

Akasurde:

PR #544, opened 2023-05-21
Docs: mention a few base images as examples, using a free one as the default

felixfontein:
Right now the example base image can only be pulled with an account. This replaces it with a free alternative (UBI 9 base image), and lists several examples (Rocky Linux 9, CentOS Stream 9, and the original AAP base image).

### DIFF
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -48,6 +48,7 @@ Below is an example version 3 EE file:
         # Other available base images:
         #   - quay.io/rockylinux/rockylinux:9
         #   - quay.io/centos/centos:stream9
+        #   - registry.fedoraproject.org/fedora:38
         #   - registry.redhat.io/ansible-automation-platform-23/ee-minimal-rhel8:latest
         #     (needs an account)
 
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -48,7 +48,7 @@ Below is an example version 3 EE file:
         # Other available base images:
         #   - quay.io/rockylinux/rockylinux:9
         #   - quay.io/centos/centos:stream9
-        #   - registry.redhat.io/ansible-automation-platform-21/ee-minimal-rhel8:latest
+        #   - registry.redhat.io/ansible-automation-platform-23/ee-minimal-rhel8:latest
         #     (needs an account)
 
     additional_build_files:
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -44,7 +44,12 @@ Below is an example version 3 EE file:
 
     images:
       base_image:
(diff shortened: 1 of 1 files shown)


docs/definition.rst
```diff
@@ -44,7 +44,12 @@ Below is an example version 3 EE file:
 
     images:
       base_image:
-        name: registry.redhat.io/ansible-automation-platform-21/ee-minimal-rhel8:latest
+        name: docker.io/redhat/ubi9:latest
+        # Other available base images:
+        #   - quay.io/rockylinux/rockylinux:9
+        #   - quay.io/centos/centos:stream9
+        #   - registry.redhat.io/ansible-automation-platform-21/ee-minimal-rhel8:latest
```
samccann:
```suggestion
        #   - registry.redhat.io/ansible-automation-platform-23/ee-minimal-rhel8:latest
```

felixfontein: Does that one work fine? I cannot access it (to test it).

samccann: I copied it from https://github.com/ansible/ansible-builder/pull/538 and @Akasurde said it works :-)

Akasurde:
```
...
Running command:
  podman build -f context/Containerfile -t ee_rhel_8 --no-cache context
STEP 1: FROM registry.redhat.io/ansible-automation-platform-23/ee-minimal-rhel8:latest AS base
STEP 2: USER root
6d0a685aeb939ba6893b35348ff5e9787d44a1a670c02f12ba8b0acb1ac56c23
STEP 3: ARG EE_BASE_IMAGE
...
```

felixfontein:
I just noticed that https://github.com/ansible/ansible-builder/blob/devel/demo/execution-environment.yml also contains some images; I've added the Fedora 38 example from there.

samccann (approved):
LGTM

nitzmahone (approved):

nitzmahone:
Thanks- definitely makes sense to show that this works with lots of different base images.

I'm a little torn on how much further to go with that on this very first sample though. I'm thinking we might want to just make the first one dirt-simple, then follow up with a number of "cookbook" style examples that delve into individual features. Ideally most folks will just need bare bones core/runner + a few collections that will auto-handle all the deps. Especially for the RH-supplied downstream images that already have Python/Core/Runner, the EE def can literally just be "base image + a couple collections", but is slightly more complex because it uses microdnf.

Assuming RHEL.next converges on `dnf5`, all will be wonderful then, right? :laughing:

samccann:
https://github.com/ansible/ansible-builder/pull/538 starts that cookbook like approach as a set of scenarios fwiw

felixfontein:
@samccann @Akasurde @nitzmahone thanks for reviewing and merging!

PR #546, opened 2023-05-23
Fix pulp integration test

Akasurde:
* Use `--no-cache` parameter to build the image from scratch since image signing is not working on intermediate/cached layers.


sivel:
Since this sounds like it would be the path many customers will be taking, should we do a bit more here instead of just correcting the test? I'm concerned that if we just fix this in tests, we might be hiding a common issue that users may run into.

PR #555, opened 2023-06-05
[RFE] Support for tags from the configuration file

### LINKED ISSUE #523: [RFE]  Allow the user to specify a default tag in the `execution-environment.yml`
If added as an optional key, to replace the default, it would allow for a "right-click" build without having to prompt the user in VSCode.

This would also have the side benefit of identifying which EE a  given `execution-environment.yml` file was for as today, there is no metadata within the execution-environment.yml indicating the name of the EE being built. 

Related: https://github.com/ansible/vscode-ansible/issues/878

Akasurde:
##### SUMMARY

* Added support for providing tags from the configuration file

Fixes: #523

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request
- Test Pull Request

##### COMPONENT NAME
demo/execution-environment.yml
docs/definition.rst
src/ansible_builder/ee_schema.py
src/ansible_builder/main.py
test/unit/test_user_definition.py

### DIFF
--- a/src/ansible_builder/ee_schema.py
+++ b/src/ansible_builder/ee_schema.py
@@ -329,7 +329,7 @@ schema_v3 = {
         },
 
         "options": {
-            "description": "Options that affects the runtime behavior",
+            "description": "Options that affect the runtime behavior",
             "type": "object",
             "additionalProperties": False,
             "properties": {
@@ -374,7 +374,7 @@ schema_v3 = {
                 },
                 "tags": {
                     "description": "Specifies the name which is assigned to resulting image if build process completes successfully",
-                    "type": ["string", "array", "null"],
+                    "type": "array",
                 },
             },
         },
--- a/src/ansible_builder/ee_schema.py
+++ b/src/ansible_builder/ee_schema.py
@@ -329,7 +329,7 @@ schema_v3 = {
         },
 
         "options": {
-            "description": "Options that effect runtime behavior",
+            "description": "Options that affects the runtime behavior",
             "type": "object",
             "additionalProperties": False,
             "properties": {
@@ -371,6 +371,10 @@ schema_v3 = {
                             "type": "string",
                         },
                     },
+                },
+                "tags": {
+                    "description": "Specifies the name which is assigned to resulting image if build process completes successfully",
+                    "type": ["string", "array", "null"],
                 },
             },
         },
@@ -444,3 +448,4 @@ def _handle_options_defaults(ee_def: dict):
         'cmd': '["bash"]',
     })
     options.setdefault('user', '1000')
+    options.setdefault('tags', [])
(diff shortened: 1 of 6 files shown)


src/ansible_builder/ee_schema.py
```diff
@@ -372,6 +372,10 @@
                         },
                     },
                 },
+                "tags": {
+                    "description": "Specifies the name which is assigned to resulting image if build process completes successfully",
+                    "type": ["string", "array", "null"],
```
Shrews: Why allow `null`? I'd rather we only accept `array` here so we don't have to do code gymnastics to determine if it's a string or a list in other parts of the code.

src/ansible_builder/ee_schema.py
```diff
@@ -329,7 +329,7 @@
         },
 
         "options": {
-            "description": "Options that effect runtime behavior",
+            "description": "Options that affects the runtime behavior",
```
Shrews:
```suggestion
            "description": "Options that affect the runtime behavior",
```

Shrews (changes requested):
We should probably test and account for a user specifying tags in both the EE file and on the command line. Not sure of the proper way to handle that case... combine the tags? ... prefer CLI over EE (similar to `--build-arg` CLI flag)?

src/ansible_builder/ee_schema.py
Akasurde: Ok.

Akasurde:
> We should probably test and account for a user specifying tags in both the EE file and on the command line. Not sure of the proper way to handle that case... combine the tags? ... prefer CLI over EE (similar to `--build-arg` CLI flag)?

Right now I am combining the tags, if the user provided tags on both cli and ee def files.

nitzmahone:
CLI should override, not combine- that could cause unexpected behavior... The in-definition tags should only override the default name if nothing else is provided (this keeps the same behavior we have today, just allows the EE def to change the default name).

src/ansible_builder/ee_schema.py
```diff
@@ -372,6 +372,10 @@
                         },
                     },
                 },
+                "tags": {
+                    "description": "Specifies the name which is assigned to resulting image if build process completes successfully",
+                    "type": "array",
```
Shrews, line 377:
```suggestion
                    "type": "array",
                    "items": {
                        "type": "string"
                    }

```

src/ansible_builder/ee_schema.py
```diff
@@ -372,6 +372,10 @@
                         },
                     },
                 },
+                "tags": {
+                    "description": "Specifies the name which is assigned to resulting image if build process completes successfully",
```
Shrews:
```suggestion
                    "description": "A list of names to assign to the resulting image if build process completes successfully",
```

Shrews (changes requested):
Suggested some modifications to the schema to make it more robust.

Also, I tested this locally and the CLI `-t` option is not overriding the tags listed in the EE.

Shrews (approved):

PR #556, opened 2023-06-06
Raise error when invalid requirements are parsed

### LINKED ISSUE #364: Invalid requirements.txt leads to container built incorrectly
If you reference a requirements.txt for python modules but it has invalid syntax `ansible-builder build` will make the image but the python modules will not be installed. Even when running `-v 3` you wont see any errors with the build process. After some time of digging I found the introspect command in the `Containerfile` and running this like `ansible-builder introspect --sanitize --user-pip=requirements.txt --user-bindep=bindep.txt --write-bindep=/tmp/src/bindep.txt --write-pip=/tmp/src/requirements.txt` finally revealed my syntax error:
```
# Sanitized dependencies for /usr/share/ansible/collections
Warning: failed to parse requirments from user, error: Invalid requirement, parse error at "'=4.0.30'"
---
python: []
system:
- 'unixODBC  # from collection user'
- 'unixODBC-devel  # from collection user'
- 'gcc  # from collection user'
- 'gcc-c++  # from collection user'
- 'python38-devel  # from collection user'

Creating parent directory for /tmp/src/bindep.txt
```

In my case I did `module=version` instead of `module==version` but this lead to a great deal of frustration trying to figure out why my modules did not exist in the container. I realize that this particular issue was user induced is on my end, but it led me to wish for a mechanism that would allow me to see when there are warnings or errors like these instead of an incomplete container build.

Akasurde:
##### SUMMARY

Fixes: #364

##### ISSUE TYPE
- Bugfix Pull Request

- Test Pull Request

##### COMPONENT NAME
src/ansible_builder/_target_scripts/introspect.py
test/unit/test_introspect.py

### DIFF
--- a/src/ansible_builder/_target_scripts/introspect.py
+++ b/src/ansible_builder/_target_scripts/introspect.py
@@ -344,8 +344,9 @@ def sanitize_requirements(collection_py_reqs):
                 consolidated.append(req)
                 seen_pkgs.add(req.name)
         except Exception as e:
-            logger.error('Failed to parse requirements from %s, error: %s', collection, e)
-            sys.exit(1)
+            msg = f'Failed to parse requirements from {collection}, error: {e}'
+            logger.error(msg)
+            sys.exit(msg)
 
     # removal of unwanted packages
     sanitized = []
--- a/src/ansible_builder/_target_scripts/introspect.py
+++ b/src/ansible_builder/_target_scripts/introspect.py
@@ -346,7 +346,7 @@ def sanitize_requirements(collection_py_reqs):
         except Exception as e:
             msg = f'Failed to parse requirements from {collection}, error: {e}'
             logger.error(msg)
-            sys.exit(msg)
+            sys.exit(1)
 
     # removal of unwanted packages
     sanitized = []
--- a/src/ansible_builder/_target_scripts/introspect.py
+++ b/src/ansible_builder/_target_scripts/introspect.py
@@ -344,7 +344,8 @@ def sanitize_requirements(collection_py_reqs):
                 consolidated.append(req)
                 seen_pkgs.add(req.name)
         except Exception as e:
-            logger.warning('Warning: failed to parse requirements from %s, error: %s', collection, e)
+            logger.error('Failed to parse requirements from %s, error: %s', collection, e)
+            sys.exit(1)
 
     # removal of unwanted packages
     sanitized = []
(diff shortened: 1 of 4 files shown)


test/conftest.py
```diff
@@ -46,6 +46,8 @@ def pytest_sessionfinish(session, exitstatus):
     # If we are the main worker thread, we've been called last, so do the cleanup.
     if worker_id is None:
         for runtime in CONTAINER_RUNTIMES:
+            if runtime not in FOUND_RUNTIMES:
+                continue
```
Shrews: This has nothing to do with invalid requirements parsing, so confusing why this is here.

Also, I'm fairly certain the reason I didn't use FOUND_RUNTIMES here when adding this code was to facilitate cleanup of images when running locally (the entire reason I added this callback) where you might have failures across different runtimes across multiple runs. Using CONTAINER_RUNTIMES makes sure we always check for left over images, even if we chose to skip a runtime at some point.

src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -344,7 +344,9 @@ def sanitize_requirements(collection_py_reqs):
                 consolidated.append(req)
                 seen_pkgs.add(req.name)
         except Exception as e:
-            logger.warning('Warning: failed to parse requirements from %s, error: %s', collection, e)
+            msg = f'Failed to parse requirements from {collection}, error: {e}'
+            logger.error(msg)
+            sys.exit(msg)
```
Shrews: This causes the error message to be printed twice:

```
[2/3] STEP 11/12: RUN $PYCMD /output/scripts/introspect.py introspect --sanitize --user-pip=requirements.txt --write-bindep=/tmp/src/bindep.txt --write-pip=/tmp/src/requirements.txt
Failed to parse requirements from user, error: Expected end or semicolon (after name and no valid version specifier)
    six=0.1
       ^
Failed to parse requirements from user, error: Expected end or semicolon (after name and no valid version specifier)
    six=0.1
       ^
```

Probably enough to just `sys.exit(1)` here.

Shrews (changes requested):

test/conftest.py
Akasurde: Sorry this was a local change. Inadvertently got into this PR.

Shrews:
Missed re-reviewing this, but looks good to me. This is technically a behavior change, so I'm wondering if maybe we should either add a new CLI switch to turn this warning into a hard error, or possible a new option under the `options` EE config?

@nitzmahone Any input on this?

PR #558, opened 2023-06-07
Document PKGMGR_PRESERVE_CACHE

### LINKED ISSUE #557: Need to document PKGMGR_PRESERVE_CACHE
This feature was added in PR #516. We should document what it is.

Akasurde:
##### SUMMARY

* documented PKGMGR_PRESERVE_CACHE

Fixes: #557

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
docs/definition.rst

### DIFF
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -150,6 +150,13 @@ Build args used by ``ansible-builder`` are the following:
     ``ANSIBLE_GALAXY_CLI_ROLE_OPTS``
       This allows the user to pass any flags, such as `--no-deps`, to the role installation.
 
+    ``PKGMGR_PRESERVE_CACHE``
+      This controls how often the package manager cache is cleared during the image build process.
+      If this value is not set, which is the default, the cache is cleared frequently.
+      If it is set to the string `always`, the cache is never cleared.
+      Any other value forces the cache to be cleared only after the system dependencies are installed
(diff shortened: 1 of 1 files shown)


docs/definition.rst
```diff
@@ -146,6 +146,13 @@ Build args used by ``ansible-builder`` are the following:
     ``ANSIBLE_GALAXY_CLI_ROLE_OPTS``
       This allows the user to pass any flags, such as `--no-deps`, to the role installation.
 
+    ``PKGMGR_PRESERVE_CACHE``
+      This allows the user to control cache maintained by the package manager used.
+      Default is `None` which clears the package manager cache frequently.
+      If set to `always` never clears the package manager cache (This is good for CI, testing base
+      images where a pre-populated package cache is handy).
+      Any value but `always` clears the cache only at the end of `install-from-bindep` (This is good for CI, squashed images).
+
```
Shrews, line 159: I'd suggest something more like:

```
This controls how often the package manager cache is cleared during the image build process.
If this value is not set, which is the default, the cache is cleared frequently. If it is set
to the string `always`, the cache is never cleared. Any other value forces the cache to
be cleared only after the system dependencies are installed in the final build stage.
```

Shrews (approved):

PR #559, opened 2023-06-08
Improve CLI usage docs

Andersson007:
usage.rst: use version 3, use single-file approach

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -21,18 +21,11 @@ file as a starting point:
     ---
     version: 3
     dependencies:
-      galaxy: requirements.yml
+      galaxy:
+        collections:
+        - name: awx.awx
 
-
-The content of ``requirements.yml``:
-
-.. code:: yaml
-
-   ---
-   collections:
-     - name: awx.awx
-
-To build an execution environment using the files above, run:
+To build an Execution Environment using the file above, run:
 
 .. code::
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -25,11 +25,9 @@ To build an execution environment using the default definition file, run:
 .. code::
 
    $ ansible-builder build
-   ...
-   STEP 7: COMMIT my-ee
-   --> 09c930f5f6a
-   09c930f5f6ac329b7ddb321b144a029dbbfcc83bdfc77103968b7f6cdfc7bea2
-   Complete! The build context can be found at: context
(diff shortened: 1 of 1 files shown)


cidrblock (approved):
awesome.  I think the single file approach is the smoothest path for most

Andersson007:
@cidrblock someone changed version to 3 in parallel.
the conflicts are solved, could you please merge if it's OK now

Shrews (commented):
Thanks for the PR. I missed that this was changed from version 1 to version 3 in the recent docs update.

I'm sort of conflicted on this dated example even being here because:
  - It repeats information that we've more completely defined in the EE def section.
  - It encourages using the default base image (which is there for backward compatibility for older EE versions) which is no longer maintained and will likely cause issues for folks who may choose to expand on this example.

I'd actually be in favor of removing this example and just linking the reader to the V3 docs in `docs/definition.rst`. Otherwise, if we want to keep this example, it should probably be expanded to include a different base image and any other options required for it to build properly.

Andersson007:
@Shrews hello, thanks for reviewing!

- removed the example
- restructured the paragraph a bit
- fixed a typo

PTAL

docs/usage.rst
```diff
@@ -11,34 +11,22 @@ Ansible Builder can execute two separate steps. The first step is to create a bu
...
 
-To build an execution environment using the files above, run:
+To build an execution environment using the default definition file, run:
 
 .. code::
 
    $ ansible-builder build
    ...
-   STEP 7: COMMIT my-awx-ee
+   STEP 7: COMMIT my-ee
```
Shrews: Technically, this output isn't correct, especially the container tag name, which is not the default. Unless the `-v` verbosity option is used, the only thing the user will see with this command is (with slight modification to context path):

```
Running command:
  podman build -f context/Containerfile -t ansible-execution-env:latest context
Complete! The build context can be found at: /path/to/context
```

Andersson007: so can i just update the output showing this ^ ?

Shrews: > so can i just update the output showing this ^ ?

Sure

Andersson007: thanks, done, PTAL

Shrews (approved):

Andersson007:
@cidrblock @Shrews thank you!

PR #564, opened 2023-06-14
Support multiple -v for verbosity

Shrews:


Akasurde (approved):

sean-m-sullivan (approved):

docs/usage.rst
```diff
@@ -177,6 +177,18 @@ Customizes the level of verbosity:
...
+the verbosity level, or supply multiples of the option. For example, the following are
+equivalent to setting the verbosity level to ``3``:
+
+.. code::
+
+   $ ansible-builder build -v 3
+
+.. code::
+
+   $ ansible-builder build -vvv
```
nitzmahone, line 187: For pedantic completeness, do we also want to mention that individual `v` args will stack? (eg, `-v -v -v` == `-vvv`, at least last time I checked in core)?

I'm totally cool if the answer is "no", BTW :wink:

src/ansible_builder/cli.py
```diff
@@ -164,12 +187,11 @@ def add_container_options(parser):
 
         n.add_argument('-v', '--verbosity',
                        dest='verbosity',
-                       type=int,
-                       choices=[0, 1, 2, 3],
+                       action=CustomVerbosityAction,
+                       nargs='?',
```
nitzmahone, line 191: `nargs='?'` is  potentially a big hammer here, since it can falsely trip when it's the last non-positional arg followed by a positional arg (and I can't recall offhand if you can "push" a token back for the next subparser). `--` makes it work, but a lot of people don't understand that.

I *think* at least for builder, It's actually OK right now, since none of our existing command subparsers currently support trailing posargs, but just something to acknowledge (as well as perhaps a good reason to actually deprecate the numeric syntax?), and to make sure that an existing runner subparser doesn't have this problem.

nitzmahone (commented):
Overall LGTM- I don't think you need to change anything for the stuff I pointed out, but something to think about, esp if we give runner the same treatment.

nitzmahone:
(and bee-tee-dubs, THANK YOU)

docs/usage.rst
Shrews: Done

src/ansible_builder/cli.py
Shrews: It should be cool for now b/c we don't have positional args, but I did add deprecating the old style in a new deprecation tracking issue https://github.com/ansible/ansible-builder/issues/565.

Issue #569, opened 2023-06-29
Insertion of  'prepend_builder' contents in generated Containerfile incorrect


tuxpreacher:
When using `prepend_builder` in `execution-environment.yml` to inject pip configuration similar to this:

```
prepend_builder:
  - RUN pip config --user set global.index-url http://container.core.example.com:80/simple/
  - RUN pip config --user set global.trusted-host container.core.example.com
  - RUN pip config --user set global.proxy http://proxy_user:pasword123@container.core.example.com:3128
``` 

The resulting `ContainerFile` has these commands _after_ a `pip install` command:

```
RUN $PYCMD -m pip install --no-cache-dir bindep pyyaml requirements-parser
RUN pip config --user set global.index-url http://container.core.example.com:80/simple/
RUN pip config --user set global.trusted-host container.core.example.com
RUN pip config --user set global.proxy http://proxy_user:pasword123@container.core.example.com:3128
```

It would be good to have these before the `RUN $PYCMD` so that pip configuration can be carried out in the builder stage.

Shrews:
As noted in https://github.com/ansible/ansible-builder/issues/566#issuecomment-1598582776, we recommend that if you need to configure `pip` that you do so in the base image, which will then be inherited by all of the other stages, including the builder stage. You should also use the `RUN $PYCMD -m pip` format since an image may have multiple python installations, and the EE definition can select which one to use (via `dependencies.python_interpreter`).

tuxpreacher:
@Shrews ooops, I missed that closed issue, sorry about that. Then my issue can be closed.

tuxpreacher:
Closing based on previous issue #566

tuxpreacher:
Closing based on previous issue #566

TheBig-svg:
@tuxpreacher so you fully ignore enterprise environments using public images.
At the end usability will finally decide if a software is crap or useful.

When we need to build an image, I don`t need ansible-builder anymore. I can put everything into the image build without the requirement to use fedora (pip_install bug) or build a new image to configure pip.

You your answer is some kind of "ignoring enterprise customer requirements" and this is RedHat software which is build for enterprise environments ;)

PR #570, opened 2023-06-29
Fix for ansible-builder issue #569

tuxpreacher:
Proposed fix for #569 . I hope I'm doing this right.


Shrews:
Proper solution discussed in the referenced issue is to use `prepend_base`.

PR #572, opened 2023-07-05
Update version 3 definition docs

Akasurde:
##### SUMMARY

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
docs/definition.rst

### DIFF
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -69,7 +69,7 @@ Here is a sample version 3 EE file. To use Ansible Builder 3.x, you must specify
 
     additional_build_steps:
       prepend_base:
-        - RUN echo This is a preprend base command!
+        - RUN echo This is a prepend base command!
         # Enable Non-default stream before packages provided by it can be installed. (optional)
         # - RUN $PKGMGR module enable postgresql:15 -y
         # - RUN $PKGMGR install -y postgresql
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -59,17 +59,22 @@ Here is a sample version 3 EE file. To use Ansible Builder 3.x, you must specify
         #   - registry.redhat.io/ansible-automation-platform-23/ee-minimal-rhel8:latest
         #     (needs an account)
 
+    # Custom package manager path for the RHEL based images
+    # options:
+    #  package_manager_path: /usr/bin/microdnf
+
     additional_build_files:
         - src: files/ansible.cfg
           dest: configs
 
     additional_build_steps:
       prepend_base:
+        - RUN echo This is a preprend base command!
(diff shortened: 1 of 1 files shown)


docs/definition.rst
```diff
@@ -69,7 +73,7 @@ Here is a sample version 3 EE file. To use Ansible Builder 3.x, you must specify
         - RUN $PKGMGR module enable postgresql:15 -y
         - RUN $PKGMGR install -y postgresql
       prepend_galaxy:
-        - ADD _build/configs/ansible.cfg ~/.ansible.cfg
+        - ADD _build/configs/ansible.cfg /runner/.ansible.cfg
```
Shrews: The example base image actually has `/home/runner` set as `$HOME`, but perhaps we can avoid the whole "knowing what $HOME is in the selected base image" issue by just using `/etc/ansible/ansible.cfg` in the example?

Shrews (changes requested):
Other things that I noticed are wrong with this example (if you could fix them, too):

- The selected base image has Python 3.8, so `ansible-core==2.14.4` is not going to work since that requires 3.9. A `2.13` version should work though.
- The selected base image uses `microdnf`, so we need to set that in `options.package_manager_path`.

docs/definition.rst
```diff
@@ -59,17 +59,22 @@ Here is a sample version 3 EE file. To use Ansible Builder 3.x, you must specify
...
+    # options:
+    #  package_manager_path: /usr/bin/microdnf
+
     additional_build_files:
         - src: files/ansible.cfg
           dest: configs
 
     additional_build_steps:
       prepend_base:
+        - RUN echo This is a preprend base command!
```
Shrews:
```suggestion
        - RUN echo This is a prepend_base command!
```

docs/definition.rst
```diff
@@ -59,17 +59,22 @@ Here is a sample version 3 EE file. To use Ansible Builder 3.x, you must specify
...
         # Enable Non-default stream before packages provided by it can be installed. (optional)
-        - RUN $PKGMGR module enable postgresql:15 -y
-        - RUN $PKGMGR install -y postgresql
+        # - RUN $PKGMGR module enable postgresql:15 -y
```
Shrews, line 74: Why comment these out?

Akasurde: Postgresql module stream is available in `ee-minimal-rhel8` or downstream images. Also, this is an optional step there is no actual use of installing postgresql.

docs/definition.rst
Akasurde: Thanks

Shrews (approved):

PR #575, opened 2023-07-10
Add pylint to CI

Akasurde:
Adding `pylint` to the CI linting tests.

The plan is to slowly enable individual checks until we get the number of errors down a bit. Disabled all checks for now except for `W0102 - dangerous-default-value`.

### DIFF
--- a/src/ansible_builder/_target_scripts/introspect.py
+++ b/src/ansible_builder/_target_scripts/introspect.py
@@ -208,7 +208,7 @@ def simple_combine(reqs):
     return fancy_lines
 
 
-def parse_args(args=None):
+def parse_args():
 
     parser = argparse.ArgumentParser(
         prog='introspect',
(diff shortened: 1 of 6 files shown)


src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -208,7 +208,7 @@ def simple_combine(reqs):
     return fancy_lines
 
 
-def parse_args(args=sys.argv[1:]):
+def parse_args(args=None):
```
Shrews, line 211: Is this change because of W0102 being enabled? If so, `args` needs to be set to the sys.args value within the function.

src/ansible_builder/cli.py
```diff
@@ -194,7 +194,7 @@ def add_container_options(parser):
                             'Integer values are also accepted (for example, "-v3" or "--verbosity 3"). Default is %(default)s.')
 
 
-def parse_args(args=sys.argv[1:]):
+def parse_args(args=None):
```
Shrews, line 197: How in the world does this not break things?

Shrews (approved):

PR #576, opened 2023-07-11
Update code sanity

Akasurde:
* Enable (C0206) consider-using-dict-items
* Enable (C0209) consider-using-f-string
* Enable (R0402) consider-using-from-import
* Enable (R1705) no-else-return
* Enable (R1725) super-with-arguments
* Enable (R1735) use-dict-literal
* Enable (W0102) dangerous-default-value
* Enable (W0707) raise-missing-from
* Enable (W1203) logging-fstring-interpolation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,10 +13,22 @@ disable = [
   # Some codes we will leave disabled
   "C0103",  # invalid-name
   "C0114",  # missing-module-docstring
+  "C0115",  # missing-class-docstring
   "C0116",  # missing-function-docstring
+
   "R0902",  # too-many-instance-attributes
 ]
 
 enable = [
+  "C0206",  # consider-using-dict-items
+  "C0209",  # consider-using-f-string
+
+  "R0402",  # consider-using-from-import
+  "R1705",  # no-else-return
+  "R1725",  # super-with-arguments
+  "R1735",  # use-dict-literal
+
   "W0102",  # dangerous-default-value
+  "W1203",  # logging-fstring-interpolation
+
 ]
--- a/src/ansible_builder/constants.py
+++ b/src/ansible_builder/constants.py
@@ -13,15 +13,15 @@ default_container_runtime = 'podman'
 base_roles_path = '/usr/share/ansible/roles'
 base_collections_path = '/usr/share/ansible/collections'
 
-build_arg_defaults = dict(
+build_arg_defaults = {
     # empty string values here still allow the build arg to be emitted into the generated Containerfile
-    ANSIBLE_GALAXY_CLI_COLLECTION_OPTS='',
-    ANSIBLE_GALAXY_CLI_ROLE_OPTS='',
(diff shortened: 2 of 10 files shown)


pyproject.toml
```diff
@@ -14,9 +14,10 @@ disable = [
   "C0103",  # invalid-name
   "C0114",  # missing-module-docstring
   "C0116",  # missing-function-docstring
-  "R0902",  # too-many-instance-attributes
```
Shrews, line 17: I think this is one we always want to have disabled.

pyproject.toml
```diff
@@ -14,9 +14,10 @@ disable = [
...
+  "C0115",  # missing-class-docstring
```
Shrews, line 16: I don't mind adding this, but let's at least keep the list ordered.

pyproject.toml
```diff
@@ -14,9 +14,10 @@ disable = [
...
 ]
 
 enable = [
   "W0102",  # dangerous-default-value
+  "C0209",  # consider-using-f-string
```
Shrews, line 24: Maybe keep the list in alphabetical order.

src/ansible_builder/containerfile.py
```diff
@@ -290,15 +301,15 @@ def _handle_additional_build_files(self):
             ee_file = Path(self.definition.filename)
             if src.is_absolute():
                 if not src.exists():
-                    logger.warning(f"User build file {src} does not exist.")
+                    logger.warning("User build file %s does not exist.", src)
```
Shrews, line 304: This and the other logging changes seems to be an unrelated change. FWIW, I'm fine with this format, but we should at least enable the pylint logging checker and the checks for logging string format (W1202, W1203) so that we are consistent. Perhaps that should be a separate change, but fine if you want to do it here.

src/ansible_builder/constants.py
```diff
@@ -13,15 +13,15 @@
 base_roles_path = '/usr/share/ansible/roles'
 base_collections_path = '/usr/share/ansible/collections'
 
-build_arg_defaults = dict(
+build_arg_defaults = {
```
Shrews, line 16: Unrelated change, but ok with it. I think there may be a pylint check code for this, actually? Might want to enable it.

src/ansible_builder/ee_schema.py
```diff
@@ -391,8 +392,8 @@ def validate_schema(ee_def: dict):
     if 'version' in ee_def:
         try:
             schema_version = int(ee_def['version'])
-        except ValueError:
-            raise DefinitionError(f"Schema version not an integer: {ee_def['version']}")
+        except ValueError as e:
+            raise DefinitionError(f"Schema version not an integer: {ee_def['version']}") from e
```
Shrews, line 396: Another unrelated change, but seems techincally correct to me. Is there a pylint check for this that we can enable?

Shrews (changes requested):

Shrews (approved):

PR #577, opened 2023-07-12
Make the capitalization of Execution Environment coherent

Fale:
My assumption is that the correct one would be "Execution Environment", since it is abbreviated "EE", but if the correct one is different, please let me know and I will change the PR to match the preferred capitalization.

### DIFF
--- a/docs/collection_metadata.rst
+++ b/docs/collection_metadata.rst
@@ -3,7 +3,7 @@
 Collection-level dependencies
 =============================
 
-When Ansible Builder installs collections into an execution environment, it also installs the dependencies listed by each collection on Galaxy.
+When Ansible Builder installs collections into an Execution Environment, it also installs the dependencies listed by each collection on Galaxy.
 
 For Ansible Builder to find and install collection dependencies, those dependencies must be defined in one of these files:
 
@@ -29,7 +29,7 @@ How to verify collection-level metadata
 
 .. note::
 
-  Running the introspect command described below is not a part of a typical workflow for building and using execution environments.
+  Running the introspect command described below is not a part of a typical workflow for building and using Execution Environments.
 
 Collection developers can verify that dependencies specified in the collection will be processed correctly by Ansible Builder.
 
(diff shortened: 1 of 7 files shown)


Shrews:
ping @acozine ... since you helped rewrite a lot of these docs, do you have any thoughts on this?

Akasurde:
cc @samccann Could you please review this? Thanks in advance.

samccann:
Hi @Fale and thanks for opening the docs PR!  So I would normally agree with you 100%. Unfortunately, we are a bit stuck. Execution environments from a Red Hat perspective, don't have capitalization.  Now the difficult part?  We as community have already gotten used to calling them EEs as an abbreviation for execution environments. I don't think we want to start calling them ees so we'll have to live with the discrepency at this point.

So I don't think we want to make the changes in this PR.

Fale:
@samccann this is messy :-/

I think we should try to think the issue in 2 ways:
1. Should the community (ie: this website and many others) be implementing Red Hat's standard? (personally I think it would make sense)
2. If so, can we document those standards (also for Plyabooks vs playbooks, Inventory vs inventory, etc)? Creating a page about it would really help to start working in a standardized direction.

Interestingly enough, it seems like the "execution environment" capitalization memo was not received everywhere in Red Hat as well, since there are a lot of inconsistencies there as well.

samccann:
@felixfontein valid points. My personal feeling is don't capitalize anything that isn't a specific product or project name.  (even tho that may differ from Red Hat's guidelines).  But i could go either way. 

I'll verify internally with RH writers about EEs and also see if they have a list we could..ahem..borrow for these decisions.

samccann:
Thanks @Fale for your patience here as we dig into the details.

So official RH style is lowercase execution environment.  

Since we have attention here - should we stop calling them EEs and just spell out execution environment everywhere? @felixfontein @Andersson007 ?

felixfontein:
I think using the abbreviation EE for execution environment is still useful. You don't always want to spell out the full name.

Andersson007:
> Since we have attention here - should we stop calling them EEs and just spell out execution environment everywhere?

My opinion would be to use at  least at the beginning of each docs `execution environment (EE)` or `execution environment (hereinafter EE)` and then allow to use the abbreviation. It's OK using things like `FQCN`, etc. as it looks good to me and saves time:) Moreover, people often use the abbreviations in their speech.

I'm not against using `execution environment` instead but in this case we should use this approach everywhere (with things like FQCN too...)

samccann:
no need for ``hereinafter` but otherwise agree, thanks

I opened a separate issue so we can formalize some of our terminology - https://github.com/ansible/ansible-documentation/issues/152

Thanks @Fale for getting the ball rolling on formalizing these terms. I'm closing this one since we have consensus to keep it lowercase (and keep EE) for now.

samccann:
AAHAHAHA omgosh I have delusions of grandeur!  Since this isn't the main ansible docs, i don't get to close etc. So up to the builder folks to decide if they agree with us at this point.

Shrews:
@samccann As you wish....  :)

PR #591, opened 2023-08-09
Warn user about using default base image in v3

Akasurde:
##### SUMMARY

* Default base image is old, using it in v3 definition might
  result in build failures. It is good idea to fail before proceeding.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
src/ansible_builder/user_definition.py
test/unit/test_user_definition.py

### DIFF
--- a/src/ansible_builder/user_definition.py
+++ b/src/ansible_builder/user_definition.py
@@ -246,12 +246,12 @@ class UserDefinition:
                     self.build_arg_defaults['EE_BUILDER_IMAGE'] = self.builder_image.name
 
             if (
-                self.version >= 3 and
-                self.build_arg_defaults["EE_BASE_IMAGE"] == constants.build_arg_defaults["EE_BASE_IMAGE"]
+                self.version >= 3
+                and self.build_arg_defaults["EE_BASE_IMAGE"] == constants.build_arg_defaults["EE_BASE_IMAGE"]
             ):
-                raise DefinitionError(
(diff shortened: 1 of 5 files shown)


src/ansible_builder/user_definition.py
```diff
@@ -245,6 +245,15 @@ def validate(self):
                     self.builder_image = ImageDescription(images, 'builder_image')
                     self.build_arg_defaults['EE_BUILDER_IMAGE'] = self.builder_image.name
 
+            if (
+                self.version >= 3
+                and self.build_arg_defaults["EE_BASE_IMAGE"] == constants.build_arg_defaults["EE_BASE_IMAGE"]
+            ):
+                logging.warning(
+                    "Using the outdated base image '%s' might "
+                    "result in the build failures.", self.build_arg_defaults['EE_BASE_IMAGE']
```
Shrews, line 254:
```suggestion
                    "result in build failures.", self.build_arg_defaults['EE_BASE_IMAGE']
```

Shrews: Downstream AAP packaged builder patches the value for `EE_BASE_IMAGE` in `constants.py`. We likely don't want this warning appearing in that case. Maybe we should just hardcode the quay image string here?

Shrews (changes requested):
Oops, sorry I missed reviewing this.

src/ansible_builder/user_definition.py
Akasurde: Done.

Shrews (approved):

PR #592, opened 2023-08-11
Porting guide for version 3

### LINKED ISSUE #561: Add docs section for converting from older builder files to v3
It was mentioned in a demo that this is fairly easy to do, but would be  a common section readers might look for.  iI'd suggest starting a small 'porting guide' for builder and add this to it.

From the demo, the steps are:
- add version 3 to the execution_environment.yml file
- if you refer to a builder image, remove it.
- if it is a version 1, you will need to add the images.

(take these steps w/ a grain of salt - needs to be verified with someone who knows.)

Akasurde:
##### SUMMARY

Fixes: #561

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
docs/definition.rst
docs/index.rst
docs/porting_guide_v3.rst

### DIFF
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -210,7 +210,7 @@ The following keys are valid for this section:
             package_pip: https://github.com/example_user/ansible/archive/refs/heads/ansible.tar.gz
 
     ``ansible_runner``
-      The version of the Ansible Runner Python package is to be installed. This value
+      The version of the Ansible Runner Python package to be installed. This value
       is a dictionary with a single key, ``package_pip``. The ``package_pip`` value
       is passed directly to `pip` for installation and can be in any format that
       pip supports. Below are some example values:
@@ -240,7 +240,7 @@ The following keys are valid for this section:
       (``python_path``).
 
     ``system``
-      The system packages are to be installed, in bindep format. This may either
+      The system packages to be installed, in bindep format. This may either
       be a filename, or a list of requirements (see below for an example).
 
 The following example uses filenames that contain various dependencies:
@@ -291,7 +291,7 @@ Specifies the base image to be used. At a minimum you *MUST* specify a source, i
 Valid keys for this section are:
 
     ``base_image``
-      A dictionary defines the parent image for the execution environment. A ``name``
+      A dictionary defining the parent image for the execution environment. A ``name``
       key must be supplied with the container image to use. Use the ``signature_original_name``
       key if the image is mirrored within your repository, but signed with the original
       image's signature key.
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -88,6 +88,8 @@ Configuration options
 
 You may use the configuration YAML keys listed here in your v3 execution environment definition file.
 
+.. _additional_build_files:
+
 additional_build_files
 ^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


Akasurde:
cc @acozine @samccann @Andersson007 @felixfontein

docs/porting_guide_v3.rst
```diff
@@ -0,0 +1,115 @@
+.. _porting_guide_v3:
+
+Ansible Builder version 3 Porting Guide
+=======================================
+
+This section discusses the behavioral changes between ``ansible-builder`` 1.2 and ``ansible-builder`` 3.0.
+
+It is intended to assist in updating your execution environment configuration files so they will work with this version of Ansible Builder.
```
felixfontein: This sounds as if ansible-builder 3.0 no longer supports the v1 and v2 config formats. Is that really the case? I thought (and I think CI runs showed before I upgraded them to use the v3 format) that the old formats are also supported.

docs/definition.rst
```diff
@@ -284,11 +291,13 @@ Specifies the base image to be used. At a minimum you *MUST* specify a source, i
 Valid keys for this section are:
 
     ``base_image``
-      A dictionary defining the parent image for the execution environment. A ``name``
+      A dictionary defines the parent image for the execution environment. A ``name``
```
felixfontein: I think this needs to stay 'defining'.

docs/definition.rst
```diff
@@ -204,7 +210,7 @@ The following keys are valid for this section:
             package_pip: https://github.com/example_user/ansible/archive/refs/heads/ansible.tar.gz
 
     ``ansible_runner``
-      The version of the Ansible Runner Python package to be installed. This value
+      The version of the Ansible Runner Python package is to be installed. This value
```
felixfontein: I think that if the `is` is added, another `that` needs to be added before it. I think it's better to leave both `is` and `that` away though. I'm not 100% sure though, English isn't my native language :)

(Same for the next change below.)

felixfontein (commented):
I haven't really looked at the porting guide itself yet.

docs/definition.rst
samccann: @felixfontein is correct. I'd suggest leaving it as is as a sentence fragment.

docs/definition.rst
```diff
@@ -234,10 +240,10 @@ The following keys are valid for this section:
       (``python_path``).
 
     ``system``
-      The system packages to be installed, in bindep format. This may either
+      The system packages are to be installed, in bindep format. This may either
```
samccann: Same comment. Basically, this is a sentence fragment, so doesn't have to follow the rules that a full sentence needs. So 'foo to be installed' is fine.  Or "Foo that is/are to be installed" also works.

docs/definition.rst
samccann: agree

samccann (commented):
What would be helpful here is a simple example of converting say a v2 ee definition file to a v3 definition file.

docs/porting_guide_v3.rst
Akasurde: I removed this line.

samccann (approved):
As I said in a prior comment, would be good to have a v2-v3 example in this porting guide but not worth holding up this PR for it.  LGTM

Shrews:
Maybe I'm totally missing the point, but the porting guide as proposed here just seems to be a watered down version of what is in `definition.rst`. What is the value add here?

samccann:
I did think that as well @Shrews - But it is at least a start (aka better than nothing). Since Builder v3 is significantly different, it does highlight at least that aspect.

That said, an alternate would be as I mentioned in the comments - an example or walkthrough of how to take a builder v1/v2 file and convert to v3.

Shrews:
@samccann Thanks for the reply. So it seems, hopefully, there will be more work in the future to improve upon this. In that case, I'll approve what we have here to get us moving forward on it.

Shrews (approved):

Issue #593, opened 2023-08-13
introspect.py - problem parsing requirements.txt - Ansible-Builder-3.0.0


spikefishjohn:
This is a cross post to a problem adding the collection paloaltonetworks.panos

[Warning: failed to parse requirements from paloaltonetworks.panos, error: not enough values to unpack (expected 1, got 0) #447](https://github.com/PaloAltoNetworks/pan-os-ansible/issues/447)

The core issue seems to be parsing the requirements.txt from the collection.

```bash
--> ec10a2c35f7
[3/4] STEP 12/13: RUN $PYCMD /output/scripts/introspect.py introspect --sanitize --user-pip=requirements.txt --write-bindep=/tmp/src/bindep.txt --write-pip=/tmp/src/requirements.txt
Warning: failed to parse requirements from paloaltonetworks.panos, error: not enough values to unpack (expected 1, got 0)
Creating parent directory for /tmp/src/requirements.txt
---
```

[requirements.txt from Palo repo](https://github.com/PaloAltoNetworks/pan-os-ansible/blob/develop/requirements.txt)

```bash
root@debian:~/TestIssue# cat Palo/requirements.txt.orig
pan-os-python==1.8.0 ; python_version >= "3.8" and python_version < "4.0" \
    --hash=sha256:843972122ef4c5bfd17354070e3ada1285e243325f5e7cf7fd969150048becd4 \
    --hash=sha256:da9153fe4a961a43d36016851ab06df45a37ea8612ed36b3b5b528f8394f34f5
pan-python==0.17.0 ; python_version >= "3.8" and python_version < "4.0" \
    --hash=sha256:9c074ea2f69a63996a6fefe8935d60dca61660e14715ac19d257ea9b1c41c6e2 \
    --hash=sha256:f4674e40763c46d5933244b3059a57884e4e28205ef6d0f9ce2dc2013e3db010
xmltodict==0.12.0 ; python_version >= "3.8" and python_version < "4.0" \
    --hash=sha256:50d8c638ed7ecb88d90561beedbf720c9b4e851a9fa6c47ebd64e99d166d8a21 \
    --hash=sha256:8bbcb45cc982f48b2ca8fe7e7827c5d792f217ecf1792626f808bf41c3b86051
root@debian:~/TestIssue#
```

I tried removing the "\\" and joining the lines in question and got the following output.

```bash
root@debian:~/TestIssue# cat Palo/requirements.txt
[...]

PR #596, opened 2023-08-16
 Supports line continuations for requirements.txt

akira6592:
#### SUMMARY

This PR allows [line continuations](https://pip.pypa.io/en/stable/reference/requirements-file-format/#line-continuations) by `\` for `requirements.txt`.

#### ISSUE TYPE

- Bugfix Pull Request

#### COMPONENT NAME

- src/ansible_builder/_target_scripts/introspect.py

### ADDITIONAL INFORMATION

related to:
  - https://github.com/ansible/ansible-builder/issues/593
  - https://github.com/PaloAltoNetworks/pan-os-ansible/issues/447

(Not a PR that solves all problems)

### DIFF
--- a/src/ansible_builder/_target_scripts/introspect.py
+++ b/src/ansible_builder/_target_scripts/introspect.py
@@ -28,7 +28,7 @@ def pip_file_data(path):
     pip_content = read_req_file(path)
 
     pip_lines = []
-    for line in pip_content.split('\n'):
+    for line in join_lines(pip_content.split('\n')):
         if line_is_empty(line):
             continue
         if line.startswith('-r') or line.startswith('--requirement'):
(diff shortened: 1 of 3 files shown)


Shrews:
Thanks for the PR.

However, I'm torn on whether or not to accept this and just demand a more rigid `requirements.txt` format for `ansible-builder`. Honestly, I'm leaning toward the latter to keep it simple for us. Looking at the pip code (https://github.com/pypa/pip/blob/49b9c0470f2a2a346293dd3bafc7c08f149db313/src/pip/_internal/req/req_file.py), they do a lot more processing of the requirements file that I'm not sure we should try to begin to duplicate, and we need to draw the line somewhere.

That being said, we are also aware that the requirements parsing library we are using is quite outdated and limited, and we need something better and more flexible. This is on our list to address in the future.

akira6592:
@Shrews 
Actually, I was wondering how far I should implement.

Eventually, I would like to support line continuations and parse the `--hash` option as well. For `paloaltonetworks.panos` collection [`requirements.txt`](https://github.com/PaloAltoNetworks/pan-os-ansible/blob/v2.17.3/requirements.txt).

FYI, the `pip-requirements-parser` seemed to support line continuations and `--hash` option.

nitzmahone:
Yeah, especially with `requirements-parser` looking somewhat dead, as a general policy I'm not feeling great about implementing a bunch of non-standard requirements grammar for even a subset of what pip supports- especially since we have to mash a bunch of these things together correctly in the end. Even something as seemingly simple as line-continuation actually complicates things we're planning, like being able to correlate a merged requirements entry to its source(s). I definitely understand the quality-of-life implications, but I also want to lean on standards for this stuff as much as is reasonably possible, and IIUC line continuation is not part of the defined standard.

akira6592:
@nitzmahone 
Thank you for the information about the situation. 

Hoping that fundamental measures will be taken, I think I should close this PR.

nitzmahone:
Closing per above

PR #598, opened 2023-08-20
fix default definition file name typo

w4hf:
replace the underscore with a '-'


Shrews (commented):
Thanks for spotting this. It seems we have one additional type of this filename in `docs/index.rst`. Would you mind correcting that one, too?

Shrews (approved):

PR #617, opened 2023-10-05
Update Galaxy definition in schema v3

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request


samccann:
@Akasurde - can you elaborate on this a bit? it's hard for me to tell if this is a schema change that should be reflected in docs or not (and if so - is this an enhancement to builder v3 schema or really a bugfix).

Shrews:
We really need to add a test for a full string input of the galaxy value to make certain that we don't break that. There is a good example in the awx-ee definition: https://github.com/ansible/awx-ee/blob/devel/execution-environment.yml#L12-L26

PR #627, opened 2023-10-20
Account for PEP668 in pip invocations

### LINKED ISSUE #604: Account for PEP668 in `pip` invocations during build
Newer versions of `pip` implement [PEP668](https://peps.python.org/pep-0668/) support, which will soon be enabled default (TL;DR, it will fail to install packages to system-managed Python environments, on which ansible-builder relies heavily).

Switching to virtual environments in EEs is not really an option for the short-medium term anyway; we need to continue to assume the ability to install any and all necessary packages in the system Python environment. 

Luckily it should be an easy fix; we just need to ensure we're setting `PIP_BREAK_SYSTEM_PACKAGES=1` in the environment for on all of builder's `pip` invocations; there shouldn't be any consequence to setting this for older pip versions, and it'll keep new ones functioning.

### LINKED ISSUE #646: ansible-builder version 3 has broken compatibility with schema version 1
The builder image is not being used in a v1 EE definition (neither the default value nor the explicitly defined value).

Example A:

```
---
version: 1
```

Example B:

```
---
version: 1
build_arg_defaults:
  EE_BASE_IMAGE: 'quay.io/ansible/ansible-runner:latest'
  EE_BUILDER_IMAGE: 'quay.io/ansible/ansible-builder:latest'
```

Both examples produce a `Containerfile` that will always use the base image for the builder image. E.g.,

```
...
# Builder build stage
FROM base as builder
...
```

It does not appear that we have any tests to cover this situation.

Shrews:
- Set the PIP_BREAK_SYSTEM_PACKAGES environment variable anywhere pip is in use to account for PEP668 which would change pip to not allow us to install in the system environment for newer versions of pip.
- Adds new `pip_install` target script to encapsulate the pip install logic. This will be called to install pip in the `base` image, and thus inherited by the other images. For v1 EE schemas, we _always_ copy this script to the `builder` image and call it since it will be separate from `base`. For v2, we only copy `pip_install` to the builder image and call it when one is defined.
- Remove `ensurepip` calls from inside target scripts so we can control pip installation external to those scripts. Also, those calls are unnecessary since we install pip in the `base` image (caveat: see the builder image hoops outlined in previous step).
- Add new `skip_pip_install` option to EE schema to manage pip installation.
- BONUS fix for issue #646 which was preventing adding a somewhat decent test for this feature.

Fixes #604 
Fixes #646

### DIFF
--- a/src/ansible_builder/containerfile.py
+++ b/src/ansible_builder/containerfile.py
@@ -127,7 +127,7 @@ class Containerfile:
         # Second stage (aka, builder): assemble (pip installs, bindep run)
         ######################################################################
 
-        if self.definition.builder_image:
+        if self.definition.builder_image or self.definition.version == 1:
             # Note: A builder image can be specified only in V1 or V2 schema.
             image = "$EE_BUILDER_IMAGE"
         else:
@@ -147,7 +147,9 @@ class Containerfile:
         if image == "base":
             self.steps.append("RUN $PYCMD -m pip install --no-cache-dir bindep pyyaml requirements-parser")
         else:
-            # For < v3 with a custom builder image, we always make sure pip is available.
+            # For an EE schema earlier than v3 with a custom builder image, we always make sure pip is available.
+            context_dir = Path(self.build_outputs_dir).stem
+            self.steps.append(f'COPY {context_dir}/scripts/pip_install /output/scripts/pip_install')
             self.steps.append("RUN /output/scripts/pip_install $PYCMD")
 
         self._insert_custom_steps('prepend_builder')
--- a/src/ansible_builder/containerfile.py
+++ b/src/ansible_builder/containerfile.py
@@ -108,7 +108,6 @@ class Containerfile:
                 "",
                 "# Galaxy build stage",
                 "FROM base as galaxy",
-                "ENV PIP_BREAK_SYSTEM_PACKAGES=1",
             ])
 
             self._insert_global_args()
--- a/src/ansible_builder/containerfile.py
+++ b/src/ansible_builder/containerfile.py
@@ -106,6 +106,7 @@ class Containerfile:
                 "",
                 "# Galaxy build stage",
                 "FROM base as galaxy",
+                "ENV PIP_BREAK_SYSTEM_PACKAGES=1",
             ])
 
             self._insert_global_args()
@@ -135,6 +136,7 @@ class Containerfile:
             "",
             "# Builder build stage",
             f"FROM {image} as builder",
+            "ENV PIP_BREAK_SYSTEM_PACKAGES=1",
             "WORKDIR /build",
         ])
 
@@ -156,6 +158,7 @@ class Containerfile:
             "",
             "# Final build stage",
             "FROM base as final",
+            "ENV PIP_BREAK_SYSTEM_PACKAGES=1",
         ])
 
         self._insert_global_args()
--- a/src/ansible_builder/containerfile.py
+++ b/src/ansible_builder/containerfile.py
@@ -76,6 +76,7 @@ class Containerfile:
             "# Base build stage",
             "FROM $EE_BASE_IMAGE as base",
             "USER root",
+            "ENV PIP_BREAK_SYSTEM_PACKAGES=1",
         ])
 
         self._insert_global_args()
@@ -87,7 +88,7 @@ class Containerfile:
                 self.steps.append(step)
 
             # We should always make sure pip is available for later stages.
-            self.steps.append('RUN $PYCMD -m ensurepip')
+            self.steps.append('RUN $PYCMD -m ensurepip --root /')
 
             if self.definition.ansible_ref_install_list:
                 self.steps.append('RUN $PYCMD -m pip install --no-cache-dir $ANSIBLE_INSTALL_REFS')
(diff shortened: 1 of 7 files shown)


nitzmahone:
This looks good, but was just playing with it while testing some other core PEP668 work Matt C and I were doing and tripped over a nasty problem that's only apparent when actually running in a PEP668-marked environment: it doesn't look like `ensurepip` can respect the envvar, and will thus refuse to work at all.

It doesn't look like any RHELish OSs are marking things yet (only recent Debian/Ubuntu), but it's easy to fake it up for testing using the following as your base image:

``` Dockerfile
FROM fedora:39
RUN touch /usr/lib64/python3.12/EXTERNALLY-MANAGED
```

Build that locally, then use the resultant image as your base in a vanilla EE with the default Python settings, and it'll :bomb: hard on `ensurepip`. It [explicitly blocks](https://github.com/python/cpython/blob/v3.12.0/Lib/ensurepip/__init__.py#L111-L120) inheritance of all `PIP_` envvars when pip is launched to bootstrap itself, and there's no other apparent way to pass that through.

So either we'll need to [look up](https://peps.python.org/pep-0668/#marking-an-interpreter-as-using-an-external-package-manager), zap (and possibly re-create?) the marker file, not use `ensurepip`, or ... something else.

Shrews:
Well that's a bunch of poo...  For reference, this builder EE is enough show the error:

```
---
version: 3

images:
  base_image:
    name: docker.io/library/fedora:39

additional_build_steps:
  prepend_base:
    - RUN touch /usr/lib64/python3.12/EXTERNALLY-MANAGED
```

sivel:
I have no idea if this is intended behavior, but you can also bypass `EXTERNALLY-MANAGED` by just supplying `--root`, with either `pip` or `ensurepip`.  If you set `--root /` that works just like we do now.

```console
[root@8f8a1523eade /]# touch /usr/lib64/python3.12/EXTERNALLY-MANAGED
[root@8f8a1523eade /]# python3 -m ensurepip
error: externally-managed-environment

× This environment is externally managed
╰─> The Python environment under /usr is managed externally, and may not be
    manipulated by the user. Please use specific tooling from the distributor of
    the Python installation to interact with this environment instead.


note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
[SNIP]
[root@8f8a1523eade /]# python3 -m ensurepip --root /
Looking in links: /tmp/tmpn6uk_n0x
Processing /tmp/tmpn6uk_n0x/pip-23.2.1-py3-none-any.whl
Installing collected packages: pip
Successfully installed pip-23.2.1
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
```

sivel:
Ok, using `--root` is expected to bypass the check:

https://github.com/pypa/pip/blob/9685f64fe8c78e1e39cd9b32e5615f42e0a01f1c/src/pip/_internal/commands/install.py#L270-L274

```
        # Check whether the environment we're installing into is externally
        # managed, as specified in PEP 668. Specifying --root, --target, or
        # --prefix disables the check, since there's no reliable way to locate
        # the EXTERNALLY-MANAGED file for those cases. An exception is also
        # made specifically for "--dry-run --report" for convenience.
```

sivel (approved):

Akasurde (approved):
LGTM

nitzmahone:
I'm going to verify one more thing to be sure, but after messing around with `ensurepip --root /`, I think it's basically working for us by accident. I didn't really like adding `ensurepip` in the first place, but it's kind of a necessary evil for common vanilla base images that wouldn't require users to always manually add a `prepend_base` and figure out how to get pip in there.

Much as I hate to say it, I think we're going to want to hide the `ensurepip` stuff behind a reusable script indirection where we can do a little more work to make sure we've got a pip available that can do what we want. The risk is that that script indirection becomes another place where everyone wants to hang "magic" to silently make pip work for $random_base_image_with_unique_busted_pip, but I think it's preferable to the game of distributed whac-a-mole I can see this one becoming down the road :laughing: .

Shrews:
Per internal discussion, going to move the ensurepip call to a new target script to encapsulate future changes that might be needed for non-RHEL distros, and add a flag to disable the call.

src/ansible_builder/containerfile.py
```diff
@@ -106,6 +108,7 @@ def prepare(self) -> None:
                 "",
                 "# Galaxy build stage",
                 "FROM base as galaxy",
+                "ENV PIP_BREAK_SYSTEM_PACKAGES=1",
```
nitzmahone: this seems superfluous at a glance, since it's always set on base (even for pre-v3)

nitzmahone (approved):
Might be able to move/remove the re-declarations of `ENV PIP_BREAK_SYSTEM_PACKAGES=1` (one looks unnecessary, the other only for pre-v3), but LGTM!

src/ansible_builder/containerfile.py
Shrews: Agreed! Will remove it before merge.

PR #631, opened 2023-11-10
Allow the .yaml file extension on the default EE file

### LINKED ISSUE #619: [Feature request]: add `execution-environment.yaml` to file lookup during the build
According to [YAML recommendations](https://yaml.org/faq.html) it would be great to support `yaml` file extension during the [lookup mechanism](https://ansible.readthedocs.io/projects/builder/en/stable/usage/#the-build-command), for example, load files in that sequence:
1) `execution-environment.yaml`
2) `execution-environment.yml`

Many modern tools respect this, including Ansible, but ansible-builder does not support this yet.

Shrews:
Allows the `.yaml` file extension on the default EE filename for use with the `ansible-builder` CLI command.

This also allows for use of the new extension for a collection `meta` EE file.

Fixes #619


src/ansible_builder/cli.py
```diff
@@ -153,7 +153,6 @@ def add_container_options(parser):
     for p in [create_command_parser, build_command_parser]:
 
         p.add_argument('-f', '--file',
-                       default=constants.default_file,
                        dest='filename',
                        help='The definition of the execution environment (default: %(default)s)')
```
Akasurde:
```suggestion
                       help='The definition of the execution environment (default: execution-environment.(yml|yaml))')
```

Akasurde (approved):
Rest LGTM

src/ansible_builder/cli.py
Akasurde: Without - 

```
  -f FILENAME, --file FILENAME
                        The definition of the execution environment (default:
                        None)
```

Shrews: Ah, good spot.

PR #634, opened 2023-12-04
introspect: add ability to specify requirements right in meta/execution-environment.yml

Andersson007:
In https://ansible.readthedocs.io/projects/builder/en/latest/collection_metadata/#collection-level-dependencies is written:
```
The meta/execution-environment.yml file containing the Python and/or bindep requirements
or referencing other files listing them. The .yaml extension is also valid on this file.
```
The `containing the Python and/or bindep requirements` part of the statement is not true. I tried and it throws a type error.
It would be nice if it works as it's convenient and it would allow not to pollute collection repos with additional files.
For the context, see https://forum.ansible.com/t/collection-inclusion-requirements-add-requirements-txt-and-bindep-txt/2611/8 (thanks @felixfontein for spotting this).

This PR allows to specify dependencies right in `meta/execution-environments.yml`. Tested locally and it works as expected.

NOTE: if you're OK with the idea and approve it in general, I'll try to fix the unit tests.

### DIFF
--- a/src/ansible_builder/_target_scripts/introspect.py
+++ b/src/ansible_builder/_target_scripts/introspect.py
@@ -63,14 +63,22 @@ def process_collection(path):
     CD = CollectionDefinition(path)
 
     py_file = CD.get_dependency('python')
-    pip_lines = []
-    if py_file:
-        pip_lines = pip_file_data(os.path.join(path, py_file))
+    # When requirements specified right in meta/execution-environment.yml:
+    if isinstance(py_file, list):
+        pip_lines = py_file
+    else:
+        pip_lines = []
+        if py_file:
+            pip_lines = pip_file_data(os.path.join(path, py_file))
 
     sys_file = CD.get_dependency('system')
-    bindep_lines = []
-    if sys_file:
-        bindep_lines = bindep_file_data(os.path.join(path, sys_file))
+    # When requirements specified right in meta/execution-environment.yml:
+    if isinstance(sys_file, list):
+        bindep_lines = sys_file
+    else:
+        bindep_lines = []
+        if sys_file:
+            bindep_lines = bindep_file_data(os.path.join(path, sys_file))
 
     return (pip_lines, bindep_lines)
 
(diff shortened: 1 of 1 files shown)


Andersson007:
ready for review (i'll fix unit tests but I need your approval first)

Shrews (commented):
I believe we are generally ok with this idea, but this is going to need some new tests to make sure it works as expected. Something to also consider is that the inline content for the non-collection EE file may also be a string. For example:

```
dependencies:
    python: |
        six
        paramiko
    system: |
        git-core [platform:rpm]
        python3.9-devel [platform:rpm compile]
```

If we are going to start expanding the schema of the collection EE, it would be nice if it could go through schema validation (`ee_schema.py`). Right now, in its current form, the validation is easy... the referenced file either exists or not. If we are going to expand it, more validation should be done.

Shrews:
After some internal discussion, maybe we can relax the strict schema validation suggestion a bit (though it would be nice to catch obvious format errors, I suppose).

Just curious, but how do you see collections dealing with using this new EE format but still trying to support folks stuck using an older version of `ansible-builder` that does not support it? Would collection authors just force folks to use older versions of the collection?

felixfontein:
Two thoughts of mine:
1. If this is added to ansible-builder now, I would wait a few years before using it anywhere, to make sure that most users (hopefully) have an ansible-builder version supporting this by then.
2. I'm not sure what older ansible-builder versions say when trying to build an EE with such a collection. Do they ignore the dependencies, or do they fail? What's the error message?
3. It might be a good idea to have versioning (similar to `version: <int>` in the EE definition), so that ansible-builder can give a nicer error message when encountering EE metadata it doesn't yet support. But even if this is added now, it again takes years so that most ansible-builder installations have this code included...

felixfontein:
(Some strange definition of "two" I have...)

Andersson007:
@Shrews @felixfontein thanks a lot for the feedback!

Also worth mentioning is that the file naming `execution-environment.yml` and `meta/execution-environment.yml` can be confusing and their fields names kinda intersect which makes it even more confusing. That's why someone can expect them work similarly in terms of options to pass the values as path or list.

However, with that doc PR merged  yesterday (thank you once more for reviewing) and taking your concerns into consideration, I think I'll just close the PR in order to keep things safe for users. I could open an issue instead, let me know. Thanks

Issue #644, opened 2024-01-29
requirements-parser breaks on 3.12+ default venvs


nitzmahone:
ansible-builder fails to run in "stock" Python 3.12+ venvs/environments. The `requirements-parser` Python dep used by `introspect.py` has an undeclared dependency on `setuptools`, since it relies on `pkg_resources` being present. This has arguably been broken for a long time, but is exacerbated by Python 3.12+ venvs explicitly excluding `setuptools` by default.

We don't really want a runtime dep on setuptools anyway, so this is another nail in the coffin of `requirements-parser`. We might be able to get what we need from `packaging` instead, but the way we inject and run `introspect` inside the container env makes me consider vendoring the needed bits instead (possibly `requirements-parser` itself).

```
mdavis@mdavis-p1:~/.virtualenvs$ python3.12 -m venv ~/.virtualenvs/tmp-312
mdavis@mdavis-p1:~/.virtualenvs$ source ~/.virtualenvs/tmp-312/bin/activate
(tmp-312) mdavis@mdavis-p1:~/.virtualenvs$ pip list

# NOTE LACK OF SETUPTOOLS

Package Version
------- -------
pip     23.2.1

[notice] A new release of pip is available: 23.2.1 -> 23.3.2
[notice] To update, run: pip install --upgrade pip
(tmp-312) mdavis@mdavis-p1:~/.virtualenvs$ pip install ansible-builder
Collecting ansible-builder
...
(tmp-312) mdavis@mdavis-p1:~/.virtualenvs$ ansible-builder
ModuleNotFoundError: No module named 'pkg_resources'
```

DorBreger:
problem for me on arch linux

DorBreger:
workaround if installed through pipx is `pipx inject ansible-builder setuptools`

ssbarnea:
IMHO, the "we do not want" as something for the future, for the moment I hotfix would be to add setuptools as a dependency at least for py312.

The fix https://github.com/ansible/ansible-builder/pull/659 was merged 3 months ago but no release was made since, so users will still face it.

Issue #472, opened 2023-04-13
Correctly excluding a package?


cidrbl0ck:
While this is not a bug, or I guess a real issue, I am trying to determine how to exclude a package.

The long story is: I'm needing to build a custom EE to include WinRM, PSRM, and the community.windows collection. Everything I read mentions to use the ee-supported-rhel8 as a BASE image. However building fails due to :

```
error: No package matches 'openshift-clients'
Error: error building at STEP "RUN assemble": error while running runtime: exit status 1
```
Since I don't use Openshift I'd like to exclude it.  Reading here: https://access.redhat.com/solutions/6999217. it says to format your context/ContainerFile as such:

```
ARG EE_BASE_IMAGE=registry.redhat.io/ansible-automation-platform-21/ee-supported-rhel8
ARG EE_BUILDER_IMAGE=registry.redhat.io/ansible-automation-platform-21/ansible-builder-rhel8

FROM $EE_BASE_IMAGE as galaxy
ARG ANSIBLE_GALAXY_CLI_COLLECTION_OPTS=
USER root

ADD _build /build
WORKDIR /build


FROM $EE_BUILDER_IMAGE as builder
ADD _build/requirements.txt requirements.txt
RUN ansible-builder introspect --sanitize --user-pip=requirements.txt --write-bindep=/tmp/src/bindep.txt --write-pip=/tmp/src/requirements.txt
RUN assemble

FROM $EE_BASE_IMAGE
ARG PKGMGR_OPTS="--nodocs --setopt=install_weak_deps=0 --setopt=rhel-8-for-x86_64-appstream-rpms.excludepkgs=ansible-core"
USER root
COPY --from=builder /output/ /output/
RUN /output/install-from-bindep && rm -rf /output/wheels
```

However this doesn't work as I'm sure it's a syntax problem with Ansible-Builder versus Podman. I'm an AAP shop so all I know is ansible-builder.
I'm looking for information on how to exclude `openshift-clients` and any other un-needed package so I can build this image successfully.

Shrews:
There is no built-in mechanism to builder to explicitly _ignore_ a Python requirement.

I will note that the `openshift-clients` package doesn't seem to be a thing on PyPI:  https://pypi.org/project/openshift-clients

However, the `openshift-client` package does exist: https://pypi.org/project/openshift-client

[...]

PR #645, opened 2024-01-30
Drop requirements-parser in favor of pep508

sivel:
This PR is not backwards compatible, as it changes the expectations of what a "requirement" are, from the pip definition, to the definition dictated by pep508. Although it should be mentioned that we never gauranteed full support of pip requirements.txt, so with that in mind, this is technically not a breaking PR, but should be noted that it _could_ break some uses.

See:
* https://github.com/ansible/ansible-builder/issues/644

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,9 @@ jobs:
           - name: '3.11'
             tox_env: integration-py311
 
+          - name: '3.12'
+            tox_env: integration-py312
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -172,6 +175,9 @@ jobs:
 
           - name: '3.11'
             tox_env: unit-py311
+
+          - name: '3.12'
+            tox_env: unit-py312
 
     steps:
       - name: Checkout
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ max-line-length=160
 include_package_data = true
 install_requires =
     PyYAML
-    requirements_parser
+    packaging
     bindep
     jsonschema
 python_requires = >=3.9
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,6 +21,7 @@ classifiers =
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
     Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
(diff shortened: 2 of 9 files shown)


sivel:
In a way, I still prefer this PR over https://github.com/ansible/ansible-builder/pull/647

Even through the alternative means we have to care less, I feel like it opens us up to more issues in the long run.

And I think the dep on `packaging` introduced here, even though we're still keeping an extra dep, also helps us with more advanced functionality for the "escape hatch" we want as part of https://github.com/ansible/ansible-builder/issues/472

webknjaz:
> from the pip definition, to the definition dictated by pep508.

They are supposed to be the same: https://pip.pypa.io/en/stable/reference/requirement-specifiers/. Do you know of any differences?

sivel:
> They are supposed to be the same: https://pip.pypa.io/en/stable/reference/requirement-specifiers/. Do you know of any differences?

Here are a few things that work in a pip `requirements.txt` that aren't `pep508`:

```
-r another-requirements.txt
--extra-index-url http://pypi.example.org/simple
https://github.com/ansible/ansible/archive/refs/heads/devel.zip#egg=ansible
git+ssh://github.com/ansible/ansible.git@devel#egg=ansible
```

webknjaz:
@sivel PEP 508 talks about individual package specifiers, not `requiremets.txt`. Requirements has always been a pip-specific format. requirements version specs are still supposed to be PEP 508 compliant, it just allows pip CLI args in addition to that, which doesn't contradict the spec that only focuses on the specifiers: https://pip.pypa.io/en/stable/reference/requirements-file-format/#structure.

webknjaz:
A fresh idea: could also implement a draft version of [PEP 735](https://peps.python.org/pep-0735/) and be done with it. It's a way forward in the wider ecosystem, anyway.

Shrews:
FWIW, this PR still allows use of `-r` or `--requirement` within the requirements file (we handle that option by recursively including file contents before handing off to `pip`). One of the tests for that is modified by this change, but I verified it still works (test might need to revert back).

So the advantage of this PR (that I can see) is just the 1-to-1 replacement of `requirements-parser` with `packaging`. We still have the issue of folks wanting support of more pip-specific options, which we've pushed back on so far.

sivel:
Closing in favor of https://github.com/ansible/ansible-builder/pull/663

PR #647, opened 2024-02-05
No sanitize pip

sivel:
An alternative to #645 

This PR just makes no attempt at deduping anything, and removes the sanitization function completely

A few notes about this PR:

1. The `--sanitize` option still exists, but is hidden, and it does nothing.
2. As a result of making `--sanitize` a noop, the "default" screen format has changed, to more closely match what `--sanitize` displayed
3. `EXCLUDE_REQUIREMENTS` only applied to pip before, simply because `sanitize_requirements` was only ever used for pip, but it now applies to system as well. We could make `EXCLUDE_REQUIREMENTS` keyed, and separate them, or put it behind an explicit toggle to the function call
4. This may push errors down deeper, as we don't care what the line is, so it would just be passed on to pip


src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -1,16 +1,16 @@
...
 import sys
 import yaml
 
-import requirements
-
-
 base_collections_path = '/usr/share/ansible/collections'
 logger = logging.getLogger(__name__)
 
+REQ_NORM_RE = re.compile(r'[-_.]+')
```
sivel, line 11: https://peps.python.org/pep-0503/#normalized-names

Shrews: Thanks for the explanation. Was a head scratcher for me.

sivel:
Closing in favor of https://github.com/ansible/ansible-builder/pull/663

PR #662, opened 2024-03-20
Suggest adt installation

audgirka:
Suggest ADT installation

Jira-https://issues.redhat.com/browse/AAP-19818

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -22,9 +22,9 @@ Install from PyPi
 
 .. note::
 
-   An **alternative** approach to install ``ansible-builder`` is using the ``ansible-dev-tools`` package.
+   An **alternative** approach to installing ``ansible-builder`` is using the ``ansible-dev-tools`` package.
    `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_ is a single Python package that includes all necessary tools to
-   set up a development environment, generate new collections, build and test the content consistently, resulting in robust automation.
+   set up a development environment, generate new collections, build and test the content.
 
    .. code-block:: shell
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -22,9 +22,7 @@ Install from PyPi
 
 .. note::
 
-   An **alternative** approach to installing ``ansible-builder`` is using the ``ansible-dev-tools`` package.
-   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_ is a single Python package that includes all necessary tools to
-   set up a development environment, generate new collections, build and test the content.
+   An **alternative** approach to installing ``ansible-builder`` is using the ``ansible-dev-tools`` package. `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_ is a single Python package that includes all necessary tools to set up a development environment, generate new collections, build and test the content.
 
    .. code-block:: shell
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -13,6 +13,22 @@ Requirements
 - The ``--container-runtime`` option must correspond to the containerization tool you use.
 - ``ansible-builder`` version ``3.x`` requires Python ``3.9`` or higher.
 
+Recommended Installation Method
+********************************
+
+The **recommended** approach to install `ansible-builder` is using the
+`ansible-dev-tools` package.
+`Ansible Development Tools (ADT) https://ansible.readthedocs.io/projects/dev-tools/`
+aims to streamline the setup and usage of several tools needed in order to
+create `Ansible https://www.ansible.com` content. ADT combines critical Ansible
+development packages into a unified Python package.
+
+.. code-block:: shell
+
+   # This also installs ansible-core if it is not already installed
+   $ pip3 install ansible-dev-tools
+
+
 Install from PyPi
 *****************
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -17,7 +17,7 @@ Requirements
 
    The **recommended** approach to install ``ansible-builder`` is using the
    ``ansible-dev-tools`` package.
-   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_
+   | `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_
    aims to streamline the setup and usage of several tools needed in order to
    create `Ansible <https://www.ansible.com>`_ content. ADT combines critical Ansible
    development packages into a unified Python package.
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -13,28 +13,26 @@ Requirements
 - The ``--container-runtime`` option must correspond to the containerization tool you use.
 - ``ansible-builder`` version ``3.x`` requires Python ``3.9`` or higher.
 
-Recommended Installation Method
-********************************
+.. note::
 
-The **recommended** approach to install `ansible-builder` is using the
-`ansible-dev-tools` package.
-`Ansible Development Tools (ADT) https://ansible.readthedocs.io/projects/dev-tools/`
-aims to streamline the setup and usage of several tools needed in order to
-create `Ansible https://www.ansible.com` content. ADT combines critical Ansible
-development packages into a unified Python package.
+   The **recommended** approach to install ``ansible-builder`` is using the
+   ``ansible-dev-tools`` package.
+   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_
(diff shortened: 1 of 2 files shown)


ssbarnea:
@audgirka This  is not rendering correct, see https://ansible--662.org.readthedocs.build/projects/builder/en/662/installation/ correct the RST syntax.

docs/installation.rst
```diff
@@ -13,6 +13,22 @@ Requirements
 - The ``--container-runtime`` option must correspond to the containerization tool you use.
 - ``ansible-builder`` version ``3.x`` requires Python ``3.9`` or higher.
 
+Recommended Installation Method
+********************************
+
+.. note::
```
bcoca: make it a note or a header with a paragraph, a header for a note is confusing, looks like someone forgot to populate the text

docs/installation.rst
```diff
@@ -13,6 +13,22 @@ Requirements
 - The ``--container-runtime`` option must correspond to the containerization tool you use.
 - ``ansible-builder`` version ``3.x`` requires Python ``3.9`` or higher.
 
+Recommended Installation Method
```
bcoca: you are not really adding a 'recommended install method' .. but pointing to a package that already includes it with it's own installation methods. This is at the very least not well phrased.

docs/installation.rst
```diff
@@ -13,6 +13,20 @@ Requirements
...
+   ``ansible-dev-tools`` package.
+   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_
+   aims to streamline the setup and usage of several tools needed in order to
+   create `Ansible <https://www.ansible.com>`_ content. ADT combines critical Ansible
+   development packages into a unified Python package.
+
+   .. code-block:: shell
+
+      # This also installs ansible-core if it is not already installed
+      $ pip3 install ansible-dev-tools
```
ssbarnea: It might make sense to also fix the other 2 examples of using `pip`, and use the same executable `pip3`, which seems to be the default one (the simple one might be missing on some platforms), same applies to using python.

sivel (changes requested):
Hi, thank you for your interest in ansible-builder.

We've looked over this proposed change, however we would like to see some of the wording changed.

Instead of recommending this as _**the**_ preferred way to install ansible-builder, we believe it should be presented as just an additional option. The reasoning for the decision making as to why a user would want to use this installation option, that includes many more tools than just builder should be explained, and also indicated what audience may be best served by doing so.  However, this should probably be a link to the `ansible-dev-tools` repo or documentation, where that information can be explained in more detail.

audgirka:
> Hi, thank you for your interest in ansible-builder.
> 
> We've looked over this proposed change, however we would like to see some of the wording changed.
> 
> Instead of recommending this as _**the**_ preferred way to install ansible-builder, we believe it should be presented as just an additional option. The reasoning for the decision making as to why a user would want to use this installation option, that includes many more tools than just builder should be explained, and also indicated what audience may be best served by doing so. However, this should probably be a link to the `ansible-dev-tools` repo or documentation, where that information can be explained in more detail.

Will look at the wording changes for the "recommended" approach. 
We've tried to keep the same message across the other projects included as part of the `adt` package.

Line 20 onwards explains briefly the use case for `ansible-dev-tools`
https://github.com/ansible/ansible-builder/pull/662/files#diff-d9b149498982c0663c3b7170398773361ed5678f1a627e9c2fd8d2c955c563dbR20

We do have a [link](https://github.com/ansible/ansible-builder/pull/662/files#diff-d9b149498982c0663c3b7170398773361ed5678f1a627e9c2fd8d2c955c563dbR20) in place for `adt` docs which can help explain user the purpose of using this package and tools included as part of it.

@ssbarnea @cidrblock ^^

docs/installation.rst
audgirka: updated

cidrblock:
Let's update the wording per @sivel request so we can get this in.  If it needs to be revised further in the futre we can do that at a later date.

docs/installation.rst
```diff
@@ -13,12 +13,27 @@ Requirements
 - The ``--container-runtime`` option must correspond to the containerization tool you use.
 - ``ansible-builder`` version ``3.x`` requires Python ``3.9`` or higher.
 
+.. note::
+
+   The **recommended** approach to install ``ansible-builder`` is using the
+   ``ansible-dev-tools`` package.
```
oraNod:
```suggestion
   Installing ``ansible-builder`` with the ``ansible-dev-tools`` package is a convenient option that offers several benefits over a standalone installation.
```

docs/installation.rst
```diff
@@ -13,12 +13,27 @@ Requirements
...
+
+   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_
+   aims to streamline the setup and usage of several tools needed in order to
+   create `Ansible <https://www.ansible.com>`_ content. ADT combines critical Ansible
+   development packages into a unified Python package.
```
oraNod:
```suggestion
   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_ is a single Python package that contains all the tools you need to create high quality Ansible content.
   The ADT package helps you quickly set up a development environment, generate new collections, and then build and test to ensure consistent and robust automation.
```

docs/installation.rst
oraNod: The original wording seems vague to me. Who recommends installing the ADT package over standalone builder?

docs/installation.rst
oraNod: I'd get more to the point and tell the reader what ADT is from the start. Also address the reader more directly through the 2nd person singular ("you").

Rather than ADT on its own as an acronym, I'd suggest "the ADT package" to avoid ambiguity.

oraNod (commented):
Left some suggestions. Feel free to take them, leave them, or modify them however you think. Hope this helps!

ssbarnea (approved):

ssbarnea:
@sivel Please take another look, that should be the text we agreed on ("alternative"), renders ok, it has the note and no extra heading.

You might want to retrigger the failed job, which fails for unrelated reason. I cannot do it myself.

docs/installation.rst
```diff
@@ -18,8 +18,18 @@ Install from PyPi
 
 .. code-block:: shell
 
-   $ pip install ansible-builder
+   $ pip3 install ansible-builder
 
+.. note::
+
+   An **alternative** approach to install ``ansible-builder`` is using the ``ansible-dev-tools`` package.
```
oraNod:
```suggestion
   An **alternative** approach to installing ``ansible-builder`` is using the ``ansible-dev-tools`` package.
```

docs/installation.rst
```diff
@@ -18,8 +18,18 @@ Install from PyPi
...
+   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_ is a single Python package that includes all necessary tools to
+   set up a development environment, generate new collections, build and test the content consistently, resulting in robust automation.
```
oraNod:
```suggestion
   set up a development environment, generate new collections, build and test the content.
```

oraNod: In the current form it's not clear if "consistently" modifies "build and test the content" or if it also modifies "set up a development environment" and "generate new collections".

If the former, then "consistently build and test the content". If the latter, then "consistently set up a development environment, generated new collections, build and test the content."

Also ", resulting in robust automation" looks like a dangling participle. If you want to keep it, maybe consider ", which results in robust automation."

docs/installation.rst
oraNod: Do we really want to put "alternative" in bold text? This is already in a note to catch user attention. I'm not sure what emphasizing "alternative" is meant to convey here tbh.

ssbarnea (approved):

oraNod (approved):

docs/installation.rst
```diff
@@ -18,8 +18,18 @@ Install from PyPi
...
 .. code-block:: shell
 
-   $ pip install ansible-builder
+   $ pip3 install ansible-builder
 
+.. note::
+
+   An **alternative** approach to installing ``ansible-builder`` is using the ``ansible-dev-tools`` package.
+   `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_ is a single Python package that includes all necessary tools to
+   set up a development environment, generate new collections, build and test the content.
```
sivel:
```suggestion
   An **alternative** approach to installing ``ansible-builder`` is using the ``ansible-dev-tools`` package. `Ansible Development Tools (ADT) <https://ansible.readthedocs.io/projects/dev-tools/>`_ is a single Python package that includes all necessary tools to set up a development environment, generate new collections, build and test the content.
```

Just consolidating to a single line, removing some unecessary line breaks

sivel (approved):

PR #663, opened 2024-03-21
Expose a way to exclude deps - option 2

sivel:
Based on #647 

```yaml
---
version: 3.1

images:
  base_image:
    name: quay.io/centos/centos:stream9

dependencies:
  python_interpreter:
    package_system: python3.11
    python_path: /usr/bin/python3.11

  ansible_core:
    package_pip: ansible-core

  ansible_runner:
    package_pip: ansible-runner

  galaxy:
    collections:
      - name: community.docker
      - name: ansible.netcommon

  exclude:
    python:
      - docker
    system:
      - python3-Cython 
```

TODO:

- [x] python excludes should work
- [x] excludes should not apply to user deps
- [x] system excludes should work
- [ ] should there be a means to exclude collections? 

NOTES:

- Only supports exclusion of first level deps, does not offer a way to exclude deps of deps


src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -203,11 +216,24 @@ def simple_combine(reqs):
                 continue
 
             base_line = line.split('#')[0].strip()
+            name_match = REQ_NAME_RE.match(base_line)
```
sivel, line 219: Some additional care is probably required here.  This will have issues currently with things like:

* URLs (`git+...`, `https://...`)
* `-r extra-requirements.txt` or any other random pip flag

In this case, we probably just need to skip the exclude checks. But we can't have it both ways, in that we can't ignore requirements parsing, and still do requirements parsing, so URLs and such just wouldn't be supported. 🙂 

The regex is also using the pep503 info, but I can't think of reasons it wouldn't work with RPM package names.

sivel: This may not actually be a problem with `-r` but would still apply to URLs

src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -203,11 +216,24 @@ def simple_combine(reqs):
...
+                continue
+            if name in EXCLUDE_REQUIREMENTS and collection not in {'user', 'exclude'}:
+                logger.debug('# Excluding requirement %s from %s', name, collection)
+                continue
+
             if base_line in consolidated:
                 i = consolidated.index(base_line)
-                fancy_lines[i] += f', {collection}'
+                if not name_only:
+                    fancy_lines[i] += f', {collection}'
```
sivel, line 231: We need to kill off fancy lines, and just spit back out a concatenation, but not consolidate duplicates.

src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -1,16 +1,17 @@
...
 
-import requirements
-
-
 base_collections_path = '/usr/share/ansible/collections'
 logger = logging.getLogger(__name__)
 
+# https://peps.python.org/pep-0503/#normalized-names
+REQ_NORM_RE = re.compile(r'[-_.]+')
+REQ_NAME_RE = re.compile(r'^([-\w.]+)')
```
sivel, line 13:
```suggestion
REQ_NAME_RE = re.compile(r'^([^-][-\w.]+)')
```

src/ansible_builder/ee_schema.py
```diff
@@ -390,16 +391,31 @@
...
     schema_version = 1
     if 'version' in ee_def:
         try:
-            schema_version = int(ee_def['version'])
+            schema_version = float_or_int(ee_def['version'])
         except ValueError as e:
             raise DefinitionError(f"Schema version not an integer: {ee_def['version']}") from e
 
-    if schema_version not in (1, 2, 3):
+    if schema_version not in (1, 2, 3, 3.1):
```
nitzmahone, line 418: Since we're abusing a combination of other tools in the absence of a first-class YAML "schema" for EE definitions , it's probably better to just support this in the existing version 3 schema, along with a small docs note that the exclude feature requires Builder >= 3.1 (and will be an error on < 3.1).

nitzmahone (changes requested):
This is the best approach that seems the least likely to break existing usages, while still allowing for various "escape hatch" impls... Probably a few more bits to implement to fully close the loop:
* A `from_collection` key for `exclude`, to allow a list of one or more collection names from which to completely ignore all deps. It's a big hammer, but would make things like "I only want to get my deps from downstream-packaged sources" at least possible without requiring EE creators to manually merge and exclude all the deps using the `python/system/galaxy` keys.
* A docs note that Python requirements in collections *should* be limited to features defined by PEP508, and that any deviation from that will be considered undefined/unsupported behavior, even if it happens to work today.
* A suggestion to partner eng that PEP508-only requirements validation might be a good candidate for a certification gate (they could easily base such a thing off #645).

Shrews:
Closing this one and continuing work in https://github.com/ansible/ansible-builder/pull/664 so as not to annoy @sivel with noise on this one.

Issue #562, opened 2023-06-09
document ansible-builder introspect command


samccann:
This command doesn't show up in the CLI Usage chapter yet.

```
ansible-builder introspect -h
usage: ansible-builder introspect [-h] [--sanitize] [--user-pip USER_PIP] [--user-bindep USER_BINDEP] [--write-pip WRITE_PIP]
                                  [--write-bindep WRITE_BINDEP] [-v {0,1,2,3}]
                                  [folder]

Loops over collections in folder and returns data about dependencies. This is used internally and exposed here for verification. This is targeted
toward collection authors and maintainers.

positional arguments:
  folder                Ansible collections path(s) to introspect. This should have a folder named ansible_collections inside of it.

options:
  -h, --help            show this help message and exit
  --sanitize            Sanitize and de-duplicate requirements. This is normally done separately from the introspect script, but this option is
                        given to more accurately test collection content.
  --user-pip USER_PIP   An additional file to combine with collection pip requirements.
  --user-bindep USER_BINDEP
                        An additional file to combine with collection bindep requirements.
  --write-pip WRITE_PIP
                        Write the combined pip requirements file to this location.
  --write-bindep WRITE_BINDEP
                        Write the combined bindep requirements file to this location.
  -v {0,1,2,3}, --verbosity {0,1,2,3}
                        Increase the output verbosity, for up to three levels of verbosity (invoked via "--verbosity" or "-v" followed by an
                        integer ranging in value from 0 to 3) (default: 2)
```

PR #664, opened 2024-03-26
Expose a way to exclude dependencies

### LINKED ISSUE #201: Feature to ignore/sanitize some bindep requirements

### LINKED ISSUE #207: Allow for top-level EE denylist

### LINKED ISSUE #319: Private python requirements impossible to install due to outdated requirements-parser
When trying to install private python requirements the outdated library requirements-parser is in the way (Probably since it had not been 

Our requirements.txt contain a couple lines in the beginning which are allowed for pip:

```
--index-url=https://some-host.example.com/repository/pypi-proxy/simple
--extra-index-url=https://some-host.example.com/repository/custompy/simple
--trusted-host=some-host.example.com
```

[Both of the open PRs](https://github.com/davidfischer/requirements-parser/pulls) on that repo would fix parts of the required settings.

A couple Links on that:
- [requirements-parser: This project is actively seeking maintainers](https://github.com/davidfischer/requirements-parser/issues/45)
- [requirements-parser: Usage of pip directly?](https://github.com/davidfischer/requirements-parser/issues/50)

Not sure which path to go but that is currently pretty much stopping us from using ansible-builder.

### LINKED ISSUE #323: Option to disable --sanitize / override collection dependencies

### LINKED ISSUE #472: Correctly excluding a package?

### LINKED ISSUE #493: Dealing with broken collection deps
There will be inevitable cases where a collection declares incorrect, conflicting, or just-plain-broken dependencies- as of builder 3.0rc1, there's no way to just ignore those deps to allow the EE build to continue.

We don't want to get into the business of transforming, constraining, or substituting build dependencies, but providing a per-collection mechanism for "I don't want to consider part of this collection's declared deps" seems like a useful (and sometimes necessary) escape hatch.

Since we're just passing through the Galaxy `requirements.yml`, the most obvious method (annotating the requirements inline) is probably a non-starter, as we'd have to filter and re-generate the inputs to Galaxy in that case. So we're left with declaring that somewhere else in the EE def.

Perhaps something like the following:
``` yaml
dependencies:
  galaxy: |
    collections:
    - ansible.utils
    - community.windows
options:
  dependency_control:
    - collection: ansible.utils
      ignore_bindep: true
      ignore_python: true
```

Could get fancier with the ability to mask individual requirements, ignore cross-collection dependencies, ignore all...

We'd take this information into account as we generate unified requirements files, dropping data from individual sources as requested. Global "never install this package" kinds of things probably aren't possible, but something like this would at least allow the installation of a broken collection version in conjunction with manual addition of the necessary deps.

### LINKED ISSUE #644: requirements-parser breaks on 3.12+ default venvs

### LINKED ISSUE #671: Add support for pip specifier syntax like `pywinrm[credssp]==0.4.3`
In our project we need to install the _pywinrm_ package. As described on https://pypi.org/project/pywinrm/ , we have added it to _requirements.txt_ with the following syntax:

```
pywinrm[credssp]==0.4.3
```

When we run `ansible-builder build`, we see that this line has been simplified to:

```
pywinrm==0.4.3
```

which means that the `requests-credssp` dependency has been forgotten.

Browsing the ansible-builder's code a bit, we found that it's introspect.py's `sanitize_requirements` function that does this.

https://github.com/ansible/ansible-builder/blob/167ed5ca8b6f589d2c87d7930d3dab92e82351f8/src/ansible_builder/_target_scripts/introspect.py#L319

Would it be possible to add support for this standard pip syntax ?

Shrews:
Based on #647 

Changes
=======

- Removes `requirements_parser` Python dependency.
- Dependency sanitization functionality is removed.
    - The `--sanitize` option still exists, but is hidden, and it does nothing.
    - `EXCLUDE_REQUIREMENTS` only applied to pip before, simply because sanitize_requirements was only ever used for pip, but it now applies to system as well.
    - Most Python `requirements.txt` entries are simply pushed down to `pip` (`-r` still causes recursive inclusion inline).
- Ability to exclude top-level Python or system dependencies from collections.
    - New `exclude` schema keyword is added to version `3` EE format and requires `ansible-builder` version `3.1` or higher. Using an older builder version simply errors.
    - Only supports exclusion of first level deps, does not offer a way to exclude deps of deps.
    - Exclusion matching is done with the _simple_ name of the requirement (no versions, etc.).
    - Exclusions can be either straight string matching, or an advanced form using regular expressions.

Bug Fixes
=======

- Comment parsing fix to no longer strip the `#` anchor from requirement URLs (e.g., `git+https://git.repo/some_pkg.git#egg=SomePackage`)

Example
======

```yaml
---
version: 3

images:
  base_image:
    name: quay.io/centos/centos:stream9

dependencies:
  python_interpreter:
    package_system: python3.11
    python_path: /usr/bin/python3.11

  ansible_core:
    package_pip: ansible-core

  ansible_runner:
    package_pip: ansible-runner

  galaxy:
    collections:
      - name: community.docker
      - name: ansible.netcommon

  exclude:
    python:
      - docker
    system:
      - python3-Cython
    all_from_collection:
      - a.b
      - c.d
```

TODO:

- [x] python excludes should work
- [x] excludes should not apply to user deps
- [x] system excludes should work
- [x] a means to exclude all deps from a collection (`all_from_collection` or `from_collection`?)
- [x] A docs note that Python requirements in collections should be limited to features defined by PEP508, and that any deviation from that will be considered undefined/unsupported behavior, even if it happens to work today.


sivel:
Here is a list of related issues:

Fixes:
* https://github.com/ansible/ansible-builder/issues/644
* https://github.com/ansible/ansible-builder/issues/493
* https://github.com/ansible/ansible-builder/issues/472
* https://github.com/ansible/ansible-builder/issues/364
* https://github.com/ansible/ansible-builder/issues/323
* https://github.com/ansible/ansible-builder/issues/319
* https://github.com/ansible/ansible-builder/issues/201

Impacts:
* https://github.com/ansible/ansible-builder/issues/562

kurokobo:
Hi, thanks for working on this!
Sorry for adding a comment before making enough test, but since this PR at this moment simply uses `\n` to separate lines for `requirements.txt` for collections (in `pip_file_data`) and process them line by line, I wonder if the collections' `requirements.txt` would behave unexpectedly if it contained line continuations with backslashes.

- Example collection: https://github.com/PaloAltoNetworks/pan-os-ansible/blob/develop/requirements.txt
- Docs about line continuations in requirement files: https://pip.pypa.io/en/stable/reference/requirements-file-format/#line-continuations

sivel:
> I wonder if the collections' requirements.txt would behave unexpectedly if it contained line continuations with backslashes

@kurokobo that is correct.  We discussed this when architecting the functionality, and should a package that is excluded have a backslash with a newline, it will likely cause problems.

We don't plan to add any specific functionality to permit or deny newlines, but we do intend to add documentation that states that it is in the best interest of everyone for collection maintainers, and EE creators to use [pep508](https://peps.python.org/pep-0508/) formatted requirements files.  We may be going so far to add linting rules, potentially via `galaxy-importer` to warn when a collection defines non-pep508 formatted requirements files.

kurokobo:
@sivel 
Thanks for providing detailed background! I have no objection to the policy, but I want to know details a bit.

If we strictly follow pep508, even any comments (`#`) would not be allowed. But this PR does not remove an implementation that intentionally ignores comments; is it a policy that requires strictly forrowing pep508, but allows comments as an exception? Or is my understanding incorrect?

F.Y.I., to support line continuations, we need just one modification (I don't think it will be adopted, but I'll write it down as I think of it😅 ).

```diff
diff --git a/src/ansible_builder/_target_scripts/introspect.py b/src/ansible_builder/_target_scripts/introspect.py
index 553cf81..c32d121 100644
--- a/src/ansible_builder/_target_scripts/introspect.py
+++ b/src/ansible_builder/_target_scripts/introspect.py
@@ -26,7 +26,7 @@ def read_req_file(path):
 
 
 def pip_file_data(path):
-    pip_content = read_req_file(path)
+    pip_content = read_req_file(path).replace('\\\n', '')
 
     pip_lines = []
     for line in pip_content.split('\n'):
```

Shrews:
> @sivel Thanks for providing detailed background! I have no objection to the policy, but I want to know details a bit.
> 
> If we strictly follow pep508, even any comments (`#`) would not be allowed. But this PR does not remove an implementation that intentionally ignores comments; is it a policy that requires strictly forrowing pep508, but allows comments as an exception? Or is my understanding incorrect?

Comments will be allowed. Any non-pep508 compliant entries are going to be passed through without change, unchecked and unverified. We plan to update the docs that using such non-compliant entries will be considered undefined/unsupported behavior.

src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -1,16 +1,20 @@
...
-
 base_collections_path = '/usr/share/ansible/collections'
 logger = logging.getLogger(__name__)
 
+# regex for a comment at the start of a line, or embedded with leading space(s)
+COMMENT_RE = re.compile(r'(?:^|\s+)#.*$')
+
+# https://peps.python.org/pep-0503/#normalized-names
+REQ_NORM_RE = re.compile(r'[-_.]+')
+REQ_NAME_RE = re.compile(r'^([^-][-\w.]+)')
```
Shrews: Using this as the deciding factor on which data we pass through untouched is not enough. For example, this `REQ_NAME_RE` regex will match:

```
git+https://git.repo/some_pkg.git#egg=SomePackage
```

and return `git`. If `git` happens to be in our exclude list, we'll lose this line and any others like it.

Shrews:
Removed association with issue #364. This PR would pass through unrecognized (and potentially invalid) requirements.

src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -1,17 +1,94 @@
...
+    # already present in image for py3 environments
+    'yaml', 'pyyaml', 'json',
+))
 
 
-base_collections_path = '/usr/share/ansible/collections'
 logger = logging.getLogger(__name__)
 
 
+class CollectionDefinition:
```
Shrews, line 36: This class didn't change, just moved within the file to a reasonable spot.

src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -1,17 +1,94 @@
...
+
+
+BASE_COLLECTIONS_PATH = '/usr/share/ansible/collections'
+
+
+# regex for a comment at the start of a line, or embedded with leading space(s)
+COMMENT_RE = re.compile(r'(?:^|\s+)#.*$')
+
+
+EXCLUDE_REQUIREMENTS = frozenset((
```
Shrews, line 20: This definition did not change. Only moved to the top of the file.

src/ansible_builder/_target_scripts/introspect.py
```diff
@@ -135,83 +258,92 @@ def has_content(candidate_file):
...
-        return os.path.join(
-            base_collections_path, 'ansible_collections',
-            namespace, name
-        )
 
-    def namespace_name(self):
-        "Returns 2-tuple of namespace and name"
-        path_parts = [p for p in self.reference_path.split(os.path.sep) if p]
-        return tuple(path_parts[-2:])
+def filter_requirements(reqs: dict[str, list],
```
Shrews, line 299: Renamed from `sanitize_requirements()`

docs/definition.rst
```diff
@@ -245,6 +260,38 @@ The following keys are valid for this section:
...
+      dependencies, nor will they apply to dependencies of dependencies (top-level only). Python dependency
+      exclusions should be a list of package names appearing under the ``python`` key name. System dependency
+      exclusions should be a list of system package names appearing under the ``system`` key name. If you
+      want to exclude *all* Python and system dependencies from one or more collections, supply the list
+      of collection names under the ``all_from_collections`` key.
+
+      The exclusion string should be the simple name of the requirement you want excluded. For example,
+      if you need to exclude the system requirement that appears as ``foo [!platform:gentoo]`` within
+      an included collection, then your exclusion string should be ``foo``. To exclude the Python
+      requirement ``bar == 1.0.0``, your exclusion string would be ``bar``.
```
Shrews: I question if using the "simple" name for exclusion is something we really want. I'm open to discussion on this.

docs/collection_metadata.rst
```diff
@@ -94,20 +92,28 @@ Then, if the ``ansible_collection`` directory is in your home directory, you can
...
 
 Python Dependencies
 ^^^^^^^^^^^^^^^^^^^
 
-Ansible Builder combines all the Python requirements files from all collections into a single file using the ``requirements-parser`` library. This library supports complex syntax, including references to other files.
+Ansible Builder combines all the Python requirements files from all collections into a single file.
+
+Certain package names are specifically *ignored* by ``ansible-builder``, meaning that Ansible Builder
+does not include them in the combined file of Python dependencies, even if a collection lists them as
+dependencies. These include test packages and packages that provide Ansible itself. The full list can
```
Akasurde, line 106:
```suggestion
dependencies. These include Ansible required packages and also test packages. The full list can
```

test/unit/test_introspect.py
```diff
@@ -83,3 +121,252 @@ def test_yaml_extension(data_dir):
         'python': {'test_collection.test_yaml_extension': ['python-six']},
         'system': {},
     }
+
+
+def test_sanitize_pep508():
```
Akasurde:
```suggestion
def test_filter_requirements():
```

Let us just drop `sanitize` word.

docs/collection_metadata.rst
Shrews: "Ansible required packages" doesn't apply here, IMO.

test/unit/test_introspect.py
Shrews: Updated in latest commit. Good catch.

src/ansible_builder/containerfile.py
```diff
@@ -144,7 +145,7 @@ def prepare(self) -> None:
         self._insert_global_args()
 
         if image == "base":
-            self.steps.append("RUN $PYCMD -m pip install --no-cache-dir bindep pyyaml requirements-parser")
+            self.steps.append("RUN $PYCMD -m pip install --no-cache-dir bindep pyyaml")
```
sivel: We need `packaging` in the container too correct?  Since it's working now, I assume it's being installed as a dep somewhere, but we probably want to explicitly include it here incase something changes with deps.

Shrews: Hrm, yeah, I think you might be right about that since introspect needs it.

Shrews: Done.

sivel (approved):

nitzmahone (approved):

djgoosen:
Hi there, unfortunately #664 appears to be breaking numerous previously known-good custom EE builds that I am seeing in a variety of different user environments recently, even those which had already been migrated to v3 spec files, etc. While I'm sure it was well-intentioned, if even a relatively simple custom EE for something like Palo Alto will no longer build if the paloaltonetworks.panos collection itself is given in its galaxy requirements alongside its underlying python requirements (pan-python and pan-os-python), which certainly appears to be the case in ansible-builder 3.1.0, then it probably should be reviewed. 

Based on the changelog, it seems like it was anticipated that it might "cause build issues in images with older versions of pip that cannot handle duplicate requirement entries". However, even on pip latest, this change (to expectations for the Ansible Builder community on how any duplicate Python dependencies will be handled and resolved at build time) is breaking things, and evidently without any definitive, prescriptive guidance so far on what exactly everyone should do going forward.

```
(ab300) $ pip3 --version
pip 24.2 from /Users/x/ab300/lib/python3.11/site-packages/pip (python 3.11)
(ab300) $ ansible-builder build -t palo-ee
Running command:
  podman build -f context/Containerfile -t palo-ee context
Complete! The build context can be found at: /Users/x/palo-ee/context
(ab300) $ ansible-builder --version
3.0.0
(ab300) $ deactivate
$ source ab310/bin/activate
(ab310) $ ansible-builder --version
3.1.0
(ab310) $ ansible-builder build -t palo-ee
Running command:
  podman build -f context/Containerfile -t palo-ee context
...showing last 20 lines of output...
++ export PATH
++ '[' -n '' ']'
++ '[' -z '' ']'
++ _OLD_VIRTUAL_PS1=
++ PS1='(venv) '
++ export PS1
++ '[' -n /bin/bash -o -n '' ']'
++ hash -r
+ '[' -f /tmp/src/upper-constraints.txt ']'
+ [[ -n '' ]]
+ install_wheels
+ '[' -f /tmp/src/build-requirements.txt ']'
+ '[' -f setup.py ']'
+ '[' -f /tmp/src/requirements.txt ']'
+ '[' '!' -f /output/requirements.txt ']'
+ /usr/bin/python3 -m pip install --cache-dir=/output/wheels -r /tmp/src/requirements.txt
WARNING: Running pip install with root privileges is generally not a good idea. Try `python3 -m pip install --user` instead.
ERROR: Double requirement given: pan-os-python (from -r /tmp/src/requirements.txt (line 13)) (already in pan-os-python==1.8.0 (from -r /tmp/src/requirements.txt (line 1)), name='pan-os-python')
Error: building at STEP "RUN /output/scripts/assemble": while running runtime: exit status 1


An error occurred (rc=1), see output line(s) above for details.
(ab310) $ 
```

Shrews:
Hello @djgoosen. In the future, it's better to open a new issue than to comment on a closed PR since we may not always notice these types of comments.

The `3.1` release was sort of a big update with some very necessary changes that were required in order to progress the project forward and eliminate some unmaintained external libraries. Backward compatibility with EEs that work with version `3.0` was never guaranteed, thus the new Y-release version.

In your particular case, upgrading `pip` within your EE will likely fix your problem. In your PR description, you are showing the `pip` version information for software _outside_ of the container image, but it is the version of `pip` _within_ the container image that matters (that is where Python packages are installed). You can either use a more recent base image with a newer `pip`, or use `additional_build_steps` to upgrade the version of `pip` within the container image.

If that does not resolve your issue, you also have the option of using the new `exclude` feature in `3.1` to ignore any dependencies from collections that may be causing you issues.

If both of those solutions fail for you, please open a new issue and provide the EE file where you are seeing the problem.

PR #676, opened 2024-05-15
Removing unused packaging files

Shrews:
These were added in PR #3.


Shrews:
Confirmed with @Spredzy that these are legacy files and another internal repo exists with the necessary packaging data.

Akasurde (approved):

PR #677, opened 2024-05-17
Support extra build command args

### LINKED ISSUE #668: Equivalent to podman build --pull?
`podman build` has a `--pull` option that sets pull policy for the base image when building a new image. By default, this is `missing`.

There appears to be no way to set or override this pull policy in ansible-builder. This means that if you have a base image of tag `:latest`, it will never be updated by builder after the initial pull.

Is there any way to get ansible-builder to update the base image when it runs? If not, I think a `--pull` option would be a useful addition to builder. This could be upgraded to `--pull-always` if container policy requires it, e.g. with `signature_required`.

I don't want to do a separate `podman pull` of the image beforehand, because that means I have to repeat the base image name in the schema and in the `podman pull`.

Shrews:
New CLI argument `--extra-build-cli-args` that acts as a pass-through to the `build` command.

Fixes #668

### DIFF
--- a/docs/scenario_guides/scenario_secret_passing.rst
+++ b/docs/scenario_guides/scenario_secret_passing.rst
@@ -3,9 +3,9 @@
 Passing Secrets
 ===============
 
-Sometimes it may be useful to use `build secrets <https://docs.docker.com/build/building/secrets/>`_ when
-creating an Execution Environment. This can be done with a combination of the use of :ref:`additional_build_steps`
-within the EE definition file, and the :ref:`extra-build-cli-args` CLI option.
+When creating an Execution Environment, it may be useful to use `build secrets <https://docs.docker.com/build/building/secrets/>`_.
+This can be done with a combination of the use of :ref:`additional_build_steps` within the EE definition file, and the
+:ref:`extra-build-cli-args` CLI option.
 
 Use the :ref:`extra-build-cli-args` CLI option to pass a build CLI argument that defines the secret:
 
--- a/docs/scenario_guides/scenario_secret_passing.rst
+++ b/docs/scenario_guides/scenario_secret_passing.rst
@@ -1,0 +1,32 @@
+.. _secret_passing:
+
+Passing Secrets
+===============
+
+Sometimes it may be useful to use `build secrets <https://docs.docker.com/build/building/secrets/>`_ when
+creating an Execution Environment. This can be done with a combination of the use of :ref:`additional_build_steps`
+within the EE definition file, and the :ref:`extra-build-cli-args` CLI option.
+
+Use the :ref:`extra-build-cli-args` CLI option to pass a build CLI argument that defines the secret:
+
+.. code::
+
+    ansible-builder build --extra-build-cli-args="--secret id=mytoken,src=my_secret_file.txt"
+
+Then, use a custom ``RUN`` command within your EE definition file that references this secret:
+
+.. code:: yaml
+
+    ---
+    version: 3
+
+    images:
+      base_image:
+        name: quay.io/centos/centos:stream9
+
+    additional_build_steps:
+      prepend_base:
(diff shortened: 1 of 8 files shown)


src/ansible_builder/main.py
```diff
@@ -240,6 +242,7 @@ def build_command(self) -> list[str]:
             if self.container_policy != PolicyChoices.IGNORE:
                 command.append('--pull-always')
 
+        command.extend(self.extra_build_cli_args.split())
```
sivel: This should probably be `shlex.split` instead, as to not naively split just on spaces, but account for shell escapes and quoting.

Shrews: done

sivel (approved):

docs/scenario_guides/scenario_secret_passing.rst
```diff
@@ -0,0 +1,32 @@
+.. _secret_passing:
+
+Passing Secrets
+===============
+
+Sometimes it may be useful to use `build secrets <https://docs.docker.com/build/building/secrets/>`_ when
+creating an Execution Environment. This can be done with a combination of the use of :ref:`additional_build_steps`
+within the EE definition file, and the :ref:`extra-build-cli-args` CLI option.
```
Akasurde:
```suggestion
When creating an Execution Environment, it might be useful to use `build secrets <https://docs.docker.com/build/building/secrets/>`_. This can be done with a combination of the use of :ref:`additional_build_steps`
within the EE definition file, and the :ref:`extra-build-cli-args` CLI option.
```

Just rephrase, not a blocker.

docs/scenario_guides/scenario_secret_passing.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Use the :ref:`extra-build-cli-args` CLI option to pass a build CLI argument that defines the secret:
```
Akasurde, line 10: @Shrews do you think we should a note here saying that - the secrets are passed in plain text and may be recorded in console logs so be careful using this parameter or something along these lines?

src/ansible_builder/cli.py
```diff
@@ -150,6 +150,11 @@ def add_container_options(parser):
         help='Squash layers in the final image (choices: %(choices)s). Defaults to "%(default)s". (podman only)'
     )
 
+    build_command_parser.add_argument(
+        '--extra-build-cli-args',
+        help='Extra arguments to pass to the container build CLI command',
```
Akasurde, line 155:
```suggestion
        help='Extra arguments to pass to the container build CLI command. Specify these arguments as a space-separated string. For example: "--no-cache --rm"',
```

docs/scenario_guides/scenario_secret_passing.rst
Shrews: Secrets are not being passed in plain text.

Akasurde: Ok. np.

src/ansible_builder/cli.py
Shrews: A comma-separated string is not the proper way to pass the args. It would look like this:

```
--extra-build-cli-args="--optA --optB -c"
```

Akasurde: Ok, then just add an example section.

docs/scenario_guides/scenario_secret_passing.rst
Shrews: Done

src/ansible_builder/cli.py
Shrews: Added an example to the usage docs.

Akasurde (approved):
LGTM

nitzmahone (approved):
Taking bets on how long before someone complains about the `shlex.split()` preventing passthru of a weirdly complex quoted thing, but since it's just a convenience anyway and they can always call the builder directly, it's probably not worth trying to be any fancier.

PR #685, opened 2024-06-20
Add Python 3.13 testing

Shrews:
Add testing environment for future Python 3.13 release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           - name: '3.12'
             tox_env: integration-py312
 
-          - name: '3.13.0-alpha - 3.13.0'
+          - name: '3.13'
             tox_env: integration-py313
 
     steps:
@@ -73,6 +73,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.py_version.name }}
+          allow-prereleases: true
 
       - name: Install tox
         run: |
@@ -107,7 +108,7 @@ jobs:
           - name: '3.12'
             tox_env: unit-py312
 
-          - name: '3.13.0-alpha - 3.13.0'
+          - name: '3.13'
             tox_env: unit-py313
 
     steps:
@@ -118,6 +119,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.py_version.name }}
+          allow-prereleases: true
 
       - name: Install tox
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,9 @@ jobs:
           - name: '3.12'
             tox_env: integration-py312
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +60,9 @@ jobs:
           - name: '3.12'
             tox_env: integration-py312
 
+          - name: '3.13.0-alpha - 3.13.0'
```
webknjaz: Nit: You can do

```suggestion
          - name: ~3.13.0-0
```

to achieve the same.

I personally prefer this syntax because it allows me to conditionally allow failures if there's a tilde in the spec: https://github.com/cherrypy/cheroot/commit/99dff70d15e189b922420cd03688f5c039bbb49e#diff-662dff05c7dfe6681c1007ae601e8573d413a03f9f9d53d951c2cb99caba4fd4R704.

webknjaz: Another way that some people like is to pass `allow-prereleases: true` to `actions/setup-python`. It's been there since v4 IIRC because people wanted to just have `3.13` in their configs select pre-release versions.

```suggestion
          - name: 3.13
```

webknjaz: Finally, it's also possible to have
```suggestion
          - name: 3.13-dev
```
as well.

tox.ini
```diff
@@ -22,19 +29,19 @@ commands =
...
 description = Run pulp integration tests
 commands =
     pytest -n auto -m "not serial" test/pulp_integration {posargs} {[shared]pytest_cov_args}
     pytest -n 0 -m "serial" test/pulp_integration {posargs} {[shared]pytest_cov_args}
 
-[testenv:integration{,-py39,-py310,-py311,-py312}]
+[testenv:integration{,-py39,-py310,-py311,-py312,-py313}]
 description = Run integration tests
 # rootless podman reads $HOME
 passenv =
```
webknjaz, line 45: This setting overrides the main test env. You may want to inherit from it to let the other var keep being bypassed here:
```suggestion
passenv =
    {[testenv]passenv}
```

.github/workflows/ci.yml
```diff
@@ -104,6 +107,9 @@ jobs:
           - name: '3.12'
             tox_env: unit-py312
 
+          - name: '3.13.0-alpha - 3.13.0'
```
webknjaz: same https://github.com/ansible/ansible-builder/pull/685/files#r1688458186

.github/workflows/ci.yml
Shrews: Ah ha! I learned something new. Thanks!

tox.ini
Shrews: Hrm, actually, turns out I don't need the `passenv = UNSAFE_PYO3_SKIP_VERSION_CHECK` at all since it is set within the tox environment and will always be passed from that point. I'll just remove that line.

webknjaz (approved):

PR #693, opened 2024-07-17
Update index.rst with base image OS info

### LINKED ISSUE #692: Docs: Recommend using a RHEL image
The [current docs](https://github.com/ansible/ansible-builder/blob/68707471bf690a10a362b21ab5eabd4732124e10/docs/index.rst?plain=1#L58) state

```
Choosing a base image
---------------------

You can use any base image you choose.
```

However in practice this is not true, as issues indicate (#636, #553).

This should change to show that RHEL images are essentially required.

onefourfive:
Add base image OS recommendations to resolve #692

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -55,7 +55,7 @@ execution environment. You can specify these items:
 Choosing a base image
 ---------------------
 
-You can use any base image you choose.
+You can use any base image you choose, as long as it's a RHEL-type image like CentOS, Fedora, or RHEL. These operating systems are directly supported.
 The smaller the base image, generally, the smaller the final image.
 However, to make Ansible Builder more efficient, you should know what packages, if any, are already installed on the base image you use.
 


docs/index.rst
```diff
@@ -55,7 +55,7 @@ execution environment. You can specify these items:
 Choosing a base image
 ---------------------
 
-You can use any base image you choose.
+You can use any base image you choose, as long as it's a RHEL-type image like CentOS, Fedora, or RHEL. These operating systems are directly supported.
```
Shrews, line 58: I think this reads slightly better.

```suggestion
The base image must be a RHEL image or one of the RHEL-based variants (CentOS, Fedora, etc.). Other distributions are not supported at this time.
```

Shrews (changes requested):

docs/index.rst
felixfontein: I wouldn't say "RHEL-based variants", since CentOS and Fedora are not based on RHEL (as Ubuntu is based on Debian). Maybe say something like "... or one of the similar distributions (CentOS, ..."?

kaplanlior: > I wouldn't say "RHEL-based variants", since CentOS and Fedora are not based on RHEL (as Ubuntu is based on Debian). Maybe say something like "... or one of the similar distributions (CentOS, ..."?

RPM based and compatible with RedHat eco system (that would just exclude OpenSUSE and SLES)

Shrews:
Closing in favor of PR #782

PR #695, opened 2024-07-24
Modify scripts/check_galaxy to return more descriptive error message(s)

### LINKED ISSUE #694: Ansible Galaxy check script error message may be somewhat misleading
In scripts/check_galaxy, it gives a ```  The 'ansible-galaxy' command is not found in the base image. ``` message if the return code of ```ansible-galaxy --version``` is not 0, which could be caused by other things, e.g. incorrect ansible.cfg even if ansible-galaxy is installed.

brookelew:
Check for both existence of ansible-galaxy and verify ansible-galaxy --version works as expected in execution environment to resolve #694 

- [x] ~Write tests?~


Shrews (commented):
Thanks for contributing!

Instead of having a 2-step check here, perhaps we could leave it as a single check (running `ansible-galaxy --version`, like before) and instead just change the first paragraph of the error message to something more generic, such as "there was an error running the ansible-galaxy command, please use the -vvv option to examine the build output for any errors".  Doesn't need to be that exact wording, but you get the idea, hopefully.

Also, yes, we should have a test for this, but your change has exposed a hole in our CI testing. The tests for these scripts live in the `pulp-integration` test subdirectory, but those tests are no longer run for reasons I won't go into, but I don't see any reason why those particular tests for these scripts can't be moved to our normal `integration` test subdirectory. I'll work on a fix for that for you so we can get this one moving along.

brookelew:
Thanks! I added some possible causes to the error message, but I'm not sure if that's necessary.

Shrews (commented):
I like this approach.

Once PR #696 merges, and after you rebase your change to pick it up, your change will break the new `test/integration/test_build.py::test_missing_ansible` integration test, but that will be easily fixable. And that would be the file where you can create a new integration test for this change using the tests in that PR as a model. You'll need to include an `ansible.cfg` that will break `ansible-galaxy` in the manner you expect.

brookelew:
Test is written, sorry for the merge instead of rebase, I didn't realise GitHub would do that automatically

test/data/v3/check_galaxy/ee-bad-ansible-cfg.yml
```diff
@@ -0,0 +1,25 @@
+version: 3
+
+images:
+  base_image:
+    name: quay.io/centos/centos:stream9
+
+dependencies:
+  ansible_core:
+    package_pip: ansible-core==2.15.12
```
Shrews: I think we can just remove the version requirement here.

test/data/v3/check_galaxy/ee-bad-ansible-cfg.yml
```diff
@@ -0,0 +1,25 @@
...
+  ansible_runner:
+    package_pip: ansible-runner==2.3.1
```
Shrews: Same here... leave off version.

test/integration/test_build.py
```diff
@@ -348,3 +348,20 @@ def test_missing_runner(cli, runtime, ee_tag, data_dir, tmp_path):
...
+    """
+    ee_def = data_dir / 'v3' / 'check_galaxy' / 'ee-bad-ansible-cfg.yml'
+
+    with pytest.raises(subprocess.CalledProcessError) as einfo:
+        cli(
+            f'ansible-builder build -c {tmp_path} -f {ee_def} -t {ee_tag} '
+            f'--container-runtime={runtime} -v3'
+        )
+
+    assert "ERROR - 'ansible-galaxy' command not functioning as expected" in einfo.value.stdout
```
Shrews, line 367: In addition to this `assert`, we should add another that checks for the actual error emitted from `ansible-galaxy` (visible in the output thanks to using `-v3`) to make sure it is the faulty `ansible.cfg` causing the error and not something else. In this case, looks like the error string we want begins with:

```
ERROR: Error reading config file (/etc/ansible/ansible.cfg): Source contains parsing errors:
```

Shrews (changes requested):
Very nice! Just a few comments inline.

Also, the example `ansible.cfg` contains way more info than is necessary. Typically in tests, you want the minimum amount of data that would cause the failure you are looking for to eliminate any chances of something else causing an error (either today or further in the future). Let's trim that down to just the essential to cause the failure.

Shrews:
Also, regarding the merge commit, I'm not 100% certain how the GH interface is going to handle that. We'd like to keep our history clean and linear, so if you could eliminate that for me in your PR, it would be much appreciated.

Shrews:
> Also, regarding the merge commit, I'm not 100% certain how the GH interface is going to handle that. We'd like to keep our history clean and linear, so if you could eliminate that for me in your PR, it would be much appreciated.

Actually, I think you can ignore this request, if you wish. The squash button should eliminate the merge commit for us.

test/integration/test_build.py
brookelew: Thanks for getting the exact line I needed!

brookelew:
If it's fixed by the squash I don't think I'll touch it, should there be a test for a completely missing ansible-galaxy install or is the incorrect ansible.cfg test fine?

Shrews:
Hi @brookelew 

Not sure if you saw, but your change is not passing our CI tests. You can see more info on the `Details` like next to each failing test. You have a linting problem, and the error message text you are asserting has changed since you modified the example `ansible.cfg`, so you'll need to account for that.

brookelew:
Oh no yeah I saw sorry was just a bit busy last week

Shrews (approved):

Shrews:
Thanks for your contribution @brookelew !

brookelew:
Thanks for supervising my first open source contribution @Shrews ! : D

PR #701, opened 2024-08-15
Add 3.1 porting guide

Shrews:
Restructures our current porting guide and adds a new porting guide entry for going from `3.0` to `3.1`.

### DIFF
--- a/docs/porting_guides/porting_guide_v3.1.rst
+++ b/docs/porting_guides/porting_guide_v3.1.rst
@@ -108,7 +108,7 @@ Environment file by adding these lines to it:
 
     additional_build_steps:
       append_base:
-        - RUN pip3 install -U pip
+        - RUN $PYCMD -m pip install -U pip
 
 That will upgrade ``pip`` to the latest version within the base image. To restrict the upgrade to a specific
 version of ``pip``, alter the upgrade command to specify that version. For example:
@@ -117,4 +117,4 @@ version of ``pip``, alter the upgrade command to specify that version. For examp
 
     additional_build_steps:
       append_base:
-        - RUN pip3 install -U pip==20.3
+        - RUN $PYCMD -m pip install -U pip==20.3
--- a/docs/porting_guides/porting_guide_v3.1.rst
+++ b/docs/porting_guides/porting_guide_v3.1.rst
@@ -1,0 +1,120 @@
+*********************************
+Ansible Builder 3.1 Porting Guide
+*********************************
+
+This section discusses the behavioral changes between ``ansible-builder`` version 3.0 and version 3.1.
+
+.. note::
+
+    We highly advise running ``ansible-builder`` with increased verbosity (``--v3``) to fully expose any error
+    messages that may help in diagnosing any problems.
+
+.. contents:: Topics
+
+Python Requirements Handling
+============================
+
+The 3.1 release significantly changes how Python and system requirements are handled by simplifying dependency
+parsing. This release removes the use of an external library that was unmaintained and parsed many types of
+Python dependencies either partially or completely incorrectly. The changes are described below.
+
+PEP 508 Standard
+----------------
+
+Python requirements files are expected to follow the `PEP 508 standard <https://peps.python.org/pep-0508/>`_.
+Builder will *expect* the requirements file to be in this format, but it makes two exceptions:
+
+#. Comments (lines beginning with ``#``) are ignored.
+#. Any line from the requirements file that is not compliant with PEP508 causes a warning to be omitted, and
+   the line is passed through to ``pip`` unmodified. It is recommended to *not* depend on this behavior as we
+   cannot guarantee that these lines will not cause issues later during the dependency installation process.
+
+The passthrough of lines not following PEP508 may cause issues that may not have been present with version 3.0
+since many of these lines were simply ignored and not passed to ``pip`` with that version.
+
+Dependency Sanitization
+-----------------------
+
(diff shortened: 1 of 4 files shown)


docs/porting_guides/porting_guide_v3.1.rst
```diff
@@ -0,0 +1,120 @@
...
+The solution requires upgrading ``pip`` within the base image to a version that contains an updated dependency resolver.
+Beginning with ``pip`` version 20.3, the dependency resolver can handle duplicate requirements whose versions do not
+conflict, so that version is the minimum required. Upgrade ``pip`` in the base image from within the Execution
+Environment file by adding these lines to it:
+
+.. code-block:: yaml
+
+    additional_build_steps:
+      append_base:
+        - RUN pip3 install -U pip
```
sivel: Should we make this `RUN $PYCMD -m pip install -U pip`?  To avoid issues where the `python` specified and `pip3` are for different python versions? Is that available at this phase?

Shrews: Yes, great point. Definitely available in the `append_base` phase.

sivel (approved):

docs/porting_guides/porting_guide_v3.1.rst
```diff
@@ -0,0 +1,120 @@
+*********************************
+Ansible Builder 3.1 Porting Guide
+*********************************
+
+This section discusses the behavioral changes between ``ansible-builder`` version 3.0 and version 3.1.
+
+.. note::
+
+    We highly advise running ``ansible-builder`` with increased verbosity (``--v3``) to fully expose any error
```
nitzmahone: Now that we support `-vvv`, should we use that here (and wherever else) for consistency with the rest of Ansible?

docs/porting_guides/porting_guide_v3.1.rst
```diff
@@ -0,0 +1,120 @@
...
+PEP 508 Standard
+----------------
+
+Python requirements files are expected to follow the `PEP 508 standard <https://peps.python.org/pep-0508/>`_.
+Builder will *expect* the requirements file to be in this format, but it makes two exceptions:
+
+#. Comments (lines beginning with ``#``) are ignored.
+#. Any line from the requirements file that is not compliant with PEP508 causes a warning to be omitted, and
+   the line is passed through to ``pip`` unmodified. It is recommended to *not* depend on this behavior as we
+   cannot guarantee that these lines will not cause issues later during the dependency installation process.
```
nitzmahone:
```suggestion
#. Any line from the requirements file that is not compliant with PEP508 causes a warning to be emitted and
   the line passed through to ``pip`` unmodified. It is not recommended to depend on this behavior, as we
   it can change suddenly between ``pip`` releases, and can cause other problems with dependency resolution.
```

docs/porting_guides/porting_guide_v3.1.rst
```diff
@@ -0,0 +1,120 @@
...
+
+The passthrough of lines not following PEP508 may cause issues that may not have been present with version 3.0
+since many of these lines were simply ignored and not passed to ``pip`` with that version.
```
nitzmahone:
```suggestion
The passthrough of non-PEP508 compliant lines may expose issues that were hidden by the,
version 3.0 dependency sanitizer, which often silently ignored and removed them.
```

docs/porting_guides/porting_guide_v3.1.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Dependency Sanitization
+-----------------------
+
+Dependency sanitization (the combining of duplicate dependencies into a single dependency entry) is no longer performed.
```
nitzmahone:
```suggestion
Dependency sanitization (the combining of duplicate dependencies into a single dependency entry) is no longer performed by ``ansible-builder``.
```

docs/porting_guides/porting_guide_v3.1.rst
```diff
@@ -0,0 +1,120 @@
...
+
+    ERROR: Double requirement given: netaddr>=0.10.1 (from -r /tmp/src/requirements.txt (line 13)) (already in netaddr (from -r /tmp/src/requirements.txt (line 4)), name='netaddr')
+    Error: building at STEP "RUN /output/scripts/assemble": while running runtime: exit status 1
+
+This error comes from ``pip`` within one of the intermediate container images when attempting to install the Python
+requirements from included collections and/or from the user supplied requirements. This intermediate image is based
+on the base image defined within the Execution Environment file, and it means that the version of ``pip`` installed
+within that image is too old to handle duplicate requirement entries.
+
+To determine where the duplicate requirements are coming from, run ``ansible-builder`` with the ``--v3`` option
```
nitzmahone:
```suggestion
To determine where the duplicate requirements are coming from, run ``ansible-builder`` with the ``-vvv`` option
```

docs/porting_guides/porting_guide_v3.1.rst
```diff
@@ -0,0 +1,120 @@
...
+to get more verbose output, then look for the output from the introspection phase. It will look similar to below:
```
nitzmahone:
```suggestion
to get more verbose output, then look for the output from the introspection phase. It will look similar to:
```

nitzmahone (commented):
A few suggested changes you can take or leave, but LGTM- thanks!

docs/porting_guides/porting_guide_v3.1.rst
Shrews: Considered that, but was afraid folks would try `-vvv` under < 3.1. I will mention both here.

PR #713, opened 2024-10-08
Improve execution of pip during container building

ssbarnea:
- Make pip less verbose as this makes harder to investigate container
  building due to being too verbose.
- Use a bind mount to prevent pip cache from being written inside the
  container but still make use of it. This will speed-up the build
  process considerably.

### DIFF
--- a/src/ansible_builder/containerfile.py
+++ b/src/ansible_builder/containerfile.py
@@ -94,7 +94,8 @@ class Containerfile:
                 self.steps.append('RUN /output/scripts/pip_install $PYCMD')
 
             if self.definition.ansible_ref_install_list:
-                self.steps.append('RUN $PYCMD -m pip install --no-cache-dir $ANSIBLE_INSTALL_REFS')
+                self.steps.append(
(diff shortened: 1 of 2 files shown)


src/ansible_builder/containerfile.py
```diff
@@ -94,7 +94,8 @@ def prepare(self) -> None:
                 self.steps.append('RUN /output/scripts/pip_install $PYCMD')
 
             if self.definition.ansible_ref_install_list:
-                self.steps.append('RUN $PYCMD -m pip install --no-cache-dir $ANSIBLE_INSTALL_REFS')
+                self.steps.append(
+                    'RUN --mount=type=cache,dst=/root/.cache/pip $PYCMD -m pip install -q $ANSIBLE_INSTALL_REFS')
```
Shrews, line 98: Using `-q` hides some install info that could be useful in debugging user issues. I'd prefer to not use this.

I did notice that it did subtract about 1.5 seconds on a very simple build, and the resulting image sizes did not seem to change by any significant amount.

Shrews (changes requested):

PR #714, opened 2024-10-08
Use newer github runners

ssbarnea:
The main reason for using the newer github runners is because they do have a newer version of podman, as the one from ubuntu-22.04 is ancient and missing some important features, like cache bind mounts.

Needed-By: #713

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   sanity:
     name: ${{ matrix.test.name }}
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     env:
       TOXENV: ${{ matrix.test.tox_env }}
 
@@ -17,7 +17,7 @@ jobs:
       matrix:
         test:
           - name: Lint
-            tox_env: linters-py310
+            tox_env: linters-py312
 
           - name: Docs
             tox_env: docs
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -26,10 +26,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Install Python 3.10
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
```
Shrews: No need to install 3.10. Just change `linters-py310` to `linters-py312`, unless that causes a bunch of unexpected linting errors we'll have to work out separately.

bindep.txt
```diff
@@ -3,3 +3,5 @@
 
 git
 python38-wheel [platform:centos-8 platform:rhel-8]
+tox [test platform:dpkg]
+python3-pip [test platform:dpkg]
```
Shrews: This shouldn't be necessary. This file probably needs to be removed from the repo (not in this change).

Shrews (changes requested):

.github/workflows/ci.yml
ssbarnea: Not sure about the builder itself but on other project we try to keep it to the minimum supported python as some linters or formatters might request or make incompatible if run with newer pythons. Even so, nowadays is much better, will do.

bindep.txt
ssbarnea: OK I will remove it but I find it quite useful for other projects. There is also the tox-extra plugin (which I maintain) which checks for bindep before running tox, saving time for first time contribs.

Shrews (approved):

PR #715, opened 2024-10-21
Replace ansible-runner with ansible-dev-tools image

fao89:
EE_BASE_IMAGE default was: quay.io/ansible/ansible-runner:latest which is a 2 years old image: https://quay.io/repository/ansible/ansible-runner?tab=tags&tag=latest

Now the default is: ghcr.io/ansible/community-ansible-dev-tools:latest which is the recommended image for execution environments according to https://github.com/ansible/creator-ee

### DIFF
--- a/src/ansible_builder/constants.py
+++ b/src/ansible_builder/constants.py
@@ -16,7 +16,7 @@ build_arg_defaults = {
     # empty string values here still allow the build arg to be emitted into the generated Containerfile
     'ANSIBLE_GALAXY_CLI_COLLECTION_OPTS': '',
     'ANSIBLE_GALAXY_CLI_ROLE_OPTS': '',
-    'EE_BASE_IMAGE': 'quay.io/ansible/ansible-runner:latest',
+    'EE_BASE_IMAGE': 'ghcr.io/ansible/community-ansible-dev-tools:latest',
     # this value is removed elsewhere for v3+ schemas
     'EE_BUILDER_IMAGE': 'quay.io/ansible/ansible-builder:latest',
     'PKGMGR_PRESERVE_CACHE': '',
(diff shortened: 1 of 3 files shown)


nitzmahone:
I thought we'd already made the base image required with v3, but apparently not. The original intent was to just remove the default altogether along with the sunsetting of the old schema, so unless you're claiming 100% compatibility with the old image (GLWT :wink:) to avoid breaking builds of v1/v2 EE defs, probably better to just leave it alone with the warning.

@Shrews is out this week, but maybe he remembers why we didn't just make that a hard error for the v3+ schema...

sivel:
Yeah, I agree that we shouldn't just replace the image.  v1 and v2 schemas shouldn't be used, and should be deprecated. We need to ensure we are preventing the default image from being used in v3

fao89:
should I close the  PR?

Shrews:
> I thought we'd already made the base image required with v3, but apparently not. The original intent was to just remove the default altogether along with the sunsetting of the old schema, so unless you're claiming 100% compatibility with the old image (GLWT 😉) to avoid breaking builds of v1/v2 EE defs, probably better to just leave it alone with the warning.
> 
> @Shrews is out this week, but maybe he remembers why we didn't just make that a hard error for the v3+ schema...

I don't recall us ever discussing making that a hard error for v3. I do remember us adding the warning about using the default with v3 and keeping the current default for backwards compatibility for older schemas.

I do think we should probably just close this PR and come up with a deprecation plan that eventually results in forcing the defining of a base image with v3. I think not having any default for v3+ is likely the best thing.

fao89:
I'm closing this PR as it seems it isn't the proper way to handle it

@nitzmahone @sivel @Shrews any recommendation for a base image for v3?

Shrews:
@fao89 We don't generally recommend specific base images. Any recent-ish RHEL-like base image will work smoothly for you (RHEL, Fedora, Centos, etc). Some good tips in the latest docs, too: https://ansible.readthedocs.io/projects/builder/en/latest/index.html#choosing-a-base-image

fao89:
thank you!

PR #717, opened 2024-10-31
Resolve Ansible Builder Errors for Podman on Windows

kevinshurtz:
Hey team!

I recently tried using Ansible Builder to create Ansible execution environments on my Windows machine with Podman, and I ran into a few issues.

1. By default, [Podman on Windows creates files with insufficient permissions to execute copied scripts](https://github.com/containers/podman/issues/15299).
2. The Containerfile includes Windows path separators in the `ENTRYPOINT` path, preventing the container from running.

These changes add the `--chmod` flag to the `COPY` commands in the Containerfile, ensuring the scripts run correctly and ensuring behavioral parity between Podman and Docker.  They also use the POSIX path separator when creating the `ENTRYPOINT` directive.

I've added unit tests to validate the path character behavior, and I've updated existing tests for the COPY directive.  After applying these changes, I was able to use Ansible Builder to create a container with a usable Ansible installation inside (using Fedora as the base image).

If you guys could incorporate these changes into the project when you get the chance, I would deeply appreciate it (and please let me know if there is anything else I can do)!

Sincerely,
Kevin

### DIFF
--- a/src/ansible_builder/containerfile.py
+++ b/src/ansible_builder/containerfile.py
@@ -149,7 +149,7 @@ class Containerfile:
         else:
             # For an EE schema earlier than v3 with a custom builder image, we always make sure pip is available.
             context_dir = Path(self.build_outputs_dir).stem
-            self.steps.append(f'COPY {context_dir}/scripts/pip_install /output/scripts/pip_install')
+            self.steps.append(f'COPY --chmod=755 {context_dir}/scripts/pip_install /output/scripts/pip_install')
             self.steps.append("RUN /output/scripts/pip_install $PYCMD")
 
         self._insert_custom_steps('prepend_builder')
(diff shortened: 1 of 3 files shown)


Shrews:
Thanks for the contribution!

Unfortunately, we don't really support Windows for builder, and do not have any testing for that platform in our CI. There are currently no plans to add support either. However, I'm totally fine with the addition of the `--chmod` portion of the change. If you care to remove the Windows-specific bits (mainly the new test, and stick to `os.path`), then I think I'd be ok with this change.

kevinshurtz:
> Thanks for the contribution!
> 
> Unfortunately, we don't really support Windows for builder, and do not have any testing for that platform in our CI. There are currently no plans to add support either. However, I'm totally fine with the addition of the `--chmod` portion of the change. If you care to remove the Windows-specific bits (mainly the new test, and stick to `os.path`), then I think I'd be ok with this change.

Hey @Shrews, thanks for the reply!

While I'm happy to remove the path-handling changes, if there is no intention to support Windows for Ansible Builder, what is the purpose of using `os.path` anywhere in the codebase?  As I understand, the purpose of this module is to select POSIX path separators on Unix-like systems and use Windows path conventions on Windows.

Since the path-handling changes in the PR merely force the usage of POSIX conventions when creating the Containerfile, it feels ironic and self-defeating to reintroduce Windows-specific path separators for the sole purpose of ensuring the project continues to misbehave on Windows.

If you would still prefer that I remove path-handling changes to ensure continued misbehavior on Windows, I'm of course still happy to do so.  Thank you again for the feedback, and please let me know if there is anything else I can share!

Sincerely,
Kevin

ibcht:
Hi, 

Bumping into this PR : I just ran into the same issue. I fixed manually by adding the expected --chmod args to COPY instructions as suggested. 

The build was successful otherwise, so it would be nice to have this PR merged. Even if builder is not meant to work on Windows, as it is already almost working, it would nicely open to new usages.

My use-case : running `ansible-builder create` to generate Containerfile and context files, then directly running podman build from the Windows host directly. Then I can manage the image directly into Podman Desktop, rather than using container in container, with image management scattered across my dev environment.

```bash
# Create Containerfile and context
podman run --rm -v /mnt/c/path/to/windows/project:/mnt/project ghcr.io/ansible/community-ansible-dev-tools:latest ansible-builder create -f /mnt/project/execution-environment.yml -c /mnt/project/ee-context

# Build from project directory
podman build  -t image-name:latest .\ee-context\
```

Issue #724, opened 2025-02-10
Allow to pass --ssh argument to 'podman build'


irozet12:
Please add --ssh argument for ansible-builder and pass it to 'podman build' command, if specified

Shrews:
We added the ability to pass arbitrary arguments in version `3.1` with the `--extra-build-cli-args` CLI arg ([docs here](https://ansible.readthedocs.io/projects/builder/en/latest/usage/#extra-build-cli-args)).

PR #725, opened 2025-02-10
Add --ssh key (#724)

irozet12:
Adding --ssh argument and passing it down to 'podman build' command


Shrews:
`--extra-build-cli-args` exists for this type of functionality. See https://ansible.readthedocs.io/projects/builder/en/latest/usage/#extra-build-cli-args

PR #727, opened 2025-02-19
Allow disabling text colors

### LINKED ISSUE #716: ansible-builder does not have option to disable console ANSI coloring
There are good use-cases (especially during testing) where user might want to disable ANSI coloring in console output.

I looked at the codebase and I was not able to find any option to disable the coloring. I would have expected at least one of the following environment variables to function:

- `ANSIBLE_NOCOLOR` (ansible)
- `ANSIBLE_FORCE_COLOR` (ansible)
- `FORCE_COLOR` (universal)
- `NO_COLOR` (universal)
- `PY_COLORS` (python)

There is some function in https://github.com/ansible/ansible-lint/blob/b1dd55d59eca5129e75be8df604a9268c166f911/src/ansiblelint/__main__.py#L498 that can be used to determine if coloring should be used or not, is very similar with one in core.

Shrews:
Allows control of when builder output text is colorized.

Bonus fix: This makes the Python 3.14 tests work.

Fixes #716 

Assisted-by: Cursor

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -203,6 +203,16 @@ will stack. For example, the following are equivalent to setting the verbosity l
    $ ansible-builder build -v -v -v
 
 
+``--no-colors``
+***************
+
+Disables ANSI text colors.
+
+If this option is not given, the default will be to enable text colors for the output.
+The ``NO_COLOR`` and ``FORCE_COLOR`` environment variables will be honored if this CLI option
+is not supplied.
+
+
 ``--prune-images``
 ******************
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -203,16 +203,6 @@ will stack. For example, the following are equivalent to setting the verbosity l
    $ ansible-builder build -v -v -v
 
 
-``--no-colors``
-***************
-
-Disables ANSI text colors.
-
-If this option is not given, the default will be to enable text colors for the output.
-The ``NO_COLOR`` and ``FORCE_COLOR`` environment variables will be honored if this CLI option
-is not supplied. If both environment variables are supplied, ``NO_COLOR`` will take precedence.
-
-
 ``--prune-images``
 ******************
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -210,7 +210,7 @@ Disables ANSI text colors.
 
 If this option is not given, the default will be to enable text colors for the output.
 The ``NO_COLOR`` and ``FORCE_COLOR`` environment variables will be honored if this CLI option
-is not supplied.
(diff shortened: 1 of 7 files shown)


nitzmahone (commented):
Overall looks okay, though I'd question if it warrants a dedicated CLI arg- for most things we guard that kind of addition pretty carefully.

I'm also curious how hard it is to get the build output to include color codes- my worry is that having this option could set an expectation that we're going to also filter or otherwise force-disable generation of color codes from the build, which could be problematic.

Shrews:
> Overall looks okay, though I'd question if it warrants a dedicated CLI arg- for most things we guard that kind of addition pretty carefully.

Ack. I can remove that bit.

 
> I'm also curious how hard it is to get the build output to include color codes- my worry is that having this option could set an expectation that we're going to also filter or otherwise force-disable generation of color codes from the build, which could be problematic.

Not sure I catch what you mean here...

webknjaz:
@Shrews there's a corner case in Rich interpreting `FORCE_COLOR` as "also produce ANSI-sequences for bold" with `NO_COLOR` as "only disable color-related ANSI but not bold".

This resulted in us settling on an additional variable `TTY_COMPATIBLE` that only supports `1` or `0` explicit values (auto-detect otherwise): https://github.com/Textualize/rich/issues/2924#issuecomment-2757673602 / https://github.com/Textualize/rich/pull/3675.

I'm wondering if this PR would benefit from also consulting that var.

Also, here's how `termcolor` handles these vars (it doesn't yet look into `TTY_COMPATIBLE` tho): https://github.com/termcolor/termcolor/blob/fd08149/src/termcolor/termcolor.py#L92-L125. In general, I think it'd be beneficial closely following whatever logic widely-used libs implement.

Shrews:
> @Shrews there's a corner case in Rich interpreting `FORCE_COLOR` as "also produce ANSI-sequences for bold" with `NO_COLOR` as "only disable color-related ANSI but not bold".
> 

I decided that simply supporting `NO_COLOR` is the simplest thing to do, and makes it much less confusing as to how it should operate.

Shrews:
NOTE: The `tox-dev` framework we use in CI sets `FORCE_COLOR=1`. If we were to support that option, it would need to be at a lower priority than `NO_COLOR`, because otherwise our tests (as they currently exist) would break with Python>=3.14. Although there [have been discussions](https://github.com/donatj/force-color.org/issues/8) about whether or not `NO_COLOR` should trump the other option, I do not see a consensus, or anything written in stone about that, except that, as of this writing, the code example on [force-color.org](https://force-color.org) still suggests that `FORCE_COLOR` is top priority. Leaving support for that option out of this change, at least for now, seems sensible.

sivel (approved):

PR #743, opened 2025-05-29
Allow downstreams to enforce that the ansible_core ref must have a version specifier

sivel:
This PR allows downstreams to patch `REQUIRE_ANSIBLE_CORE_PIN` in `constants.py` to a value of `True` to force EE definitions to contain a pin for the `ansible-core` specification.

This would help ensure that users building custom EEs would have to put thought into the version of ansible-core that they intend to be using, without unexpected upgrades in the building process pulling in a newer version they have not tested against.

No evaluation of the pin is done, this is a minor safety net.

### DIFF
--- a/src/ansible_builder/user_definition.py
+++ b/src/ansible_builder/user_definition.py
@@ -243,7 +243,7 @@ class UserDefinition:
     def _validate_ansible_core_ref(self):
         """
         If a downstream has patched REQUIRE_ANSIBLE_CORE_PIN validate
-        that the 'ansible_core' ref constains a version constraint.
+        that the 'ansible_core' ref contains a version constraint.
 
         :raises: DefinitionError exception if version constraint is invalid or missing
         """
--- a/src/ansible_builder/user_definition.py
+++ b/src/ansible_builder/user_definition.py
@@ -159,20 +159,7 @@ class UserDefinition:
 
     @property
     def ansible_core_ref(self):
-        ref = self.raw.get('dependencies', {}).get('ansible_core', {}).get('package_pip', None)
-
-        if ref and constants.REQUIRE_ANSIBLE_CORE_PIN:
-            try:
-                req = Requirement(ref)
-            except InvalidRequirement as e:
-                raise DefinitionError("Invalid package requirement specified for 'ansible_core'") from e
-
-            if not req.specifier:
-                raise DefinitionError(
-                    "Value for 'ansible_core' must contain a version constraint, such as 'ansible-core==2.16.*'"
-                )
-
-        return ref
+        return self.raw.get('dependencies', {}).get('ansible_core', {}).get('package_pip', None)
 
     @property
     def ansible_runner_ref(self):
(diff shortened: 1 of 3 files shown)


Shrews (changes requested):
I'd much rather see this validation done in the `UserDefinition.validate()` method where everything else is validated.

sivel:
> I'd much rather see this validation done in the `UserDefinition.validate()` method where everything else is validated

Done.

src/ansible_builder/user_definition.py
```diff
@@ -239,6 +240,30 @@ def _validate_additional_build_files(self):
             if dest.is_absolute() or '..' in dest.parts:
                 raise DefinitionError(f"'dest' must not be an absolute path or contain '..': {dest}")
 
+    def _validate_ansible_core_ref(self):
+        """
+        If a downstream has patched REQUIRE_ANSIBLE_CORE_PIN validate
+        that the 'ansible_core' ref constains a version constraint.
```
Shrews:
```suggestion
        that the 'ansible_core' ref contains a version constraint.
```

test/unit/test_user_definition.py
```diff
@@ -173,17 +173,35 @@ def test_v3_ansible_install_refs(self, exec_env_definition_file):
...
+                'ansible_runner': { 'package_pip': 'ansible-runner==2.3.1'}
+             }
+            }
+            """
+        )
+        monkeypatch.setattr(constants, 'REQUIRE_ANSIBLE_CORE_PIN', True)
+        definition = UserDefinition(path)
+        with pytest.raises(DefinitionError) as error:
+            definition.validate()
+        assert "Value for 'ansible_core' must contain a version constraint" in str(error.value.args[0])
```
Shrews, line 205:
```suggestion

    def test_v3_ansible_install_ref_bad_req(self, monkeypatch, exec_env_definition_file):
        path = exec_env_definition_file(
            """
            {'version': 3,
             'images': { 'base_image': {'name': 'base_image:latest'}},
             'dependencies': {
                'ansible_core': {'package_pip': 'ansible-core=2.6.0'},
                'ansible_runner': { 'package_pip': 'ansible-runner==2.3.1'}
             }
            }
            """
        )
        monkeypatch.setattr(constants, 'REQUIRE_ANSIBLE_CORE_PIN', True)
        definition = UserDefinition(path)
        with pytest.raises(DefinitionError) as error:
            definition.validate()
        assert "Invalid package requirement specified for 'ansible_core'" in str(error.value.args[0])
```

Shrews (commented):
Suggested an additional test for complete code coverage, and a spelling correction. Otherwise, LGTM.

sivel:
I've accepted the review changes, looks like there is at minimum a linting issue which I'll resolve though. I just want to note that the resolution of these things doesn't mean this needs merged.

This is an idea, which we ultimately need to decide whether we want it or not.

Shrews (approved):
LGTM, but not merging until we decide we actually want this.

PR #745, opened 2025-05-30
Simplify triage labeling

Shrews:
Simplify the triage labeling to just label every new issue and PR with `needs_triage`.


Akasurde (approved):

.github/issue_labeler.yml
```diff
@@ -1,2 +1,2 @@
 needs_triage:
-  - '.*'
+  - '/.*/'
```
webknjaz, line 2: I think this isn't a regex but an fnmatch expression. So if I got it right, this would be something like
```suggestion
  - >-
    **/*
```

Shrews: I wish there was some way to actually test it, but I was following the docs for the action that suggested that regex: https://github.com/marketplace/actions/regex-issue-labeler#label-all-issues

webknjaz: Ah, I confused it with another action.

webknjaz (approved):

Issue #749, opened 2025-06-20
Preserve Symlinks When Copying Additional Build Files


clickthisnick:
### Ansible Builder version

3.1.1.dev24+g6579c32a

### Python version

3.12

### Operating System and Environment

```console
RHEL8
```

### Container Runtime

Docker version 25.0.2, build 29cf629

### Summary

Currently, [shutil.copy2(source, dest)](https://github.com/ansible/ansible-builder/blob/devel/src/ansible_builder/utils.py#L203) is used to copy additional build files. However, this function follows symlinks by default (follow_symlinks=True), meaning it copies the target file's contents instead of the symlink itself. If the symlink points to a file that doesn't exist yet, this results in a `FileNotFoundError`.

_**Why This Matters**_
In our use case, we rely on symlinks to credentials that:

Are not present in the repo, but will be available at runtime (e.g., mounted into the container or downloaded by Ansible tasks).

By following symlinks at build time, ansible-builder tries to resolve them immediately—even though the target files may not exist yet—causing the build to fail unnecessarily.

Steps to Reproduce
```bash
mkdir -p ansible/src
cd ansible
ln -s /tmp/credentials src/credentials

cat <<EOF > execution-environment.yml
---
version: 3

additional_build_files:
  - src: src
    dest: src
EOF
```

ansible-builder create
This results in:

```vbnet
FileNotFoundError: [Errno 2] No such file or directory: 'src/credentials'
```

Suggestion
Change the call to shutil.copy2(source, dest, follow_symlinks=False) so that symlinks are preserved instead of dereferenced. 

This ensures:
* Symlinks are carried into the container as- is
* Failures are avoided when the symlink target isn't present at build time
* Updates to the symlink targets at runtime are reflected correctly

### Steps to reproduce

1.

```bash
mkdir -p ansible/src
cd ansible
ln -s /tmp/credentials src/credentials

cat <<EOF > execution-environment.yml
---
version: 3

additional_build_files:
  - src: src
    dest: src
EOF
```

2.

`ansible-builder create`

### Expected Results

Symlinks should be copied over.

### Actual Results

The content of the symlink is copied over and if the symlink points to a fiel that doesn't exist yet this results in a `FileNotFoundError`

clickthisnick:
Sending over a pull request

Shrews:
Fixed by PR #750

PR #750, opened 2025-06-20
Preserve Symlinks When Copying Additional Build Files

clickthisnick:
Fixes issue reported in https://github.com/ansible/ansible-builder/issues/749

### DIFF
--- a/src/ansible_builder/utils.py
+++ b/src/ansible_builder/utils.py
@@ -189,12 +189,12 @@ def copy_file(source: str, dest: str, ignore_mtime: bool = False) -> bool:
         raise Exception(f"Source {source} can not be a directory. Please use copy_directory instead.")
     if Path(dest).is_dir():
         raise Exception(f"Destination {dest} can not be a directory. Please use copy_directory instead.")
-    if not os.path.exists(dest):
-        logger.debug("File %s will be created.", dest)
-        should_copy = True
-    elif Path(source).is_symlink() and Path(dest).is_symlink() and os.path.realpath(source) == os.path.realpath(dest):
+    if Path(source).is_symlink() and Path(dest).is_symlink() and os.readlink(source) == os.readlink(dest):
         logger.debug("Symlink %s already exists and matches.", dest)
         should_copy = False
+    elif not os.path.exists(dest):
+        logger.debug("File %s will be created.", dest)
+        should_copy = True
     elif not filecmp.cmp(source, dest, shallow=False):
         logger.warning('File %s had modifications and will be rewritten', dest)
         should_copy = True
--- a/src/ansible_builder/utils.py
+++ b/src/ansible_builder/utils.py
@@ -163,65 +163,50 @@ def copy_directory(source_dir: Path, dest: Path):
 
 def copy_file(source: str, dest: str, ignore_mtime: bool = False) -> bool:
     """
-    Copy a source file to a destination file within a container runtime context.
+    Used to copy a source file to a destination file in the container runtime
+    context subfolder.
 
-    This preserves build cache correctness: the file is only copied if it doesn't exist or has changed.
-    Use `copy_directory()` for directories.
+    This utility function helps in maintaining the build cache. We only want
+    to copy the file if it doesn't exist, or if it has changed between builds.
+    See the `copy_directory()` function for the directory copy equivalent.
 
-    :param source: Source file path.
-    :param dest: Destination file path.
-    :param ignore_mtime: If True, modification times are ignored in change detection.
-    :returns: True if the file was copied, False otherwise.
-    :raises: Exception if source or destination is a directory.
+    :param source str: Path to a source file.
+    :param dest str: Path to a destination file within the context subdir.
+    :param ignore_mtime bool: Whether or not mtime should be considered.
+
+    :returns: True if the file was copied, False if not.
+
+    :raises: Exception if called with the path to a directory. This helps to
+        catch programming errors.
     """
-    source_path = Path(source)
-    dest_path = Path(dest)
+
+    should_copy = False
 
     if os.path.abspath(source) == os.path.abspath(dest):
         logger.info("File %s was placed in build context by user, leaving unmodified.", dest)
         return False
-
-    if source_path.is_dir() or dest_path.is_dir():
-        raise Exception(f"{'Source' if source_path.is_dir() else 'Destination'} {source} can not be a directory. Please use copy_directory instead.")
-
-    # Handle symlink logic
-    if source_path.is_symlink():
-        source_target = os.readlink(source)
-
-        if dest_path.is_symlink():
-            dest_target = os.readlink(dest)
-            if source_target == dest_target:
-                logger.debug("Symlink %s already exists and matches.", dest)
-                return False
-            logger.debug("Symlink %s target differs and will be overwritten.", dest)
-        elif dest_path.exists():
-            logger.debug("Destination %s is a regular file and will be replaced by symlink.", dest)
-
+    if Path(source).is_dir():
+        raise Exception(f"Source {source} can not be a directory. Please use copy_directory instead.")
+    if Path(dest).is_dir():
+        raise Exception(f"Destination {dest} can not be a directory. Please use copy_directory instead.")
+    if not os.path.exists(dest):
+        logger.debug("File %s will be created.", dest)
+        should_copy = True
+    elif Path(source).is_symlink() and Path(dest).is_symlink() and os.path.realpath(source) == os.path.realpath(dest):
+        logger.debug("Symlink %s already exists and matches.", dest)
+        should_copy = False
+    elif not filecmp.cmp(source, dest, shallow=False):
+        logger.warning('File %s had modifications and will be rewritten', dest)
+        should_copy = True
+    elif not ignore_mtime and os.path.getmtime(source) > os.path.getmtime(dest):
+        logger.warning('File %s updated time increased and will be rewritten', dest)
         should_copy = True
 
-    else:
-        # Regular file logic
-        if not dest_path.exists():
-            logger.debug("File %s will be created.", dest)
-            should_copy = True
-        elif not ignore_mtime and os.path.getmtime(source) > os.path.getmtime(dest):
-            logger.warning("File %s updated time increased and will be rewritten", dest)
-            should_copy = True
-        elif not filecmp.cmp(source_path, dest_path, shallow=False):
-            logger.warning("File %s had modifications and will be rewritten", dest)
-            should_copy = True
-        else:
-            should_copy = False
-
     if should_copy:
-        if dest_path.is_symlink() or dest_path.exists():
+        if Path(dest).is_symlink() or (Path(source).is_symlink() and os.path.exists(dest)):
             os.unlink(dest)
 
-        if source_path.is_symlink():
-            os.symlink(os.readlink(source), dest)
-        else:
-            shutil.copy2(source, dest)
-
+        shutil.copy2(source, dest, follow_symlinks=False)
     else:
         logger.debug("File %s is already up-to-date.", dest)
 
--- a/src/ansible_builder/utils.py
+++ b/src/ansible_builder/utils.py
@@ -200,7 +200,7 @@ def copy_file(source: str, dest: str, ignore_mtime: bool = False) -> bool:
         should_copy = True
 
     if should_copy:
-        shutil.copy2(source, dest)
+        shutil.copy2(source, dest, follow_symlinks=False)
     else:
         logger.debug("File %s is already up-to-date.", dest)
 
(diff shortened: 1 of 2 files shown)


src/ansible_builder/utils.py
```diff
@@ -200,7 +200,7 @@ def copy_file(source: str, dest: str, ignore_mtime: bool = False) -> bool:
         should_copy = True
 
     if should_copy:
-        shutil.copy2(source, dest)
+        shutil.copy2(source, dest, follow_symlinks=False)
```
Shrews, line 209: One thing I noticed is that this will now error on subsequent runs of `create` because the symlink will already exist (does not happen on regular files).

Shrews (commented):
Besides the potential issue I pointed out, I/we need to give some thought on how your use case of using symlinks is going to interact with the code that attempts to preserve the build cache by checking file stats. I'm not certain if there is any impact there at all.

src/ansible_builder/utils.py
clickthisnick: ack good call - can fix that but wondering about something else now - if I remove a file and then run create twice it doesn't get removed from the context.

I usually just `rm -rf context` when I've used ansible-builder so I didn't even think of running it twice - but in this case it should remove the removed files from the context and the resulting image right?

Reproduction:
```bash
mkdir -p ansible/src
touch ansible/src/foo
cat <<EOF > ansible/execution-environment.yml
---
version: 3

dependencies:
  ansible_core:
    package_pip: ansible-core==2.18.6
  ansible_runner:
    package_pip: ansible-runner==2.4.1
  python_interpreter:
      package_system: "python312"
      python_path: "/usr/bin/python3.12"

images:
  base_image:
    name: quay.io/ansible/ansible-runner:latest

additional_build_files:
  - src: src
    dest: src

additional_build_steps:
  append_base: |
    COPY _build/src /src
EOF

cd ansible
ansible-builder create
ansible-builder build
docker run ansible-execution-env ls /src
> foo
rm src/foo
ansible-builder create
ansible-builder build
docker run ansible-execution-env ls /src
> foo (but I deleted foo create/build did not remove from the context)
touch src/apple
ansible-builder create
ansible-builder build
docker run ansible-execution-env ls /src
> apple (It added this to the context)
> foo (But never removed this)
```

clickthisnick: this was a bit tricky because of how symlinks are handled including ones that point to files that don't exist but updated the code and unit tests to reflect what I think this function should do

I still don't quite grok that once a file is in the build context/cache running ansible-builder create won't remove it

src/ansible_builder/utils.py
```diff
@@ -163,44 +163,65 @@ def copy_directory(source_dir: Path, dest: Path):
...
-
-    :returns: True if the file was copied, False if not.
-
-    :raises: Exception if called with the path to a directory. This helps to
-        catch programming errors.
+    :param source: Source file path.
+    :param dest: Destination file path.
+    :param ignore_mtime: If True, modification times are ignored in change detection.
+    :returns: True if the file was copied, False otherwise.
+    :raises: Exception if source or destination is a directory.
```
Shrews: Let's not reformat this.

src/ansible_builder/utils.py
```diff
@@ -163,44 +163,65 @@ def copy_directory(source_dir: Path, dest: Path):
...
-        logger.debug("File %s will be created.", dest)
-        should_copy = True
-    elif not filecmp.cmp(source, dest, shallow=False):
-        logger.warning('File %s had modifications and will be rewritten', dest)
-        should_copy = True
-    elif not ignore_mtime and os.path.getmtime(source) > os.path.getmtime(dest):
-        logger.warning('File %s updated time increased and will be rewritten', dest)
+
+    if source_path.is_dir() or dest_path.is_dir():
+        raise Exception(f"{'Source' if source_path.is_dir() else 'Destination'} {source} can not be a directory. Please use copy_directory instead.")
```
Shrews: This isn't correct in that it always references `{source}`. Generally, reformatting of code that isn't specific to your change is just "noise" that makes the code more difficult to review. I suggest returning it to the original format.

src/ansible_builder/utils.py
```diff
@@ -163,44 +163,65 @@ def copy_directory(source_dir: Path, dest: Path):
...
+
+        if dest_path.is_symlink():
+            dest_target = os.readlink(dest)
+            if source_target == dest_target:
+                logger.debug("Symlink %s already exists and matches.", dest)
+                return False
+            logger.debug("Symlink %s target differs and will be overwritten.", dest)
+        elif dest_path.exists():
+            logger.debug("Destination %s is a regular file and will be replaced by symlink.", dest)
+
```
Shrews: I'm wondering if maybe it would be better if, instead of comparing _what_ the symlinks point to, we just let the `filecmp.cmp()` call below take care of deciding if there are changes? I would think the contents would be more important than the actual file being pointed to. I _think_ with the original code, the `filecmp.cmp()` would have worked with symlinks as-is. Probably needs testing though.

src/ansible_builder/utils.py
```diff
@@ -163,44 +163,65 @@ def copy_directory(source_dir: Path, dest: Path):
...
+        elif not filecmp.cmp(source_path, dest_path, shallow=False):
+            logger.warning("File %s had modifications and will be rewritten", dest)
+            should_copy = True
+        else:
+            should_copy = False
+
     if should_copy:
-        shutil.copy2(source, dest)
+        if dest_path.is_symlink() or dest_path.exists():
+            os.unlink(dest)
```
Shrews, line 207: No need to remove normal files here, just symlinks.

src/ansible_builder/utils.py
```diff
@@ -163,44 +163,65 @@ def copy_directory(source_dir: Path, dest: Path):
...
+
+        if source_path.is_symlink():
+            os.symlink(os.readlink(source), dest)
+        else:
+            shutil.copy2(source, dest)
+
```
Shrews: Could we not just set `follow_symlinks` in `copy2()` to the correct value based on if it is a symlink or not?

Shrews (changes requested):
Left you some comments inline. In general, unnecessary code or comment reformatting just adds noise for the reviewer, so it's best to avoid that when possible.

src/ansible_builder/utils.py
clickthisnick: it works because of `if source_path.is_dir() or dest_path.is_dir():` but ack - will revert

src/ansible_builder/utils.py
clickthisnick: This was to handle the case where:

source = symlink
dest = file

when I run `ansible-builder create`  I expect dest to now be a symlink

that maybe untrue or too much of an edge case - I can guard on that specific case in order to not remove the file each time

src/ansible_builder/utils.py
clickthisnick: yea good point - will update - thanks for the comments

src/ansible_builder/utils.py
clickthisnick: `filecmp.cmp()` with symlinks make the unit tests indeterministic - I install pytest-repeat and ran them 1000 times and copied which I would expect to be False would switch between True and False.

I kept the reproduction steps in the unit tests, but added an extra case before `filecmp.cmp()` which explicitly checks for the symlink conditions

I couldn't track why `filecmp.cmp()` was doing this tho - I was thinking it was something in the os or the pytest fixture for creating tmp files but couldn't figure it out

test/unit/test_utils.py
```diff
@@ -145,3 +157,179 @@ def test_copy_directory(tmp_path):
...
+
+    copied = copy_file(str(source), str(dest))
+    # At this point source and dest are both just files but because they were symlinks previously there is some issue
+    # Potentially it is in tmp_path or the operating system
+    # pytest -vvv -s --count=1000 test/unit/test_utils.py::test_regular_file_overwrites_symlink
+    # We would expect this to be False
+    # 1000 runs using pytest-repeat
+    # 37 times it is False
+    # 963 times it is True
+    assert copied in (False, True)
```
clickthisnick: I am confused by this and couldn't think of a way to get around it

src/ansible_builder/utils.py
```diff
@@ -189,7 +189,10 @@ def copy_file(source: str, dest: str, ignore_mtime: bool = False) -> bool:
         raise Exception(f"Source {source} can not be a directory. Please use copy_directory instead.")
     if Path(dest).is_dir():
         raise Exception(f"Destination {dest} can not be a directory. Please use copy_directory instead.")
-    if not os.path.exists(dest):
+    if Path(source).is_symlink() and Path(dest).is_symlink() and os.readlink(source) == os.readlink(dest):
```
clickthisnick, line 192: I couldn't figure out why `not filecmp.cmp(source, dest, shallow=False)` wasn't working properly for symlinks so I just added this as an explicit check.

It needs to be before:
`elif not os.path.exists(dest):`

To handle the case where the symlink is pointing to a file that does not yet exist (on purpose)

test/unit/test_utils.py
Shrews: That's very odd. I changed the assert to check for `False` and ran the test 5000 times and it never failed.

src/ansible_builder/utils.py
Shrews: `filecmp.cmp()` should totally work with symlinks. Can you show me an example where it isn't working?

Shrews: Oh, maybe you meant `filecmp.cmp()` didn't work _because_ the link was pointing to a non-existing file? If so, then yeah, an exception would be thrown in that case since that function attempts to `stat` the actual file (not the link). I.e., it follows the symlink for comparing.

test/unit/test_utils.py
clickthisnick: I'm developing within a container maybe that is what is going on - I'll try on mac osx directly

I'm fine with just changing the assertion to False, removing the pytest-repeat comments in the tests and knowing that my environment is weird and it might invalidate the cache for me

Does that seem okay given you ran it 5k times and it always worked 😅

Shrews: Yes, definitely change it to the expected value, otherwise this isn't really a good test. Can remove all of the comments about the tests failing too. It likely has something to do with your environment. In the meantime, I'll test it some more on my other systems, just to verify..

clickthisnick: Interesting - the CI run reproduced the same issue I was experiencing locally

clickthisnick: It def wasn't deterministically failing for me - curious if the CI ran again if it would pass/fail on a different interpreter

Shrews: I still cannot reproduce this on my systems. But, on a hunch, can you try adding a call to `filecmp.clear_cache()` in between the consecutive `copy_file()` calls and see if you still get the failures on your machine?

clickthisnick: Your suggestion fixes things - I didn't realize it had a cache, was focusing on the files themselves. The outcome (False -> True) was cached is the issue:

After First Copy (Before Cache Clearing):
`
filecmp._cache:
{('/tmp/pytest-of-clickthisnick/pytest-122/test_regular_file_overwrites_s0/source.txt', '/tmp/pytest-of-clickthisnick/pytest-122/test_regular_file_overwrites_s0/dest_symlink', (32768, 3, 1751046892.509871), (32768, 3, 1751046892.509871)): False}
`

After Cache Clearing and Second Copy:
`
filecmp._cache:
{('/tmp/pytest-of-clickthisnick/pytest-122/test_regular_file_overwrites_s0/source.txt', '/tmp/pytest-of-clickthisnick/pytest-122/test_regular_file_overwrites_s0/dest_symlink', (32768, 3, 1751046892.509871), (32768, 3, 1751046892.509871)): True}
`

clickthisnick: Do you have any preference what should happen here

If I call `ansible-builder create` twice there is no shared cache between the runs so that's okay
but if `copy_file` gets twice in the same program it could have this issue

-----

I'm thinking:

1.) In `copy_file` we could look if the input paths are already in cache and pop them (or just clear all cache) before using `filecmp.cmp()` to check
 
 2.) we could just add to the unit test which is probably fine for most current use-cases since each `ansible-builder create` has its own cache in memory

clickthisnick: minimal reproduction case that works at least in my env
```python
import filecmp, shutil

with open("/tmp/foo", "w") as f:
    f.write("hello")

with open("/tmp/bar", "w") as f:
    f.write("world")

# Should be False
print(filecmp.cmp("/tmp/foo", "/tmp/bar", shallow=False))

# Overwrite bar with foo (same size)
shutil.copyfile("/tmp/foo", "/tmp/bar")

# Still returns False due to stale cache!
print(filecmp.cmp("/tmp/foo", "/tmp/bar", shallow=False))  # ❌ Incorrect

# Correct after clearing cache
filecmp.clear_cache()
print(filecmp.cmp("/tmp/foo", "/tmp/bar", shallow=False)) 
```

Shrews: We don't need to modify `copy_file()` here. This only happens in the test because `filecmp.cmp()` is called back-to-back on the same file after it changes which, [according to the `filecmp` docs](https://docs.python.org/3/library/filecmp.html#filecmp.clear_cache), can cause cache issues. This isn't going to happen in the normal code. So just putting the `clear_cache()` call in the tests, with a comment on why we are adding it, is the proper solution, IMO.

Shrews (approved):
Thanks for the contribution.

Since this is a slight change in behavior, I think it is best suited for the next 3.2 release, and so we won't backport it for another 3.1.x release.

PR #752, opened 2025-07-08
Add pip check feature

### LINKED ISSUE #416: Add `pip check` after final install
The `pip` depsolver currently only issues warnings when an incompatible version of a package is installed (eg, when a later install with a conflicting dep breaks a previous one)- someone looking carefully at their container build output might see the warning text that an incompatible dep was installed, but there's no way to make that an error today.

Adding a `pip check` at the end of the build (available in some form since ~pip 9) will, for more modern versions of pip, re-do the dep check for all installed packages and issue an error listing the broken deps (which would fail the build). We'll probably also need EE definition syntax to disable/ignore the results of this check, but surfacing the problem by default is probably the right thing to do. Suspect we'll find a lot more busted collection deps this way, for better or worse.

Shrews:
New Feature: Validate installed Python dependencies in final build stage

- Calls `pip check` immediately after Python dependencies are installed in final build stage.
- Adds `skip_pip_check` option to the EE schema to control the new behavior (enabled by default).

Fixes #416

### DIFF
--- a/src/ansible_builder/ee_schema.py
+++ b/src/ansible_builder/ee_schema.py
@@ -372,6 +372,10 @@ schema_v3 = {
                     "description": "Disables the installation of pip in the base image",
                     "type": "boolean",
                 },
+                "skip_pip_check": {
+                    "description": "Disables verifying installed Python packages have compatible dependencies",
+                    "type": "boolean",
+                },
                 "workdir": {
                     "description": "Default working directory, also often the homedir for ephemeral UIDs",
                     "type": ["string", "null"],
@@ -476,6 +480,7 @@ def _handle_options_defaults(ee_def: dict):
 
     options.setdefault('skip_ansible_check', False)
     options.setdefault('skip_pip_install', False)
+    options.setdefault('skip_pip_check', False)
     options.setdefault('relax_passwd_permissions', True)
     options.setdefault('workdir', '/runner')
     options.setdefault('package_manager_path', '/usr/bin/dnf')
--- a/docs/definition.rst
+++ b/docs/definition.rst
@@ -457,6 +457,12 @@ builder runtime functionality. Valid keys for this section are:
       pip manually if the current method of pip installation does not work for you.
       The default is ``False``.
 
+    ``skip_pip_check``
+      This boolean value controls whether or not a call to ``pip check`` is performed in the final
+      build stage of the build process. If this call errors because it finds incompatible dependencies
+      of installed Python packages, the build will fail. To disable this validation, set the value to ``True``.
+      The default is ``False``.
+
     ``relax_passwd_permissions``
       This boolean value controls whether the ``root`` group (GID 0) is explicitly granted
       write permission to ``/etc/passwd`` in the final container image. The default entrypoint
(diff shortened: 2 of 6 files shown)


Akasurde (approved):

src/ansible_builder/ee_schema.py
```diff
@@ -372,6 +372,10 @@
                     "description": "Disables the installation of pip in the base image",
                     "type": "boolean",
                 },
+                "skip_pip_check": {
```
nitzmahone, line 375: Doesn't the presence of `additionalProperties: False` mean this (and anything else that's been added since 3.0 released) is technically a breaking change to the v3 schema? Dunno if it's worth doing anything about at this point...

nitzmahone (commented):
Change looks good, though we should probably clarify our schema policy on backward/forward-compatibility for a given schema version.

src/ansible_builder/ee_schema.py
Shrews: This is a conversation worth having since this is the first time we've encountered this situation.

Adding a new schema feature is nothing new. For example, we added these new schema keywords in the `3.1` release:

- [new `tags` option](https://github.com/ansible/ansible-builder/pull/555)
- [new `exclude` keyword](https://github.com/ansible/ansible-builder/pull/664)

The new `skip_pip_check` feature is targetted for a new `3.2` release, similar to those changes referenced above. 

What makes this change different, however, is that it changes how pre-existing EEs, which do NOT yet reference this new schema option, operate when run with the as-yet-unreleased `3.2` version. Their builds would now fail _if_ `pip check` finds any errors, whereas with version `3.1` it would succeed. That was the behavior suggested by your issue report (unless I misunderstood?). That behavior itself would be the thing that could be considered a breaking change, not the introduction of a new schema keyword.

Shrews: As an alternative, we could just close the feature as "not planned" and document the use case in our `Common Scenarios` section of our documentation that instructs users how to add their own call to `pip check` in the `append_final` build stage.

nitzmahone: I'm fine with either. IIRC builder doesn't advertise as semver, so introducing this (technically) breaking change on a new Y release shouldn't be too big a deal. I still think it's a decent check to be on by default, but there will certainly be cases that need an easy "off" switch.

Conceptually, these "additional schema values" things from 3.1/3.2 aren't that far off from the runner kwargs problem, except that builder's kind of a one-shot thing where there's not too much worry about persistence and concurrent usage by multiple versions that need to guarantee a common understanding. I'm more OK in this case with just "yeah, it's all schema v3, but older builder releases will barf on new keys, don't do that", because the alternative is either declaring newer schema versions for every addition or moving to a "guess the version", neither of which are fun...

Shrews: I'd prefer to do the alternative solution here and not force this check on users for this schema version. Having an existing and working EE suddenly start failing if built with a newer version doesn't sound like a great UX. I'll add a new PR for our docs that describes adding this check manually instead.

PR #780, opened 2026-02-24
Migrate setup.cfg to pyproject.toml

Shrews:


Akasurde (approved):

tox.ini
```diff
@@ -68,3 +68,7 @@ allowlist_externals = /bin/sh
 commands =
     /bin/sh -c "rm -rf {toxinidir}/test/coverage/*"
     /bin/sh -c "rm -rf {toxinidir}/docs/{_,}build"
+
+[flake8]
```
webknjaz, line 72: @Shrews I'd prefer this in a `.flake8` so that it's not mixed up in other tool configs.

setup.cfg
```diff
@@ -1,53 +0,0 @@
...
-    Programming Language :: Python :: 3.12
-    Programming Language :: Python :: 3.13
-    Programming Language :: Python :: 3.14
-    Programming Language :: Python :: 3 :: Only
-    Topic :: Software Development :: Build Tools
-    Topic :: System :: Systems Administration
-    Topic :: Utilities
-
-[flake8]
-# W503 - Line break occurred before a binary operator
```
webknjaz, line 33: @Shrews forgot to copy the explanation?

pyproject.toml
```diff
@@ -1,9 +1,60 @@
 [build-system]
-requires = ["setuptools>=45, <=72.1.0", "setuptools-scm[toml]>=6.2, <=8.1.0"]  # pin max versions of build deps and update as needed
+requires = ["setuptools>=77.0.3, <=82.0.0", "setuptools-scm[toml]>=6.2, <=9.2.2"]  # pin max versions of build deps and update as needed
```
webknjaz, line 2: @Shrews you may want to also reproduce https://github.com/jazzband/pip-tools/pull/2261 to address a few warnings in `setuptools`.

webknjaz:
@Shrews have you compared the diff of before/after of what `wheel unpack` extracts? Also `twine check --strict`.

tox.ini
Shrews: I don't see a benefit from doing this. Since our tests are `tox`-driven anyway, I prefer to leave it as-is, and it's one less file.

setup.cfg
Shrews: Nope. Keeping it wasn't necessary.

pyproject.toml
Shrews: I'm not seeing the relationship between that PR and builder. If there is something valuable there you think builder could benefit from, submit a PR and we can discuss it there.

Shrews:
> @Shrews have you compared the diff of before/after of what `wheel unpack` extracts? Also `twine check --strict`.

Yes.

pyproject.toml
webknjaz: fix: https://github.com/ansible/ansible-builder/pull/786

PR #795, opened 2026-03-11
Add missing class docstrings across modules and tests

Shrews:
- Added docstrings for multiple classes in source and test code to improve clarity and consistency.
- Removed pylint suppression for missing-class-docstring as the issue is now resolved.

### DIFF
--- a/src/ansible_builder/policies.py
+++ b/src/ansible_builder/policies.py
@@ -47,10 +47,9 @@ class BaseImagePolicy(ABC):
     """
     Defines an abstract base class for creating and managing podman policy files.
 
-    This class serves as a blueprint for derived classes that aim to define signed
-    identity types and generate policy data for podman. It provides an interface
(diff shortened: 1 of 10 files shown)


src/ansible_builder/policies.py
```diff
@@ -32,7 +44,14 @@ class SignedIdentityType(Enum):
 
 
 class BaseImagePolicy(ABC):
+    """
+    Defines an abstract base class for creating and managing podman policy files.
 
+    This class serves as a blueprint for derived classes that aim to define signed
+    identity types and generate policy data for podman. It provides an interface
+    to specify the signed identity types and generate corresponding policy data,
+    and implements functionality to save the generated policy to a file.
```
s-hertel: This seems a little redundant, could just be `This class provides an interface to specify the signed identity types and generate corresponding policy data, and implements functionality to save the generated policy to a file.`

s-hertel (approved):

PR #811, opened 2026-05-07
feat: add platform support for image builds

### LINKED ISSUE #651: Concerning mutil arch builds
Using a simple starting point:

```yaml
---
version: 3

dependencies:
  ansible_core:
    package_pip: ansible-core==2.16.3
  ansible_runner:
    package_pip: ansible-runner
  python_interpreter:
    package_system: "python3.12"

images:
  base_image:
    name: registry.fedoraproject.org/fedora-minimal:39

options:
  package_manager_path: /usr/bin/microdnf
```

multi-arch can be accomplished with:
```
$ ansible-builder create                                                                           
Complete! The build context can be found at: /home/bthornto/github/builder-march/context
$ podman manifest create ee-test
$ podman build --platform linux/amd64,linux/arm64  --manifest ee-test -f context/Containerfile
```

```
$ podman manifest inspect ee-test | grep 'arch\|digest'
            "digest": "sha256:969841c4a366c835c8a0ddf28688487bf29b363af1b9853e8d8b3c6860576a1c",
                "architecture": "arm64",
            "digest": "sha256:4068e51486c0d89313c3542b5b21e9e8a15a7b83d4937db3983f498471045600",
                "architecture": "amd64",
```

```
$ podman images --digests | grep '6a1c\|5600'         
<none>                                     <none>      sha256:4068e51486c0d89313c3542b5b21e9e8a15a7b83d4937db3983f498471045600  d2f04819aabd  10 minutes ago  250 MB
<none>                                     <none>      sha256:969841c4a366c835c8a0ddf28688487bf29b363af1b9853e8d8b3c6860576a1c  e08bdf83069c  10 minutes ago  335 MB
```
Later pushing as needed `podman manifest push ee-test`

Some questions come to mind:

1) Can it be assumed the resulting images are configured in the same manner they would be had `ansible-builder build` been used instead? (Are there any reasons to not do the above when a multi-arch is needed?)
2) Is there an opportunity here for the user to provide an alternative to the built-in "build" command https://github.com/ansible/ansible-builder/blob/devel/src/ansible_builder/main.py#L189  If supported in the `execution-environment.yml` file it could be used later or shared to rebuild in the same manner if needed. (the tag and manifest keywords would need to be account for)  If supplied as podman/docker specific this could acount for buildx.
3) (if not) Can we document the above approach in a FAQ or examples section of the docs?  What would be the best place? (happy to PR)

thank again for all the builder work, everytime I use it I reappreciate how handy it is.

-brad

dbrennand:
## Summary

This PR adds support for passing `--platform` through `ansible-builder build` to the underlying container runtime build command.

It includes:
- a new `--platform` CLI option for `ansible-builder build`
- plumbing in `AnsibleBuilder` to append `--platform=<value>` to the build command
- documentation covering single-platform and multi-arch usage
- unit coverage for single-platform and multi-arch command generation
- integration coverage verifying the platform flag is passed through during builds

Fixes https://github.com/ansible/ansible-builder/issues/651 - Specifically https://github.com/ansible/ansible-builder/issues/651#issuecomment-1957821051


Shrews:
How is this an improvement from using `--extra-build-cli-args` to pass the argument?

Shrews:
I'm going to go ahead and close this as I don't think this is something we want. Specifically, the `--extra-build-cli-args` option was added so that we aren't endlessly adding new options to ansible-builder in an attempt to match the underlying build command (`docker build` or `podman build`) CLI options. If there is a problem with this specific build option for the build platform, please open an issue where we may discuss it and together plan a new feature based on that discussion.

dbrennand:
The premise of this was to provide a simpler UX when creating multi-arch EEs with Podman. There has already been discussion in https://github.com/ansible/ansible-builder/issues/651#issuecomment-1957821051 which mentions that when using podman with `ansible-builder`, you need to chain `ansible-builder create` and then separately run `podman build --platform ... --manifest`.

